PR #34, opened 2017-11-20
Adding SLES 11 PPC 64 support 

bblondin:
Added support of SLES 11 PPC64 + minor fixes

- Download zlib and git
- Compile from source
- Updated name titles for SLES 11 x86 install of git-core
- Fixed Java 8 SLES 11 response file name
- Missing when x86 on git-core


ansible/playbooks/sles.yml
```diff
@@ -65,24 +65,82 @@
...
       zypper: name={{item}} state=installed update_cache=yes
       with_items:
         - git-core
         - perl-Error
       when:
         - ansible_distribution_major_version == "11"
+        - ansible_architecture == "x86"
       tags: build_tools
+      
+    - name: Download Zlib - SLES 11 PPC64
```
jdekonin: This might not be as likely for zlib, but maybe zlib and git should be in their own files and not specific to a platform.  For example on rhel7 the default git doesn't support options required and will likely need git rebuilt from source as well.

bblondin: Moving forward we need to start looking at if breaking out selections such as git is the best method for managing these playbooks. As the this playbooks are getting very long

PR #39, opened 2017-11-21
RHEL 7 s390 updates

bblondin:
Disabled 'Enable rhel-7-server-extras-rpms Repo' on s390 hosts
- ansible_architecture != "s390x"

Changed 'ansible_processor_vcpus' to use 'ansible_processor_cores' when the host is s390
This is a workaround as ansible only shows 1 cpu on s390 when using ansible_processor_vcpus
cmake and ccache

Disabled Docker install on S390
- ansible_architecture != "s390x"

### DIFF
--- a/ansible/playbooks/rhel.yml
+++ b/ansible/playbooks/rhel.yml
@@ -218,7 +218,7 @@
       file:
         path: "{{ item }}"
         state: absent
-      with items:
+      with_items:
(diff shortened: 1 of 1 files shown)


gdams:
I don't think we currently have any rhel7 s390x hosts but I am happy to add this in case we get some

PR #30, opened 2017-10-20
Add XML::Parser module into playbooks

renfeiw:
- Test framework will require XML::Parser module to parse playlist.xml
files

Signed-off-by: Renfei Wang <renfeiw@ca.ibm.com>

PR #60, opened 2017-11-27
Add XML::Parser to rhel ansible scripts

jdekonin:
This is a follow up on #30 which was closed without resolution.

### DIFF
--- a/ansible/playbooks/rhel.yml
+++ b/ansible/playbooks/rhel.yml
@@ -35,6 +35,7 @@
         - bind-utils
         - cpio
         - cups-devel
+        - expat-devel
         - freetype-devel
         - gcc
         - gcc-c++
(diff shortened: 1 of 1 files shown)


ansible/playbooks/rhel.yml
```diff
@@ -35,6 +35,7 @@
         - bind-utils
         - cpio
         - cups-devel
+        - expat-devel
```
gdams: Is this meant to have been moved?

jdekonin: Yes, it was meant to be moved.  Previously it was only installed if rhel7 and ppcle were set and required as a git src compile.  Its a core dependency for XML::Parser so I moved expat to the generic install section.

bblondin: Tested, package is available on all archs and RHEL 6/7

PR #58, opened 2017-11-26
Ansible guide improvements

karianna:

PR #61, opened 2017-11-27
Fixed issues caused by azul repository

bblondin:
This change broke the playbook for all archs other than x86
https://github.com/AdoptOpenJDK/openjdk-infrastructure/commit/75195f8750abbf18869220ac5f84633bf3cc5e58

GPG error occurs while executing apt-get after "Add the azul repository to apt" was added to the playbook.

Issue: Azul Zulu's GPG key needs to be appiled to the instance.

Solution:
Add:
- name: Add Azul Zulu GPG Package Signing Key
  apt_key:
    url: http://repos.azulsystems.com/RPM-GPG-KEY-azulsystems
    state: present

Remove: zulu-9-amd64 from main build_tools

Add:
    - name: Install zulu-9-amd64
      apt: pkg={{ item }} state=latest
      with_items:
        - zulu-9-amd64
      when:
      tags: build_tools

### DIFF
--- a/ansible/playbooks/ubuntu.yml
+++ b/ansible/playbooks/ubuntu.yml
@@ -36,6 +36,11 @@
       apt_repository: repo='deb http://repos.azulsystems.com/ubuntu stable main'
       tags: patch_update
 
+    - name: Add Azul Zulu GPG Package Signing Key
+      apt_key:
+        url: http://repos.azulsystems.com/RPM-GPG-KEY-azulsystems
+        state: present
+        
     - name: OS update -- apt-get upgrade
       apt: upgrade=safe update_cache=yes
       tags: patch_update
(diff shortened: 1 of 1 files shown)


gdams:
@bblondin so this was actually added because we weren't able to just `apt-get install openjdk-9-jdk`. Have you been able to install openjdk9 using apt?

karianna:
I had a slightly different way of specifying that apt-key in #58 but this is a better PR as it only installs 9 on x86

bblondin:
We need to be careful when we applying a chance to a playbook as it will effect all of the different OS versions and Platforms (architectures) supported in that playbook. 

As with this package "zulu-9-amd64" the name says it only works on amd64 so we need to ensure it has a ```when``` condition  ```ansible_architecture == "x86_64"``` so that it doesn't fail on other architecture like ppc64, ppc64le, armv7l, etc

karianna:
@bblondin Agreed - @neomatrix369 and I are working on a bunch of Vagrant files so that people can sanity test before submitting PR's, see #58

PR #67, opened 2017-11-30
Install windows resources

vsebe:
- install java 7 jdk instead of jre
- add java 8 bin directory to path
- add check for freemarker and freetype
- install freetype, freemarker into c:/openjdk directory
- grant jenkins user permissions on freetype

### DIFF
--- a/ansible/playbooks/windows.yml
+++ b/ansible/playbooks/windows.yml
@@ -74,7 +74,13 @@
         path: C:\temp
         state: directory
       tags: basic_config
-        
+
+    - name: Create C:\openjdk directory
+      win_file:
+        path: C:\openjdk
+        state: directory
+      tags: basic_config
+
     - name: Create cmd.exe shortcut
       win_shortcut:
         src: C:\Windows\System32\cmd.exe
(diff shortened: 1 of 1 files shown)


vsebe:
@bblondin please review, thank you!

ansible/playbooks/windows.yml
```diff
@@ -193,24 +206,56 @@
...
+    - name: Test if freetype is already installed
+      win_stat:
+        path: 'C:\openjdk\freetype-2.5.3'
+      register: freetype_installed
+      tags: freetype
+
+    - name: Test if freetype is already installed
+      win_stat:
+        path: 'C:\openjdk\freetype-2.5.3'
+      when: (freetype_installed.stat.exists == false)
```
jdekonin: You perform 2 checks here on the directory path setting two different registers.  Not sure this is what you want to do.

bblondin: the second one needs to have its name updated (check if already downloaded)
and path changed to c:\temp\

ansible/playbooks/windows.yml
```diff
@@ -74,7 +74,20 @@
...
+      register: openjdk_installed
+      tags: basic_config
+
+    - name: Create C:\openjdk directory
+      win_file:
+        path: C:\openjdk
+        state: directory
+      when: (openjdk_installed.stat.exists == false)
+      tags: basic_config
+
```
bblondin: This section of code is redundant ```Test if C:\openjdk directory exists```
as ```win_file``` automatically does the same thing. If the folder is already there it will skip.
lines 78-82 and 88 can be removed

ansible/playbooks/windows.yml
```diff
@@ -326,6 +373,16 @@
...
+    - name: Add JAVA_HOME/bin to Path
+      win_path:
+        name: Path
+        elements: 
+          - 'C:\Program Files\Java\jdk8u144-b01\bin'
+        scope: machine
+        state: present
+      when: (java8_installed.stat.exists == false)
+      tags: java8
+
```
bblondin: In what situation would java8 not exist?

ansible/playbooks/windows.yml
vsebe: No I do not want it, good catch @jdekonin!

ansible/playbooks/windows.yml
vsebe: New machine

ansible/playbooks/windows.yml
vsebe: removed

vsebe: also added clean up for the *.zip files

PR #70, opened 2017-12-03
Remove CCACHE from APT

gdams:
We HAVE to use ccache 3.1.9 so fetching from APT won't work for us hence the manual steps below

### DIFF
--- a/ansible/playbooks/ubuntu.yml
+++ b/ansible/playbooks/ubuntu.yml
@@ -50,7 +50,6 @@
       with_items:
         - autoconf
         - bison
-        - ccache
         - cpanminus
         - cpio
         - flex


karianna:
Can you not specify version in apt or point to an apt repo that has it?

gdams:
Not sure, @bblondin may know?

bblondin:
That becomes harder to do, as each OS version many have a different ccache as default and the way it names older ones can differ. 

It might be best to download the version we want and compile it. as we do in RHEL
https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/rhel.yml

bblondin:
```
    - name: Download ccache.tar.gz
      get_url:
        url: https://www.samba.org/ftp/ccache/ccache-3.1.9.tar.gz
        dest: /home/{{ Jenkins_Username }}/ccache.tar.gz
        mode: 0440
      tags: ccache
      
    - name: Extract ccache
      unarchive:
        src: /home/{{ Jenkins_Username }}/ccache.tar.gz
        dest: /home/{{ Jenkins_Username }}
        copy: False
      tags: ccache

    - name: Running ./configure & make for CCACHE
      shell: cd /home/{{ Jenkins_Username }}/ccache-3.1.9 && ./configure && make -j {{ ansible_processor_vcpus }} && sudo make install
      become: yes
      tags: ccache

```

gdams:
we already do this in the ubuntu playbook but it gets skipped because ccache has already been installed by apt

jdekonin:
You could add a version check.  If the version installed is greater than 3.1.9 uninstall, then reinstall 3.1.9...

Out of curiosity why is this change originating from a branch off the main repo and not off a personal fork?

karianna:
@gdams has been caught! ;-)

gdams:
Haha yeah I used the github web editing tool! 👍 I will delete the branch once the PR is landed

PR #71, opened 2017-12-05
Add openj9 package requirements for ubuntu based systems

jdekonin:
* add gcc-4.8, g++-4.8
* add curl, numa, freetype

### DIFF
--- a/ansible/playbooks/ubuntu.yml
+++ b/ansible/playbooks/ubuntu.yml
@@ -52,9 +52,12 @@
         - bison
         - cpanminus
         - cpio
+        - curl
         - flex
         - g++
         - gcc
+        - g++-4.8
+        - gcc-4.8
         - git
         - libasound2-dev
         - libcups2-dev
@@ -62,6 +65,7 @@
         - libelf-dev
         - libexpat1-dev
         - libffi-dev
+        - libfreetype6-dev
         - libx11-dev
         - libxext-dev
         - libxrender-dev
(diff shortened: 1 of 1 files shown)


jdekonin:
Tested with ansible tower and openstack machines on x86 and ppc.  Verified package availability on s390.

karianna:
Hmm, I thought we were already building on Ubuntu?

karianna:
Also not convinced that the symlinking wouldn't break the hotspot builds...

jdekonin:
From what I can tell of the history the hotspot builds run on fedora2.4 (gcc 6.3.1) and not on ubuntu but I get your meaning.  If we were to run on ubuntu this might not be what is wanted.  How do you propose that I set the gcc to use within the machine setup scripts?

karianna:
@jdekonin I think this might be a conversation on the #getopenjdk channel in slack.  Are you on there?

jdekonin:
yes I am.  And there is a build for openj9 on ubuntu ppcle, but if you need to rebuild a machine based on the current script you will need to add these changes to get a functional system.

jdekonin:
Symlinks removed.

ansible/playbooks/ubuntu.yml
```diff
@@ -53,16 +53,21 @@
...
+        - gcc-4.8
         - git
         - libasound2-dev
         - libcups2-dev
         - libdwarf-dev
         - libelf-dev
         - libexpat1-dev
         - libffi-dev
+        - libfreetype6-dev
+        - libnuma-dev
```
bblondin: This package "libnuma-dev" is not available on arm 32bit - armv7l or at least not in the default repos.
(raspbian jessie)

      when:
        - ansible_architecture != "armv7l"

jdekonin:
Right you are.  I didn't think to test against arm.  Modification made to limit numa install on armv7l, but I am thinking this should likely be on aarch64 no?

PR #73, opened 2017-12-06
cmake - ignore_errors: yes

bblondin:
added ignore_errors: yes to Cleanup cmake


karianna:
What errors would be ignored?

PR #79, opened 2017-12-07
file vs win_file error

bblondin:


karianna:
Why do we want to ignore errors?

bblondin:
The playbook shouldn't fail in the middle of its run over something that is trivial and would cause no harm.

PR #87, opened 2017-12-08
Add checks for cmake to avoid performing cmake actions on each run of…

jdekonin:
… the playbook


jdekonin:
Tested against x86 rhel6 and ppcle rhel7 on AWX.

Is there a requirement for cmake 3.7.2 specifically or just a requirement for cmake.  If its specific to the version these changes aren't valid.  Depends on the reason cmake is 'locked' to this version.

PR #97, opened 2017-12-15
Split out jck ansible config into jckservices and execution playbooks

sxa555:


gdams:
Perhaps we could do a run on the jckservices machine to set it up/confirm this playbook is fully working?

PR #105, opened 2017-12-19
Disabled adding jenkins user to audio group armv7l

bblondin:
Disabled adding jenkins user to audio group for armv7l


bblondin:
type-o

PR #108, opened 2017-12-22
Add https support for git on linux arm

### LINKED ISSUE #107: git clone fails using https on rasp pi
Using the current ansible script on a rasp pi I get this error :  
```
lnxarmpirt-i1:/tmp $ git clone https://github.com/eclipse/openj9.git
Cloning into 'openj9'...
fatal: Unable to find remote helper for 'https'
```
Fix is to add the libcurl4-openssl-dev package and (re)compile/install git.  I'll submit a change for this shortly.  

Edit:  the alternative fix is to apt-get the git and git-core packages, but that would only work if git version is > 2.0 (openj9 requirement).

jdekonin:
Fixes #107

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Common/vars/Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Common/vars/Ubuntu.yml
@@ -49,6 +49,7 @@ Additional_Build_Tools_x86_64:
 
 Additional_Build_Tools_aarch64:
   - libpng-dev
+  - libcurl4-openssl-dev
 
 Test_Tool_Packages:
   - ant


gdams:
I was under the impression that this was required on raspberry pi's? If so it is most likely that this is an `armv7l` dependency rather than an `aarch64` dependency. Please correct me if I am wrong though!

jdekonin:
Technically its a git requirement for https functionality so it will be needed on all arm setups as long as the script is also compiling git from source.  A bit of research shows the package name whether its v7 or v8 is the same.  But I couldn't answer for certain this would work on a v8 device without access to one so I could test.

I know @bblondin has done a number of setups, if he or anyone else could confirm/deny/advise??

This isn't urgent it can certainly wait until folks get back from year end activities.

bblondin:
Additional_Build_Tools_aarch64 only effects 64 bit arm and thus would have no effect on a system like: lnxarmpirt-i1 as its 32 bit ```armv7l```

gdams:
@karianna did you mean to merge this?

karianna:
@gdams yeah merged by accident - want me to create a PR to reverse it?

gdams:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/115

PR #118, opened 2018-01-08
Update AIX scripts

jdekonin:


ansible/playbooks/aix.yml
```diff
@@ -34,7 +34,7 @@
     ################################
     - name: Download AIX filesystem configuration script
       get_url:
-        url: https://raw.githubusercontent.com/AdoptOpenJDK/openjdk-infrastructure/master/ansible/playbooks/scripts/AIX_filesystem_config.sh
+        url: https://raw.githubusercontent.com/jdekonin/openjdk-infrastructure/master/ansible/playbooks/scripts/AIX_filesystem_config.sh
```
bblondin: Why are we pointing to your username?

jdekonin: Its the only way I could test, this would need to be changed prior to commit.

I have tested against my own fork and this works only on the 7.1.4 distro as there are version specific rpms.  Thats a future todo, if/when further versions are required.

I also encountered an issue with installing/upgrading the zlib rpm.  The version (1.2.11-1.aix5.1) commented out is required for openjdk to build, but that version breaks python.  My current answer for that is it needs to be performed manually.  Not ideal, but then at least the rest of the system setup can be performed via ansible.

ansible/playbooks/scripts/AIX_RPM_PKGs_Install.sh
```diff
@@ -7,65 +7,139 @@
...
-rpm -Uvh tk-8.6.6-1.aix5.1.ppc.rpm 
-rpm -Uvh tkinter-2.7.13-1.aix6.1.ppc.rpm 
-rpm -Uvh python-tools-2.7.13-1.aix6.1.ppc.rpm
-rpm -Uvh git-2.2.2-3.aix5.1.ppc.rpm 
+rpm -e $(rpm -qa | grep cloud-init) 2>/dev/null					# Remove 'cloud-init' 
+wget http://www.bullfreeware.com/download/bin/3183/bash-4.4-3.aix6.1.ppc.rpm
+rpm -Uvh bash-4.4-3.aix6.1.ppc.rpm
+
+wget http://www.bullfreeware.com/download/bin/3517/libgcc-6.3.0-1.aix7.1.ppc.rpm
+rpm -Uvh libgcc-6.3.0-1.aix7.1.ppc.rpm
```
sxa555: This failed for me because AIX-rpm saying:
`AIX-rpm < 7.2.0.0. is needed by libgcc-6.3.0.1`
I have `AIX-rpm-7.2.0.1-1` on my system.

sxa555:
Is there a reason not to use `rpm -Uvh` directly with the wget/ftp URLs instead of just giving the URL to `rpm`? I could understand if we were caching the RPMs in case the playboks but the way it's set up just now there's no checking fo the `.rpm` files' presence so it'll re-download if you run the script.

jdekonin:
I actually didn't know it could be done that way until recently, I just hadn't updated the script.  I am testing the yum/AIX toolbox method we were discussing via slack.  If that works better I am fine with switching to that method.

ansible/playbooks/scripts/AIX_RPM_PKGs_Install.sh
```diff
@@ -7,65 +7,139 @@
...
+rpm -Uvh zip-3.0-2.aix6.1.ppc.rpm
+
+wget ftp://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/sed/sed-4.4-1.aix6.1.ppc.rpm
+rpm -Uvh sed-4.4-1.aix6.1.ppc.rpm
+
+wget ftp://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/m4/m4-1.4.17-1.aix6.1.ppc.rpm
+rpm -Uvh m4-1.4.17-1.aix6.1.ppc.rpm
+
+wget http://www.bullfreeware.com/download/bin/3555/gettext-0.19.8-1.aix6.1.ppc.rpm
+rpm -Uvh --nodeps gettext-0.19.8-1.aix6.1.ppc.rpm
```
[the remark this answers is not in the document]
sxa555: (Comment likely irrelevant with the switch to yum)

ansible/playbooks/aix.yml
```diff
@@ -17,8 +17,8 @@
       file: path=/tmp/playbook_tmp state=directory mode=0755
       tags: layout
 
-    - name: Create rpm tmp folder /tmp/playbook_tmp/rpms
-      file: path=/tmp/playbook_tmp/rpms state=directory mode=0755
+    - name: Create rpm tmp folder /tmp/playbook_tmp/pkgs
+      file: path=/tmp/playbook_tmp/pkgs state=directory mode=0755
       tags: layout
```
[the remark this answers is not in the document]
jdekonin: Agreed, I'll be reworking this section and other with checks to permit ansible tower execution more than once and not having the machine left in an unusable state.

ansible/playbooks/scripts/AIX_RPM_PKGs_Install.sh
jdekonin: Once I have the yum method functioning 100%, which I have very close, this will not be an issue.

ansible/playbooks/scripts/AIX_RPM_PKGs_Install.sh
```diff
@@ -7,65 +7,139 @@
...
+wget http://www.bullfreeware.com/download/bin/3232/gawk-4.1.4-2.aix6.1.ppc.rpm
+rpm -Uvh gawk-4.1.4-2.aix6.1.ppc.rpm
+
+wget http://www.bullfreeware.com/download/bin/3944/perl-5.24.0-3.aix6.1.ppc.rpm
+rpm -Uvh perl-5.24.0-3.aix6.1.ppc.rpm
+
+wget http://www.bullfreeware.com/download/bin/2536/autoconf-2.69-2.aix6.1.noarch.rpm
+rpm -Uvh autoconf-2.69-2.aix6.1.noarch.rpm
+
+wget http://www.bullfreeware.com/download/bin/3544/vim-common-7.4-2.aix6.1.ppc.rpm
```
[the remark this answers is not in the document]
jdekonin: vi on aix is not user friendly imo, adding vim was simply for the admin.  Its not a requirement for a build/test machine.

ansible/playbooks/aix.yml
```diff
@@ -6,125 +6,278 @@
...
+        dest: /tmp/openssl-1.0.2.1300.tar.gz
+      tags: openssl
+
+    - name: Extract Openssl 
+      shell: gzip -d /tmp/openssl-1.0.2.1300.tar.gz
+      tags: openssl
+    
+    - shell: tar -xvf /tmp/openssl-1.0.2.1300.tar -C /tmp/
+      tags: openssl
+
```
jdekonin: The previous 3 steps could be done in one "uncompress", like I do in other spots in the file, but there is no unzip installed by default, and its not installed until the yum step(s).

jdekonin:
I believe this is ready for review.  I have this working via ansible tower on AIX releases : 
- 7.1.3.[5|9]
- 7.1.4.[3|4]
- 7.2.0.[2|4]
- 7.2.1.[3|4]

ansible/playbooks/aix.yml
sxa555: Not sure why `unzip` is relevant here: This should do it for the last two steps at least: `gzip -cd /tmp/openssl-1.0.2.1300.tar.gz | tar xf - -C /tmp/`

ansible/playbooks/aix.yml
```diff
@@ -6,125 +6,282 @@
...
+  # Uninstall conflicting packages from base image   #
+  # if they were installed via rpm unless yum exists #
+  ####################################################
+    - stat:
+        path: /usr/bin/yum
+      register: yum
+
+    - debug:
+        msg: "Yum installed, skipping download and installation"
+      when: yum.stat.islnk is defined
```
bblondin: ``` when: yum.stat.islnk is defined ```

Is ``` yum ``` a symlink?
```

islnk | Tells you if the path is a symbolic link | success, path exists and user can read stats | boolean | False

```

Should it be ```yum.stat.exists == True ```

ansible/playbooks/aix.yml
```diff
@@ -6,125 +6,282 @@
...
-        src: /home/ubuntu/aix/java/Java8_64.jre.tar.gz
-        dest: /tmp/playbook_tmp/java/Java8_64.jre.tar.gz
+    - name: Install yum and dependencies
+      shell: /tmp/yum.sh
+      register: result.yum
+      ignore_errors: yes
+      when: yum.stat.islnk is not defined
+      tags: yum
+
+    - name: Ensure the yum package index is up to date
```
bblondin: This set of commands will update everything to latest not just updating the cache so, I think the title here should be changed

ansible/playbooks/aix.yml
```diff
@@ -260,24 +596,62 @@
...
+        fi
+      register: extendlv_result
+      changed_when: "'Extending' in extendlv_result.stdout"
+      notify:
+        - restart machine
+      tags: swap
+
+  ################
+  # Pass or Fail #
+  ################
```
bblondin: The Slack section here can be removed

ansible/playbooks/aix.yml
jdekonin: Technically, stat is on /usr/bin/yum which is a link to /opt/freeware/bin/yum, both of which are added with the yum.sh 'install'.

I think the method is valid, but I could also change to exists method if you feel strongly about it.

ansible/playbooks/aix.yml
jdekonin: Update made.

ansible/playbooks/aix.yml
bblondin: no, that's fine just wanted to confirm

ansible/playbooks/aix.yml
jdekonin: The unarchive method can transfer the file and unpack in one step instead of 3 (or 2 with your proposed change) but it depends on unzip.  I was simply meaning that I couldn't use the ansible preferred method as the warning when you run it in AWS states "Consider using unarchive module rather than running tar".  I simplified with your suggestion.

jdekonin:
I believe I have commented and/or updated to all responses.

gdams:
does this cover the creation of ramdisks?

jdekonin:
No it does not.  Those changes were not made in the previous version.  I'd suggest a follow on issue for adding that functionality.

gdams:
Can someone create a follow up issue for adding ramdisk?

PR #122, opened 2018-01-10
add ppcle machines

gdams:
After merging, this will need to be run again on jckservices


sxa555:
I've already run these rules with a modification on my local machine so shouldn't need to be run again at the moment.

PR #128, opened 2018-01-12
Add PPCLE JCK machines to infrastructure.md

sxa555:
Hostnames on the machines themselves don't have "ppcle" in them (They are jck-ubuntu-16-04-[12]) - I have left the table with the name as they are registered in jenkins.


gdams:
Is there a reason that the packet machine was marked as JCK only?

PR #130, opened 2018-01-12
Add fontconfig-devel on rhel for Java10+ builds

jdekonin:


jdekonin:
Tested against : 
- rhel6; ppc, x86 and 390
- rhel7; ppc, x86 and 390

PR #134, opened 2018-01-17
Ensure iptables changes are made permanently

sxa555:
Executes `iptables-save` after setting the firewall rules so they survive the machine being restarted


sxa555:
This didn't work - `iptables-save` isn't working as I had expected from the documentation I read.

gdams:
Should be a case of changing the command to be `/sbin/iptables-save > /etc/iptables/rules`

We should probably add `iptables-persistent` to the apt section as well just incase it doesn't exist on the system

PR #160, opened 2018-01-31
add symlinks for java 7 and java 8

gdams:
This is required for OpenJ9 windows builds.

FYI @vsebe


bblondin:
Should we also do the same for Java9?

gdams:
already added to the Java9 PR

jdekonin:
Why do we need symlinks?  Why not reference the sdk install directory directly?

Issue #89, opened 2017-12-11
Number of hosts require adding to Nagios monitoring


tellison:
We have brought a number of new machines on-line recently, but they need adding to the Nagios monitoring service.

e.g. 
build-win2012-1 (37.58.103.195)
build-win2012-2 (37.58.103.196)
jck-softlayer-x86-ubuntu-16.04-1 (159.122.210.205)
jck-softlayer-x86-ubuntu-16.04-1 (159.122.210.194)
build-cloudcone-x64-ubuntu-16.04-1 (??)
build-plymouth-x64-solaris11-1 (offline?)

possibly more

tellison:
P'haps the zOS machines can't play in this scheme?
build-marist-s390x-zOS-1
build-marist-s390x-zOS-2

tellison:
The host following host should be removed from Nagios
build-joyent-x64-ubuntu-16.04-1

tellison:
New host for JCK services.  WOuld be good to have Nagios check these keep working to avoid strange JCK failures.

jck-softlayer-x64-ubuntu-16.04-3 (aka `jckservices.adoptopenjdk.net`) (158.176.72.83)

geraintwjones:
As far as I can tell, build-joyent-x64-ubuntu-16.04-1 is not in Nagios. Did someone remove it?

geraintwjones:
> jck-softlayer-x86-ubuntu-16.04-1 (159.122.210.205)

Logged on to 159.122.210.205. It reports itself as `jck-softlayer-x64-ubuntu-1604-1`, so that's what I've added to nagios.

geraintwjones:
Added `jckservices.adoptopenjdk.net` (158.176.72.83)

geraintwjones:
Working through the list of servers documented in the wiki [here](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/infrastructure.md), adding those to Nagios if they haven't alteady been added.

karianna:
Is this something we could add to ansible bootstrapping for future boxes?

geraintwjones:
I was thinking about that. I think it is something we should be doing with Ansible. I'll take a look at what we've already got there tomorrow.

bblondin:
yes, last year I wrote a bash script for Nagios Admins to easily create machine templates.
Today I told another look at, I have rewrote it ```Nagios_Ansible_Config_tool.sh``` 
It is now working in my environment. 

[...]

PR #172, opened 2018-02-03
#89 Add Ansible playbook to set up  CentOS machine for Nagios monitoring

geraintwjones:
Create an Ansible playbook that sets up a CentOS machine ready to be
monitored from Nagios.

CentOS is derived from RedHat, so the playbook is a copy of the RedHat
playbook nagios_RedHat.yml, with some minor changes.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Nagios_Plugins/tasks/nagios_CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Nagios_Plugins/tasks/nagios_CentOS.yml
@@ -1,0 +1,129 @@
+#########################################################################################
+# AdoptOpenJDK - Ansible Playbook to install Nagios plugins on CentOS 7 on x86 hardware #
+#########################################################################################
+
+###########################################################################
+# License Information:                                                    #
+# Nagios core and its plugins are lincesed under GPL                      #
+# For more information please see: https://www.nagios.com/legal/licenses/ #
+###########################################################################
+
+  - debug: msg='Installing Nagios plugins'
+  ###############
+  # Nagios user #
+  ###############
+  - name: Create Nagios user
+    action: user name=nagios state=present
+    ignore_errors: yes
+    when:
+      - Nagios_Plugins == "Enabled"
+      - ansible_architecture == "x86_64"
+    tags: nagios_plugins
+
+  - name: Set authorized key for Nagios user
+    authorized_key:
+      user: nagios
+      state: present
+      key: "{{ lookup('file', '/Vendor_Files/keys/nagios.key') }}"
+    when:
+      - Nagios_Plugins == "Enabled"
+      - ansible_architecture == "x86_64"
+    tags: nagios_plugins
+
+  - name: Nagios user account policy - expire date
+    command: chage -E -1 nagios
+    when:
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Nagios_Plugins/tasks/nagios_CentOS.yml
```diff
@@ -0,0 +1,129 @@
...
+# License Information:                                                    #
+# Nagios core and its plugins are lincesed under GPL                      #
+# For more information please see: https://www.nagios.com/legal/licenses/ #
+###########################################################################
+
+  - debug: msg='Installing Nagios plugins'
+  ###############
+  # Nagios user #
+  ###############
+  - name: Create Nagios user
```
gdams: Can any of these tasks be extracted out into the [main.yml](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Nagios_Plugins/tasks/main.yml)? I imagine that many of these are the same commands for each Linux distro? Thoughts @geraintwjones @bblondin

geraintwjones: Possibly, assuming there are tasks that are common to all platforms. I'll wait to see @bblondin's thoughts, but I'd rather refactor this once we've got all platform playbooks written, and we know exactly what is common.

geraintwjones: I have compared the 4 existing Linux playbooks (one being the CentOS one that's part of this pull request) and have determined that there are only 4 common tasks, all of which are to do with creating a Nagios user.

These are...
- Set authorized key for Nagios  user
- Set Nagios user account policy - expire date
- Set Nagios user account policy - max days

So maybe we could move these out into main.yml.

However, I'd like @bblondin to confirm that the order in which the tasks are executed isn't important. In the CentOS and RHEL playbooks, creating the nagios user is the first thing done, but in SLES and Ubuntu it is not.

bblondin: I like this idea, it would clean things up. 
However at this stage I would like to have it working for all major OS first, then perform the clean up

bblondin: RHEL and CentOS will likely be identical

geraintwjones: Looks like I'm now able to squash and merge. However, since these are my changes, I feel that would be the wrong thing to do, that somebody else should do it.

geraintwjones: @gdams We're going to do te refactoring separately, under a different issue. Could you approve this one please?_

PR #180, opened 2018-02-07
Add FreeBSD support + rename Linux playbook to Unix playbook

gdams:
2 key changes
- Add FreeBSD support
- Rename `AdoptOpenJDK_Linux_Playbook` to `AdoptOpenJDK_Unix_Playbook`

The majority of the changes are minor:
- Ansible can't detect the number of CPU's on FreeBSD so I had to remove all of the `make -j` flags for FreeBSD.
- FreeBSD doesn't have a `root` group so I had to use the `wheel` group (perhaps this could be an arg)
- Ansible detects the FreeBSD arch as `amd64` instead of `x86_64`.
- By default FreeBSD doesn't enable account expiration so I removed the account expiry checks for FreeBSD.


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
```diff
@@ -12,6 +12,16 @@
   shell: git --version >/dev/null 2>&1
   ignore_errors: yes
   register: git_installed
+  when:
+    - ansible_distribution != "FreeBSD"
+  tags: git_source
+
+- name: Test if git is installed on path FreeBSD
+  shell: git --version >/dev/null
```
gdams: @bblondin `git --version >/dev/null 2>&1` didn't work on FreeBSD so I had to create a separate check that runs `git --version >/dev/null`. I wasn't sure if the latter command would work on all systems in which case we could just update the initial check command

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/FreeBSD.yml
```diff
@@ -0,0 +1,64 @@
...
+Test_Tool_Packages:
+  - apache-ant
+  - mercurial
+  - perl5
+  - xauth
+  - xorg
+  - pulseaudio
+  - p5-App-cpanminus
+
+crontab_Patching: "/usr/sbin/pkg upgrade"
```
bblondin: A '-y' flag is required for pkg upgrade and we should also be doing an update as well

/usr/sbin/pkg update && /usr/sbin/pkg upgrade -y

more info:
https://www.freebsd.org/cgi/man.cgi?query=pkg-upgrade&apropos=0&sektion=0&manpath=FreeBSD+10.0-RELEASE&arch=default&format=html

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
bblondin: This was for use of the conditions. Have you test all scenario
not install, install but wrong version, install and right version

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Swap_File/tasks/main.yml
```diff
@@ -50,36 +50,48 @@
 - name: Create swap file - via fallocate
   command: "fallocate -l 2G {{ swap_path }}"
   when:
-    - (swap_root.rc != 0 and swap_tmp.rc != 0) 
+    - (swap_root.rc != 0 and swap_tmp.rc != 0)
+    - ansible_distribution != "FreeBSD"
```
bblondin: it would be nice if the 'when' are the same
in this line ```- ansible_distribution != "FreeBSD"``` is on its own
but line 62 its combined

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Swap_File/tasks/main.yml
```diff
@@ -50,36 +50,48 @@
...
-  when: 
+  when:
     - (swap_root.rc != 0 and swap_tmp.rc != 0)
+    - ansible_distribution != "FreeBSD"
+  tags: swap_file
+
+- name: Set swap file permissions FreeBSD
+  file: path="{{ swap_path }}"
+        owner=root
+        group=wheel
```
bblondin: why 'wheel' ?

bblondin:
1) It appear that you have prevented FreeBSD from calling "Test_Tool_Packages"

2) Although FreeBSD doesn't supprt the "chage" command, there is another solution.
(root/zeus/Jenkins)

``` 
pw usermod -e dd-mmm-yy root
```

More info: https://www.freebsd.org/cgi/man.cgi?pw(8)


3) You appear to be skipping GIT_Source/tasks (for the most part)
as you are install git from the package manager.
We choose to compile from source on each platform to ensure the git version is at least at the minimum level. Is will only be an issue on FreeBSD if we use a lower OS version that comes with lower version of git.



4) Compiling with ``` make -j ```

Yes there is a bug in Ansible on FreeBSD, where is can't get the right number of cores.

Try this workaround:

```
# add this to /roles/Common/tasks/FreeBSD.yml

  - name: Get number of cores on FreeBSD
    shell: grep -c processor /proc/cpuinfo
    register: FreeBSD_processor_count


# Add this make -j {{ FreeBSD_processor_count }} everywhere you compile

# example:
  - name: Compile and install git from source
    shell: cd /tmp/git-2.15.0 && ./configure --prefix=/usr --without-tcltk && make clean && make -j {{ FreeBSD_processor_count }} && make install
    when:
      - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout | version_compare('2.15', operator='lt') )
      - ansible_distribution == "FreeBSD"
    tags: git_source

````

5)  ``` - (ansible_architecture == "x86_64" or ansible_architecture == "amd64") ```
why not override it at the beginning? /roles/Common/tasks/main.yml
this way you only have to change it once, and when the next guy adds just ansible_architecture == "x86_64" it won't stop if from running on amd64

```
- name: If ansible_distribution is amd64 set it to x86_64
  set_fact:
    ansible_distribution: "x86_64"
  when:
    - ansible_distribution == "amd64"

```

bblondin:
Also need to update the "Playbook Layout: (tree view)" in the README.md file

/AdoptOpenJDK_Unix_Playbook/README.md

gdams:
Merged, we will need to update AWX to point to Unix rather than Linux now

PR #187, opened 2018-02-08
Security - OS hardening and minor fixes

bblondin:
1) Security - Force key based authentication for ssh

- This will prevent all users (including root) from ssh'ing into a system using a password. While still allowing a password to be used from the console. This will increase our level of OS security.

```
    - PermitRootLogin without-password
    - PasswordAuthentication no
    - ChallengeResponseAuthentication no
```
2) Security - no-agent-forwarding + Keybox

- Adding keybox's ssh key to root's authorized_key file and removing the no-agent-forwarding lines in root's authorized_key file will ensure we have access to the machine, if it wasn't removed before running the playbook

3) Moved all users account policy's settings to Security role
(root jenkins nagios zeus)

- This machines the code clear
- Removed from the following roles
/roles/Common/tasks/main.yml
/roles/Jenkins_User/tasks/main.yml
/roles/Nagios_Plugins/tasks/main.yml
/roles/Superuser/tasks/main.yml

4) Nagios's ssh key should be a var

- The authorized_key for the nagios user was linked directly to the key file. Changed it to use a variable.

5) Vendor Variables

- Changed Vendor section to use variable

6) docker wget key

- Change to apt_key vs shell wegt

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
@@ -22,7 +22,6 @@
       - qstat
     when:
       - ansible_architecture == "x86_64"
-      - (ansible_distribution_major_version == "14" or ansible_distribution_major_version == "16")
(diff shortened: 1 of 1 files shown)


bblondin:
:( conflicts 

I'll have to rewrite

bblondin:
Closing will rewrite and create a new PR

PR #193, opened 2018-02-09
delete infrastructure.md in favour of inventory

gdams:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/inventory.yml is a much clearer way of viewing hosts now


geraintwjones:
I imagine a world where we won't have an inventory because jobs (builds, tests) will request macines on-demand from a cloud. I say "a" cloud and not "the" clud because the way I inderstand it we have a number of clouds, access donated by our sponsors, to choose from.

gdams:
@geraintwjones the list is still in github, just [here](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/inventory.yml) and in yml now

bblondin:
This is a great step forward, I really like the work done here. Great job.

PR #202, opened 2018-02-12
add linaro aarch64 machines

gdams:
We use box number as a jump box to number 2 (we only have 1 external IP). This also gives us a chance to fixup any bugs with centos7 on aarch64 which I don't believe that we have ever run the playbook against

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -15,7 +15,7 @@ hosts:
     - 1and1:
         win2008r2-x64-1: {ip: 109.228.48.187}
       
-    - linaro
+    - linaro:
         centos74-armv8-1: {ip: 64.28.99.122}
         centos74-armv8-2: {ip: 64.28.99.122, port: 2222}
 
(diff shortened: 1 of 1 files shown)


ansible/inventory.yml
```diff
@@ -14,6 +14,10 @@ hosts:
 
     - 1and1:
         win2008r2-x64-1: {ip: 109.228.48.187}
+      
+    - linaro
+        centos74-armv8-1: {ip: 64.28.99.122}
+        centos74-armv8-2: {ip: 64.28.99.122, port: 2222}
```
geraintwjones: @bblondin I think this new `port:` attribute needs to be added to the Nagios script, otherwise won't Nagios just complain that there are two machines that have the same IP address?

bblondin: Code changes are required to handle this: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/204

geraintwjones: What would happen if this change, adding port, was to be merged? Would something break?

PR #216, opened 2018-02-14
add Marcus Wornast

gdams:


geraintwjones:
I've added Marcus to KeyBox, and he's added his public key.

Issue #184, opened 2018-02-07
Install xvfb on headless (linux) machines 


smlambert:
1) install xvfb
Much like the jck machines have xvfb installed (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/ubuntu-jck.yml#L26), we need it to be installed on all machines.  

2) remove other x11 configuration
The other x11 server setup that is currently being done in the playbooks (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/x11/tasks/main.yml) is likely not needed, unless you know of someone else using them.

It is not required to start the display as the ubuntu-jck playbook does, as we can/should control that from the Jenkinsfile.  

More details on this approach (suggested by @tellison): https://github.com/AdoptOpenJDK/openjdk-tests/issues/49#issuecomment-312652929

sxa555:
Mac will require a different solution as I understand it. @gdams has my issue to do it for the JCK boxes (our first attempt didn't work properly) but if anyone else has experience of sensible ways of doing this they're welcome to volunteer ;-)

sxa555:
@smlambert Do the other X packages explicitly need to be removed from the existing boxes? Or are you just asking for them to be removed from the playbook as they're not required to your knowledge?

smlambert:
I do not believe they need to be removed from the playbook.  I mentioned it only as a code hygiene suggestion, since I am not aware of anyone who needs/depends on them.

geraintwjones:
What is the relationship between this issue and #217?

sxa555:
@geraintwjones Other than the query over whether it's necessary to remove any existing packages https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/217#issuecomment-365758060  there's no significant difference - the other one specifically references the playbook, I suppose you could consider this one "executing the playbook to resolve Shelley's issue" they're

geraintwjones:
I was going to add  xorg-x11-server-Xvfb to CentOS and RedHat. However, someone's beaten me to it on RedHat, so I'll just make the CentOS change now.

geraintwjones:
So all playbooks modified, and changes merged. Now need to run them.

karianna:
@geraintwjones Can this be closed now?

sxa555:
I reckon so, therefore closing :-)

PR #217, opened 2018-02-14
Add xvfb and acl to playbooks

jdekonin:
- package installs permit inclusion of an ubuntu machine to jck testing


gdams:
Do we still require https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Linux_Playbook/roles/x11/tasks/main.yml? I think that Shelley pointed out that the testing team did not require this on any of their machines

jdekonin:
This partially resolves #184; it does not address the X11 removal portion.

I am not sure if 'acl' is a requirement or not for a jck client machine.  Can you comment as to what requires this installed @sxa555?

sxa555:
Trying to remember what it was that required it now but I had something in the process fail without it. It's not strictly necessary to remove x11 though. To be honest I suspect xvfb will pull in a healthy number of the X11 packages anyway

PR #237, opened 2018-02-26
Add cmake for openj9/omr compilation

jdekonin:


jdekonin:
This isn't an immediate requirement, but in the near future cmake will be a requirement for building openj9 (and omr).

PR #241, opened 2018-02-27
add api setup playbook

gdams:
This PR adds a playbook to recreate the api server. I have also added a new section to the inventory called `service` which is where all of the service machines will be stored


gdams:
good shout, I'll change it

PR #250, opened 2018-03-03
RHEL 7 has issues with fallocate switched to dd

bblondin:
RHEL 7 has issues with fallocate switched to dd


ansible/inventory.yml
```diff
@@ -80,6 +80,10 @@ hosts:
 
     - scaleway:
         ubuntu1604-x64-1: {ip: 51.15.76.107}
+    
+    - softlayer:
```
gdams: weren't these already added in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/248

bblondin:
no longer needed

Issue #253, opened 2018-03-03
Nagios Plugin issue RHEL 7 x


bblondin:
There is something wrong with this selection of code, it gets skipped 
https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/main.yml

```
- name: Configure, make and make install nagios-plugins
  shell: cd /tmp/nagios-plugins-2.2.1/ && ./configure --prefix=/usr/local/nagios && make -j {{ ansible_processor_vcpus }} && make install
  become: yes
  when:
    - Nagios_Plugins == "Enabled"
    - ansible_distribution != "Ubuntu"
    - (ansible_distribution != "RedHat" and ansible_distribution_major_version != "7" and ansible_architecture != "ppc64le")
  tags: nagios_plugins
```

It then causes a failure later in the play due to the folder not being there

``` 
TASK [Nagios_Plugins : Download add-on check_yum plugin] ***********************
fatal: [test-softlayer-rhel7-x64-1]: FAILED! => {"changed": false, "failed": true, "msg": "Destination /usr/local/nagios/libexec does not exist"}```

bblondin:
Pull Request: #255

bblondin:
PR approved

PR #255, opened 2018-03-04
Nagios Plugin issue RHEL 7 x #253

bblondin:
This will address issue #253 
Moved the 'Install plugins' from main.yml (Nagios_Plugins) into nagios_{{ ansible_distribution }}.yml


bblondin:
I thinks its ok for now, but yes we could make improvements to it

PR #260, opened 2018-03-06
WIP: Create BUILDING.md

gdams:
An initial draft of a supported levels table.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,14 @@
+### Supported platforms
+
+The community does not build or test against end-of-life distributions (EoL).
+Thus we do not recommend that you use our binaries on end-of-life or unsupported
+platforms in production.
+
+|  System      | Version                          | Architectures        | Notes            |
+|--------------|----------------------------------|----------------------|------------------|
+| GNU/Linux    | kernel >= 2.6.32, glibc >= 2.12  | x64                  |                  |
+| macOS        | >= 10.10                         | x64                  |                  |
+| Windows      | >= Windows 7/2008 R2             | x86, x64             | vs2012           |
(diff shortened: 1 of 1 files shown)


jdekonin:
There are two other dimensions to this as well.; release and originator  There are platforms that OpenJDK lists as minimum supported and those that OpenJ9 lists as minimum supported. 

OpenJDK is published [here](https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms)
Not sure about OpenJ9, I will see if I can find something on their site and update with a url.

karianna:
See https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms

sxa555:
For completeness on this, there is also my initial stab at a supported platforms list at https://github.com/AdoptOpenJDK/openjdk-build/wiki/%5BWIP%5D-Minimum-OS-levels

karianna:
Yeah this should be in the openjdk-build repo and is OK as a wiki I think.  We should mostly just reference the official OpenJDK and OpenJ9 docs for each version and inly add our extra bits for platforms that OpenJDK doesn't cover (build instruction wise)

sxa555:
Sounds like the document I already have in there is the right place to flesh out any further detail, unless we want separate documents for build environments vs running our supplied binaries (My doc is the latter)
@gdams Where did those figures for kernel and glibc versions come from - did you just pull them from a particular OS?

sxa555:
This can probably now be considered superceded by the combination of the compiler table in the [new Build FAQ](https://github.com/AdoptOpenJDK/openjdk-build/pull/1457) and the wiki document referenced above

sxa555:
No further updates ... closing

PR #265, opened 2018-03-07
Added xorg-x11-server-Xvfb

geraintwjones:
Adding add-xvfb-to-oraclelinux because that's what's used on RHEL, which this o/s is based on.


geraintwjones:
Can someone review this please?

PR #273, opened 2018-03-08
Add jckservices to inventory.yml (and add to updating instructions)

sxa555:


ansible/inventory.yml
```diff
@@ -104,3 +105,8 @@ hosts:
     - softlayer:
         ubuntu1604-x64-1: {ip: 159.122.210.205}
         ubuntu1604-x64-2: {ip: 159.122.210.194}
+
+  - jckservices:
```
gdams: I think that this should be moved to `infrastructure:` as it's a generic machine hosting stuff that we use for jck testing

sxa555: Fair comment - done

PR #278, opened 2018-03-08
Jenkins node helper API

VermaSh:
The class NodeHelper partially addresses feature requests in AdoptOpenJDK/openjdk-infrastructure#259
- Add machines to Jenkins: addNewNode()
- Determine machine stats
	- getCpuCount(String computerName)
	- getLocation(String computerName)
	- getOsPatchLevel(String computerName)
	- getMemory(String computerName)
	- getSpaceLeftInGb(String computerName)
	- getTotalSpace(String computerName)

- Set machine description based on machine stats
	- Machine stats retrieved from the above functions are used for the description

- Determine machine capabilities
	- getEndian(String computerName)
	- getOsArch(String computerName)
	- getOsInfo(String computerName)

- Add labels to nodes based on machine capabilities
	- addLabel(String computerName, String label)
		- Overwrites the labels with the ones in the arguments
		- weekly maintaince job will call this function with labels from the inventory file plus labels returned from constructLabels(String computerName)
	- appendLabel(String computerName, String label)
		- does only an append

- Other useful functions
	- constructLabels(String computerName)
		- this function returns a string containing the default labels
		- as of right now, it only returns OS, arch and Endian (only if it is ppc running linux)
		- here is a list of labels which have been implemented so far:
			- constructOsLabel(String computerName)
			- constructCPULabel(String computerName)
			- constructMemoryLabel(String computerName)
			- constructEndianLabel(String computerName)
			- constructPlatformLabel(String computerName)
			- constructArchLabel(String computerName)
			- constructKernelLabel(String computerName)

- Few side notes:
	- computerName: is what the node name on Jenkins
	- getComputer(String computerName)
		- This was created with the intention to have computer name validations in one place.
		  At the moment it's very basic but it can be extended to enforce a standardized naming convention

### DIFF
--- a/src/NodeHelper.groovy
+++ b/src/NodeHelper.groovy
@@ -1,0 +1,1006 @@
+/* (C) Copyright IBM Corporation 2018.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+
+import jenkins.model.Jenkins;
+import hudson.model.Computer;
+import hudson.util.RemotingDiagnostics;
+import hudson.remoting.Channel;
+import hudson.slaves.DumbSlave;
+import hudson.slaves.CommandLauncher;
+import hudson.model.Slave;
+import hudson.model.Node.Mode;
+import hudson.plugins.sshslaves.SSHLauncher;
+
+class  NodeHelper {
+
+    // TODO: Move strings out to a config file
+
+    /* Java Web Start (Windows)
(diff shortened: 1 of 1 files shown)


gdams:
Thanks for this @VermaSh. There is a lot of code here and I feel that perhaps it would help to have a call with you and the rest of the infra team so you could talk us through/demo what this code can do and how to use it?

VermaSh:
@gdams Absolutely! I have a few jobs on Jenkins that I have been using to the API, I can walk the team through it. I just need a day to iron out a the kinks. 
How does Tuesday sound?

gdams:
our next weekly infra call is on Tuesday, 13 March 15:00 – 16:00 if you can join that? It will be at hangout.adoptopenjdk.net

VermaSh:
ahh. yes, I'll join that call

gdams:
does this need to be in the a `src` folder? I feel that renaming this to `jenkins-node-helper` may be clearer

AdamBrousseau:
It needs to be in the `src` folder at the root level in order to be used as a Jenkins [Shared Library] (https://jenkins.io/doc/book/pipeline/shared-libraries/)

gdams:
moving to https://github.com/AdoptOpenJDK/openjdk-jenkins-helper

karianna:
Documented in the TSC

PR #310, opened 2018-03-15
Aix playlist revisions

bblondin:
- Added `libiconv` as its required by `libunistring`
- Added `ulimit` settings to address `out of memory` issues in `cpan`
- Changed link to download `ant` to use apache's archive site as the main site will change
- Added title (name) to `stat` command
- `Uninstall conflicting packages` -- added `ignore_errors: True` so that it doesn't fail if none of these packages are installed
- After Perl 24 is installed, update symlink, this prevents failure from cpan when it finds perl 20

### DIFF
--- a/ansible/playbooks/aix.yml
+++ b/ansible/playbooks/aix.yml
@@ -140,7 +140,7 @@
         - http://www.oss4aix.org/download/RPMS/cmake/cmake-3.7.2-1.aix6.1.ppc.rpm
       tags: rpm_install
 
-    - name: Ensure /opt/freeware/bin is on the PATH
+    - name: Ensure perl from /opt/freeware/bin is the default in /usr/bin
       shell: mv /usr/bin/perl /usr/bin/perl.old && ln -s /opt/freeware/bin/perl /usr/bin/
       ignore_errors: True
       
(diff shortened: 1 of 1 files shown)


ansible/playbooks/aix.yml
```diff
@@ -136,12 +134,16 @@
...
         - http://www.oss4aix.org/download/RPMS/mktemp/mktemp-1.7-1.aix5.1.ppc.rpm
+        - http://www.bullfreeware.com/download/bin/2328/libiconv-1.14-22.aix6.1.ppc.rpm
         - http://www.bullfreeware.com/download/bin/2591/libunistring-0.9.6-2.aix6.1.ppc.rpm
         - http://www.bullfreeware.com/download/bin/3944/perl-5.24.0-3.aix6.1.ppc.rpm
         - http://www.oss4aix.org/download/RPMS/cmake/cmake-3.7.2-1.aix6.1.ppc.rpm
-      when: yum.stat.islnk is not defined
       tags: rpm_install
 
+    - name: Ensure /opt/freeware/bin is on the PATH
+      shell: mv /usr/bin/perl /usr/bin/perl.old && ln -s /opt/freeware/bin/perl /usr/bin/
```
jdekonin: Why not just remove and recreate?  And you missed something on the end of the 'ln' cmd.

bblondin: In case you still want the older version, however the title (name) is miss leading I'll update that

`ln -s /opt/freeware/bin/perl /usr/bin/` works fine, you only need to add a file name at then end if you should to change the name.

PR #316, opened 2018-03-19
Add freemarker role for unix playbooks

jdekonin:


bblondin:
Could we move the cleanup to the cleanup role?

jdekonin:
Cleanup moved to cleanup role.

jdekonin:
This is not the cause.  This adds freemarker to the /home/$jenkins_user location via ansible script.  When sourceforge was unavailable for a few days, I thought adding locally controlled version a good idea.  And better yet, since the file version never changes add it to the machine so as to not increase the time of a compile job with machine setup actions.

karianna:
@jdekonin You're totally correct and my apologies!

PR #329, opened 2018-03-26
Install Openjdk9 on the build machines (for bootstrapping JDK10)

sxa555:
Untested as yet - but throwing it up so that anyone who's got more experience than me on the current playbook structure can comment on it and see if there's anything I've missed before I test it ;-)

Fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/324


sxa555:
Adjusted to use the ansible `unarchive` task to do the download following suggestion from @pnstanton

sxa555:
> default('build:test:!*zos*:!*win*:!*macos*:!*aix*') }}" This takes care of the others, as the play wouldn't run on AIX MacOS Windows or z/OS
I'm not quite sure what you mean by this or where that's from - is there another change I need to make other than excluding the `armv7l` as per the start of your comment @bblondin

sxa555:
Done the update to exclude `armv7l`

PR #331, opened 2018-04-04
WIP: Adding pkg-config into the ansible playbooks

adamfarley:
This package is required on zLinux to enable OpenJDK to find
freetype without patching, as it changes the locations we search
for freetype to include the default location.

Tried contributing the patch so we search the *right* locations
for freetype, but OpenJDK has chosen not to accept this.

Signed-off-by: Adam Farley <adam.farley@uk.ibm.com>

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
@@ -31,6 +31,7 @@ Build_Tool_Packages:
   - openssl-devel
   - perl-CPAN
   - perl-devel
+  - pkgconfig
   - unzip
   - wget
   - xz
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
@@ -19,6 +19,7 @@ Build_Tool_Packages:
   - libelf1
   - make
   - ntp
+  - pkg-config
   - unzip
   - wget
   - zip


sxa555:
Are you sure those package names are correct? On my RHEL7 machine `pkg-config` is provided by the `pkgconfig` package:
```
[sxa@sxat470p playbooks]$ rpm -qf `which pkg-config`
pkgconfig-0.27.1-4.el7.x86_64
```

adamfarley:
Changing the title to WIP (work in progress). I'll run builds on SLES and RHEL to confirm we've got the right package name, and that it fixes the problem.

sxa555:
I don't think you want `pkgconfig.s390x` on the file for RHEL systems across all architectures. `pkgconfig` should be adequate

adamfarley:
On the zLinux RHEL machine I consulted, it said there were two packages listed for "yum list pkgconfig":
- pkgconfig.s390x (64-bit version, installed)
- pkgconfig.s390 (31-bit version, not installed)

Completely forgot that this playbook is used for all architectures. It's an odd configuration, where RHEL uses a different config file to SLES, but zLinux shares a file with xLinux.

I've verified that specifying "pkgconfig" in this context will cause RHEL to find the pkgconfig.s390x package.

adamfarley:
Thanks. :)

sxa555:
FYI looks like the package is already on the SLES build machine `build-marist-sles12-s390x-1` and the RHEL one is a bit sick at the moment (Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/323) so all the currently live machines should be good.

adamfarley:
Thanks sxa, the extension undo has been merged. Here's a build to make sure things work.

https://ci.adoptopenjdk.net/job/openjdk10_openj9_build_s390x_linux/20/

Job should go live inside the next 10 mins.

PR #354, opened 2018-05-01
Register Visual Studio DIA SDK shared libraries

keithc-ca:
OpenJ9 uses the ddrgen tool from OMR; that tool depends on msdia100.dll which must be registered. This change arranges that both the 32-bit and 64-bit variants are registered as part of installing Visual Studio 2010.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
@@ -20,6 +20,10 @@
   when: (vs2010_installed.stat.exists == false)
   tags: MSVS_2010
 
+- name: Register Visual Studio 2010 DIA SDK shared libraries
+  raw: 'for /r "C:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK" %F in ( msdia*.dll ) do regsvr32 /s "%~F"'
+  tags: MSVS_2010
+
 - name: Reboot machine after Visual Studio 2010 installation
   win_reboot:
     reboot_timeout: 1800
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2013/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2013/tasks/main.yml
@@ -23,6 +23,10 @@
   when: (vs2013_installed.stat.exists == false)
   tags: MSVS_2013
 
+- name: Register Visual Studio Community 2013 DIA SDK shared libraries
+  raw: 'for /r "C:\Program Files (x86)\Microsoft Visual Studio 12.0\DIA SDK" %F in ( msdia*.dll ) do regsvr32 /s "%~F"'
+  tags: MSVS_2013
+
 - name: Reboot machine after Visual Studio installation
   win_reboot:
     reboot_timeout: 1800


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
```diff
@@ -20,6 +20,10 @@
   when: (vs2010_installed.stat.exists == false)
   tags: MSVS_2010
 
+- name: Register Visual Studio 2010 DIA SDK shared libraries
+  raw: 'for /r "C:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK" %F in ( msdia*.dll ) do regsvr32 /s "%~F"'
+  tags: MSVS_2010
+
```
jdekonin: How will this impact a machine that also has Microsoft Visual Studio 12.0 installed?  That has its own version of DIA SDK, should those also be registered?  Or would this cause conflict for the same machine that uses both toolchains?

keithc-ca: VS12 has a distinct CLSID for the 'Debug Information Accessor' implementation so it won't interfere with the one for VS10.

Unfortunately, VS12 doesn't register its implementation so this PR will need to be amended.

keithc-ca:
Is there something I can do to help get this merged?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
```diff
@@ -20,6 +20,10 @@
   when: (vs2010_installed.stat.exists == false)
   tags: MSVS_2010
 
+- name: Register Visual Studio 2010 DIA SDK shared libraries
+  raw: 'for /r "C:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK" %F in ( msdia*.dll ) do regsvr32 /s "%~F"'
```
tellison: What does the wildcard pick up?  Isn't it sufficient to register `msdia100.dll` alone?

tellison:
Sorry @keithc-ca a number of folks are promoting adoptopenjdk at a conference this week.  Just looking now.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
keithc-ca: Without the wildcard, it will invoke regsvr32 in *every* subdirectory, not just those with the DLLs we seek to register.

tellison: Ok, so I was imprecise :-), I meant why the need for the loop over the wildcard match, i.e. why not just
`regsvr32 /s "c:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK\bin\amd64\msdia100.dll"`

tellison:
Figured we can marge this version and continue the discussion in parallel.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
keithc-ca: There are two DLLs and I think both should be registered. If you prefer, we could say:
```
regsvr32 /s "c:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK\bin\msdia100.dll"
regsvr32 /s "c:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK\bin\amd64\msdia100.dll"
```
but I wasn't sure whether/how you would express two commands in ansible.

PR #360, opened 2018-05-07
Updates for RHEL6

vsebe:
Fixes for gcc 4.8.5:
- install libmpc-devel package instead of building mpc from source
- install 32 bits glibc package
- add c++ symlink
- add libstdc++.so.6.0.19 symlink
Move dwarf and elf packages to build tools
Add ant-apache-regexp package to test tools

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
@@ -9,6 +9,7 @@ Build_Tool_Packages:
   - bind-utils
   - cpio
   - cups-devel
+  - elfutils-libelf-devel
   - expat-devel
   - fontconfig-devel
   - freetype-devel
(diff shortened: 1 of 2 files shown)


vsebe:
@karianna or @gdams could you please merge this PR, thank you!

karianna:
Hmmm, I've just realised we've missed a trick with these scripts and that's commenting *why* we need each dependency.  I'll open up an issue for this.

sxa555:
@vsebe What was the reason for the 32-bit glibc package being needed here? And what was failing without the symlink (This is potentiaily overwriting a default library). I haven't done that with the gcc7 PR which I committed today.

sxa555:
@karianna Did you create an issue to cover dependency docs? Trying to decide where the best place for that information is. Inline in the playbook? Wiki in this repo? Somewhere else?

karianna:
@sxa555 Hmm, it's hard to decide what to do here.  I'm not sure if we let the code be the documentation or to have it in a wiki.  Code is harder to scan through and summarise, but Wiki's bitrot.

vsebe:
@sxa555 - This replaced the [gcc prerequisites](https://gcc.gnu.org/install/prerequisites.html) built from source with available prerequisites packages (via package management system). 
In order to build GCC, the C standard library and headers must be present for all target variants for which target libraries will be built (and not only the variant of the host C++ compiler). 
The `--disable-multilib`  configure option could be used to build a 64-bit only compiler (not used in the playbook hence the 32bit library).  
gcc 4.8.5 is installed in /opt not in /usr, thus the symlink.

PR #362, opened 2018-05-08
Add CUDA installation to CentOS 6 and RHEL 6

vsebe:

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
@@ -15,7 +15,7 @@
     dest: "/tmp/cuda9_linux-run"
   when:
     - cuda_installed.stat.isdir is not defined
-    - (ansible_distribution == "Ubuntu" and ansible_distribution_major_version == "16") or (ansible_distribution == "RedHat" and ansible_distribution_major_version == "7") and (ansible_distribution == "centos" and ansible_distribution_major_version == "7" )
+    - (ansible_distribution == "Ubuntu" and ansible_distribution_major_version == "16") or (ansible_distribution == "RedHat" and (ansible_distribution_major_version == "6" or ansible_distribution_major_version == "7")) or (ansible_distribution == "CentOS" and (ansible_distribution_major_version == "6" or ansible_distribution_major_version == "7") )
     - ansible_architecture == "x86_64"
   tags: nvidia_cuda_toolkit
 
@@ -23,7 +23,7 @@
   shell: sh /tmp/cuda9_linux-run --silent --toolkit --override
   when:
     - cuda_installed.stat.isdir is not defined
-    - (ansible_distribution == "Ubuntu" and ansible_distribution_major_version == "16") or (ansible_distribution == "RedHat" and ansible_distribution_major_version == "7") and (ansible_distribution == "centos" and ansible_distribution_major_version == "7" )
+    - (ansible_distribution == "Ubuntu" and ansible_distribution_major_version == "16") or (ansible_distribution == "RedHat" and (ansible_distribution_major_version == "6" or ansible_distribution_major_version == "7")) or (ansible_distribution == "CentOS" and (ansible_distribution_major_version == "6" or ansible_distribution_major_version == "7") )
     - ansible_architecture == "x86_64"
   tags: nvidia_cuda_toolkit
 
@@ -60,7 +60,7 @@
 
 - name: Download NVidia CUDA Repo RPM - RHEL 7 on PPC64LE
   get_url:
-    url: http://developer.download.nvidia.com/compute/cuda/repos/rhel7/ppc64le/cuda-repo-rhel7-9.0.176-1.ppc64le.rpm
+    url: https://developer.download.nvidia.com/compute/cuda/repos/rhel7/ppc64le/cuda-repo-rhel7-9.0.176-1.ppc64le.rpm
     dest: /tmp/cuda-repo-rhel7-9.ppc64le.rpm
   when:
     - cuda_installed.stat.isdir is not defined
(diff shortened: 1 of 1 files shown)


vsebe:
@gdams please review, thx

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
```diff
@@ -11,19 +11,19 @@
 # RedHat 7 and Ubuntu 16 on x86_64
 - name: Download NVidia CUDA toolkit
   get_url:
-    url="https://developer.nvidia.com/compute/cuda/9.0/Prod/local_installers/cuda_9.0.176_384.81_linux-run"
-    dest="/tmp/cuda9_linux-run"
+    url: https://developer.nvidia.com/compute/cuda/9.0/Prod/local_installers/cuda_9.0.176_384.81_linux-run
```
tellison: Does OpenJ9 require this particular version?  The latest release as of 25-Jan-2018 is 9.1.85.
Same comment applies for different OSes listed below.

tellison: The Nvidia toolkit is not OSS, and [the license](https://docs.nvidia.com/cuda/eula/index.html) should be reviewed for suitability in AdoptOpenJDK usage.  My layman reading of Section 1.2.1 (x) is of a field of use restriction that is incompatible with our outgoing license; *but* that requires a proper legal interpretation - **don't take my word for it**.

vsebe: No it does not. OpenJ9 requires [CUDA 7.5](https://developer.nvidia.com/cuda-75-downloads-archive).
Is there an issue with the [7.5 license](http://developer.download.nvidia.com/compute/cuda/7.5/Prod/docs/sidebar/EULA.pdf).
If there are no legal issues then I'll downgrade the CUDA version.

keithc-ca: OMR currently allows use of versions 5.5 through 9.0, inclusive.

My reading of the JIT code in GpuHelpers.cpp [1] is that it supports versions 5.5 through 7.5 on Windows; newer versions might work on Linux (via a symbolic link named libcudart.so), but I don't know how much testing has occurred. @gita-omr may have more information to share.

I suggest we install and build with version 7.5 which yields JDKs that will be more widely usable (they will work with a system which has a runtime version 7.5 through 9.0 installed - with the above caveat for JIT); if built with 9.0 they will require runtime version 9.0 (versions 7.5, 8.0 will not be accepted by OMR) and may not work for JIT.

[1] https://github.com/eclipse/openj9/blob/72c97f3db39b765967856190c3f306cd3cce7ee7/runtime/compiler/trj9/runtime/GPUHelpers.cpp#L369

keithc-ca: Another consideration is that CUDA 8.0 and newer *don't* support Visual Studio 2010 which is used to compile JDK8 on Windows. Using CUDA 7.5 for all platforms, at least for consistency, seems sensible.

karianna:
@tellison Is this OK to go through now?

sxa555:
ping @tellison - this appears to be waiting on you based on the last comment

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
keithc-ca: OMR now supports up to version 9.2. CUDA 9.2 is the only version that is compatible with newer versions of gcc (e.g. 7).

sxa555: @keithc-ca I'm concerned that your last two messages are now in conflict - you recommended 7.5, but that one won't work when we build with gcc7. Does this mean your recommendation is to use 9.2 or higher now?

keithc-ca: Given the need to support gcc7, I don't see much choice at this point. I think we need to use CUDA 9.2 (and fix any JIT issues that arise from that). @gita-omr Care to comment?

karianna:
@vsebe Also has merge conflicts

vsebe:
@karianna - rebased, merge conflict fixed.
@tellison is this CUDA version still an issue?

tellison:
Thanks for checking back - I seem to recall I had questions about the CUDA usage in this context, and resulting builds; but it was a while ago now and I can't remember the details.
I'm happy to step out of the way.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
```diff
@@ -91,8 +91,8 @@
 # SLES
 - name: Download NVidia CUDA toolkit Repo RPM for SLES12 on x86_64
   get_url:
-    url="http://developer.download.nvidia.com/compute/cuda/repos/sles122/x86_64/cuda-repo-sles122-9.0.176-1.x86_64.rpm"
-    dest="/tmp/sles12_cuda9_repo.rpm"
+    url: http://developer.download.nvidia.com/compute/cuda/repos/sles122/x86_64/cuda-repo-sles122-9.0.176-1.x86_64.rpm
```
karianna: Just noticed this - can we go https?

vsebe: Fixed, ready for your review, thank you!

PR #390, opened 2018-06-26
Add YAMLlint phase to Travis build

inponomarev:
Hello,

I saw your call for help on Twitter (https://twitter.com/adoptopenjdk/status/1008636577885630465) and found out that although you have lots of Ansible code here, your build pipeline only does syntax check.

There are three great tools that I believe can be helpful for every Ansible code base (in the ascending order of complexity): `yamllint`, `ansible-lint` and `molecule`. Using `molecule` one can do fully-fledged automated testing for all the Ansible code, just like for Java code. Yamllint and Ansible-lint are just static analysis tools that unify the code and find out some problems.

In this PR, I
* added `yamllint` phase to Travis build
* fixed all the critical errors from `yamllint` (thought there are still many warnings to fix)

If you like this, I could also add Ansible-lint and fix all the Ansible-lint warnings (you have plenty of them, e. g. using `shell` module instead of `command` in a number of places)


karianna:
@inponomarev - do join us on adoptopenjdk.net/slack :-)

PR #393, opened 2018-06-28
JCK Softlayer Windows firewall rule

sxa555:
One of the windows JCK machines is not able to connect to jckservices.adoptopenjdk.net as there is no firewall hole for it. This change corrects that


sxa555:
The should all disappear when it's merged as there is only really one file difference - the other changes are already in the master branch

sxa555:
Grrr - the checks are failing but on an installation step before it's able to run the checks. Possibly related to https://github.com/AdoptOpenJDK/openjdk-build/issues/310

karianna:
@sxa555 Latest travis run picks up 3 errors in the linting (which is what failed the build) - I *think* you can ignore the warnings (won't fail the build) so if you just patch the whitespace for the 3 errors it should be OK

sxa555:
@karianna Maybe I'm going mad but those three indentation errors that it's showing (lines 12,20, 63 - shown below)  don't look like errors in the file as the indentation is what I'd expect it to be relative to the stuff around them- can you check that I'm not doing anything obviously stupid

Here are the lines:
```
      include_vars: variables/adoptopenjdk_variables.yml
      - ant
      - kadmin.local -q "addprinc -pw user1 user1/jckservices.adoptopenjdk.net@ADOPTOPENJDK_NET"
```

karianna:
Looking at the raw file I think 

line 12: `  - name: Load AdoptOpenJDKs variable file`?
line 20: `- iptables-persistent`
line 63: `- kadmin.local -q "addprinc -pw `pwgen -1` admin/admin@ADOPTOPENJDK_NET"`

Not that the linting errors make much more sense there....

Shall I try editing the PR directly and see if that gets past one of the errors?

sxa555:
Since the list of commits showing in this is a mess I'm superceding with https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/453

PR #411, opened 2018-07-19
Add coreutils to yum install package list

### LINKED ISSUE #410: Make readlink available on AIX build machines
openjdk11's configure looks for `readlink` - this is not available by default on AIX. We can install it with `yum install coreutils` but that requires the `mktemp` package to be removed. This will need to be done in the ansible playbooks

jdekonin:
* readlink is a new requirement for jdk11
* readlink is supplied by coreutils as is mktemp
* remove rpm installation of mktemp as it conflicts with coreutils
* Fixes #410

### DIFF
--- a/ansible/playbooks/aix.yml
+++ b/ansible/playbooks/aix.yml
@@ -111,6 +111,7 @@
           with_items:
             - autoconf
             - bison
+            - coreutils
             - cups-devel
             - cups-libs
             - flex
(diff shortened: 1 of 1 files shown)


jdekonin:
I have manually updated build-osuosl-ppc64-aix-71-1 and build-osuosl-ppc64-aix-71-2 since this playbook doesn't run well multiple times on the same system.

PR #422, opened 2018-08-02
Update Ansible script to install VS2017 for JDK11

ChengJin01:
The changes are to install VS2017 so as to compile
JDK11 on Windows.

Signed-off-by: CHENGJin <jincheng@ca.ibm.com>

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -42,5 +42,6 @@
     - ANT
     - MSVS_2010
     - MSVS_2013
+    - MSVS_2017
     - NVidia_Cuda_Toolkit
     - NTP_TIME
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
@@ -1,0 +1,38 @@
+---
+################################
+# Visual Studio Community 2017 #
+################################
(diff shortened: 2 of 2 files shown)


ChengJin01:
FYI:  @jdekonin

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
```diff
@@ -17,7 +17,7 @@
 # .\ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert
 # .\ConfigureRemotingForAnsible.ps1 -SkipNetworkProfileCheck
 
-- hosts: "{{ groups['Vendor_groups'] | default('build*win*:test*win*') }}"
+- hosts: all
```
jdekonin: Please undo this line.  This topic needs to be addressed in another issue which I'll open now.

ChengJin01: Already reverted.

ChengJin01:
> The download link could be changed to point directly at the download file, https://aka.ms/vs/15/release/vs_community.exe...

Already updated to the download link.

ChengJin01:
> I'm assuming that this visual studio will sit cleanly along the others? 

I am not sure what you meant. But VS2017 is installed in a separate directory different from 2010/2013.

> ...but are the spaces in windows paths going to cause a problem?

According to the command above   **win_command: 'regsvr32 /s "{{ item }}"'**,
it should work as the item is already quoted in there.

karianna:
Please note that Ansible tower will still need to run this to update our windows hosts.  

@gdams - this is on a cron now correct?

gdams:
not for windows!

PR #427, opened 2018-08-07
changes required to build OpenJFX on Linux

johanvos:


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Ubuntu.yml
```diff
@@ -25,17 +28,28 @@ Build_Tool_Packages:
...
   - libwww-perl
   - libx11-dev
   - libxext-dev
+  - libxml2-dev
   - libxrender-dev
+  - libxslt1-dev
   - libxt-dev
   - libxtst-dev
+  - libxxf86vm-dev
   - make
```
jdekonin: Whats the purpose of the additional packages?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
```diff
@@ -23,30 +23,30 @@
 
 - name: Download cmake
   get_url:
-    url: https://cmake.org/files/v3.7/cmake-3.7.2.tar.gz
-    dest: /tmp/cmake-3.7.2.tar.gz
+    url: https://cmake.org/files/v3.11/cmake-3.11.4.tar.gz
+    dest: /tmp/cmake-3.11.4.tar.gz
     mode: 0440
```
jdekonin: If there is going to be a newer version of cmake installed, should the older version be removed?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -27,7 +27,9 @@
     - x11
     - NTP_TIME
     - gcc_48
+    - gcc_64
```
jdekonin: Why gcc 6.4 and not something newer like 7.3+ which contains the Spector backports?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Ubuntu.yml
jdekonin: FIrst thing I looked at this morning....answer is obvious `build OpenJFX on Linux` as the pr clearly states.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
johanvos: Should be ok too, as long as it is 5 or higher. The gcc_64 yml however will install it from source, which takes ages. I think it would be better to install a binary (6.4 or 7.3+)?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
johanvos: I'm using the same version for OpenJDK and OpenJFX, so bumping the version (and removing the older) should work.

karianna: I'm not convinced of this - what versions of OpenJDK were you building?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
karianna: I think ansible lets you directly download a binary...

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
jdekonin: I don't believe OpenJDK has a dependency on cmake; OpenJ9 on the other hand will in the near future.  I think @dnakamura is who is implementing the use of cmake for OMR and OpenJ9.  Devin, should/would 3.11.4 work okay for OMR/OpenJ9?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
jdekonin: 7.3 is available on a number of releases, but not the older ones.  For example on x86 CentOS69 is used which will need to compile from source.  

I agree its slow, but its a one time setup cost.  I am not sure what the support timelines are for openjdk for supporting gcc7.3+, for OMR the support is there, for OpenJ9 is close from what I've 'heard', its for this reason I suggested 7.3.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
dnakamura: oh yeah. more than good enough. We try to be compatible with 3.4, because thats the stableish version we have on z/os.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
sxa555: I've got the PR in to install GCC 7.3 on the build boxes so this shouldn't be needed any more - can you remove from your PR please @johanvos 
Depending on architecture you'll find the gcc in either:
`/opt/rh/devtoolset-2/root/usr/bin/gcc` (LInux/x86-64)
`/usr/local/gcc/bin/gcc-7.3` (Linux/ppc64le & Linux/s390x)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Ubuntu.yml
sxa555: Can we separate these out into a separate OpenJFX section so we know which packages were added for that please?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
sxa555: I think I'm ok with this. @karianna do you have any remaining concerns?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_64/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
+---
```
sxa555: File no longer needed since we'll be using gcc 7 :-)

karianna:
Also needs rebasing.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
karianna: Should be OK

karianna:
@johanvos Can you rebase and update this PR please?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Ubuntu.yml
sxa555: Also we've moved our build machines to be CentOS/RHEL - is there a specific need for Ubuntu to build OpenJFX?

sxa555: (Addition to previous comment - according to https://wiki.openjdk.java.net/display/OpenJFX/Building+OpenJFX `yum install mercurial bison flex gperf ksh pkgconfig libpng12-devel libjpeg-devel libxml2-devel libxslt-devel systemd-devel glib2-devel  gtk2-devel libXtst-devel pango-devel freetype-devel` will do the trick)

sxa555:
Have resolved conflicts

sxa555:
@CJKwork Can you do a PR this for the CentOS/RHEL playbooks please as per the details in the link I mentioned on the OpenJFX wiki? Ieally create a new stanza in the Common `vars` directory for CentOS and RHEL with the packages required for OpenJFX and include installing those in the `roles/main.yml`

karianna:
Looks like ansible lint failed.

sxa555:
It seems to be tripping on the adoptopenjdk10 role which is no longer part of this PR. That role is almost a direct copy of the adoptopenjd9 one but with a line `skip_ansible_lint` in the `tags` section. I suspect this is another case where there's a smarter way of doing something, but I'll unify it with `adoptopenjdk9` for now under https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/555

sxa555:
I'm going to do a final test of this before merging

sxa555:
Looks to be running ok, so other than the fact this is only for Ubuntu and we now build on RHEL I'm happy to merge. Tagging @karianna @johnoliver for any final veto on merging (Will do so in the next couple of days if no objections)

PR #441, opened 2018-08-21
Add Ubuntu repo for GCC 7.3

jdekonin:


jdekonin:
I am unable to test against an Linux ARM distro, so I have purposely put this together to avoid installing on ARM.  If someone can verify that these packages are available for the ARM distro I can update.  Otherwise, this is the safe option.

PR #442, opened 2018-08-22
Install gcc-7.3 on CentOS 6 x86_64

vsebe:
- add tasks to install devtoolset-7 packages

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
@@ -50,6 +50,16 @@
     - ansible_distribution_major_version == "6"
   tags: build_tools
 
+- name: Install repos for devtoolset-7 on CentOS6 x86_64
+  yum:
+    name: centos-release-scl
+  state: installed
+  when:
+    - ansible_distribution == "CentOS"
+    - ansible_distribution_major_version == "6"
+    - ansible_architecture == "x86_64"
+  tags: build_tools
+
 ############################
 # Build Packages and tools #
 ############################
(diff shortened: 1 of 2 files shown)


vsebe:
@jdekonin please review

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
```diff
@@ -50,6 +50,16 @@
...
   tags: build_tools
 
+- name: Install repos for devtoolset-7 on CentOS6 x86_64
+  yum:
+    name: centos-release-scl
+  state: installed
+  when:
+    - ansible_distribution == "CentOS"
+    - ansible_architecture == "x86_64"
+    - ansible_distribution_major_version == "6"
```
jdekonin: This repo is specific to Cent6?  And x86?  Because the packages listed below will install for all Cent releases and architectures.

vsebe: As per [Developer Toolset 7](https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/) is specific to CentOS 6, 7 x86-64.
This should address CentOS 6 x86-64 only .

jdekonin:
Not merging until @sxa555 and/or @johanvos have a chance to comment.

sxa555:
I want to make sure we have a solution that's scalable to all architectures and can be easily used by people running the ansible scripts themselves before approving this. Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/438

sxa555:
Actually `gcc_73_compiler_x86_64` isn't defined anywhere as far as I can see so that's not going to work

sxa555:
I've superceded this by https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/451 since there was a missing definition and it was easy enough to merge a fixed version into what I did there.

PR #443, opened 2018-08-23
Fix conditional for autoconf version check 

vsebe:

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/autoconf/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/autoconf/tasks/main.yml
@@ -19,7 +19,7 @@
     force: no
     mode: 0755
   when:
-    - ((autoconf_version.rc == 0) and (autoconf_version.stdout | version_compare('2.69', operator='lt')))
+    - ((autoconf_version.stdout == '') or ((autoconf_version.stdout != '') and (autoconf_version.stdout is version('2.69', operator='lt', strict=True))))
     - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
     - ansible_distribution_major_version == "6"
     - ansible_architecture == "x86_64"
@@ -31,7 +31,7 @@
     dest: /tmp
     copy: False
   when:
-    - ((autoconf_version.rc == 0) and (autoconf_version.stdout | version_compare('2.69', operator='lt')))
+    - ((autoconf_version.stdout == '') or ((autoconf_version.stdout != '') and (autoconf_version.stdout is version('2.69', operator='lt', strict=True))))
     - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
     - ansible_distribution_major_version == "6"
     - ansible_architecture == "x86_64"
@@ -40,7 +40,7 @@
 - name: Build and install Autoconf v2.69
   shell: cd /tmp/autoconf-2.69 && ./configure && make && make install
   when:
-    - ((autoconf_version.rc == 0) and (autoconf_version.stdout | version_compare('2.69', operator='lt')))
+    - ((autoconf_version.stdout == '') or ((autoconf_version.stdout != '') and (autoconf_version.stdout is version('2.69', operator='lt', strict=True))))
     - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
     - ansible_distribution_major_version == "6"
     - ansible_architecture == "x86_64"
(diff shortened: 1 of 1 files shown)


vsebe:
@jdekonin please review

vsebe:
If autoconf is not installed then install it.
If autoconf is installed and is an older version then install.
If autoconf is installed and is the required version then noting to do.

jdekonin:
This does not check to see if an new version is installed, but I think that is a corner case that will not happen on new installation as default version is old, so I am okay with not having that check.

If autoconf is not 2.69 and the system default is installed it will be overwritten which then leaves that package in an uncertain state.  I think @sxa555 pointed that out in another issue, and that we should standardize on putting packages we compile and install into /usr/local/ then make sure this one is on the path first.  Can you confirm that I got that right @sxa555?

sxa555:
I would tentatively agree @jdekonin, although I'd be happy to have the base OS package removed, install into `/usr/local` and then symlink `/usr/bin/autoconf` to there. I just don't want to have to overwrite anything from an installed package.

Side note: although I'm curious as to how on earth our current machines got into the state they're in. It would appear we have a SuSE package that's installed and providing 2.69 and that's not in this role :-\

```
[jenkins@build-joyent-centos69-x64-1 ~]$ rpm -qi autoconf
Name        : autoconf                     Relocations: (not relocatable)
Version     : 2.69                              Vendor: obs://build.opensuse.org/home:monkeyiq
Release     : 12.2                          Build Date: Sat 20 Jul 2013 08:53:17 UTC
Install Date: Tue 29 May 2018 15:23:59 UTC      Build Host: build32
Group       : Development/Tools             Source RPM: autoconf-2.69-12.2.src.rpm
Size        : 2308411                          License: GPLv2+ and GFDL
Signature   : DSA/SHA1, Sat 20 Jul 2013 08:53:26 UTC, Key ID a8a0d77fdffafc43
URL         : http://www.gnu.org/software/autoconf/
Summary     : A GNU tool for automatically configuring source code
Description :
GNU's Autoconf is a tool for configuring source code and Makefiles.
Using Autoconf, programmers can create portable and configurable
packages, since the person building the package is allowed to
specify various configuration options.

You should install Autoconf if you are developing software and
would like to create shell scripts that configure your source code
packages. If you are installing Autoconf, you will also need to
install the GNU m4 package.

Note that the Autoconf package is not required for the end-user who
may be configuring software with an Autoconf-generated script;
Autoconf is only required for the generation of the scripts, not
their use.
[jenkins@build-joyent-centos69-x64-1 ~]$ 
```

sxa555:
From discussion in #infrastructure - @karianna and I are in agreement that of the following options, either 2 or 3 are the best.

1. Fix the playbooks as-is as per this PR (i.e. trying to install from source direct into /usr/bin, probably overwriting anything installed by an RPM which will invalidate the RPM DB)
2. Build from source into `/usr/local`, remove any installed autoconf RPM, and symlink into `/usr/bin/autoconf` (Or 2a - you could add `/usr/local/bin` to the PATH but that's slightly less desirable IMHO
3. Use the third party RPM referred to in issue 371 ... And possibly host it ourselves in case it disappears. (edited)

vsebe:
@jdekonin - wait with the review, still working on it.  
And yes, it should not be installed if a newer version is already installed.

1. If configured with default options, than `make install` adds the /usr/local/bin to the PATH (before /usr/bin). Thus no symlinks would be required.
Symlinks usually require more maintenance. 

2. There is also possible to install it in standard location: `/usr/bin`. It can be done by using the `--prefix` option: e.g ./configure --prefix=/usr ....`. This way the `make install` would install files in /usr/bin, /usr/lib, etc (uninstall the older autoconf before 2.69 is installed). Also, there would be no symlinks required.

Either solution will do. 
In general, any change in the way a tool is installed requires to be documented and usually is more work for us to maintain the playbooks.

vsebe:
Assuming there are no dependencies that also require to be updated (as is the case here), it is preferable to install a tool from a binary than an rpm (not always available for all platforms, not always free).

sxa555:
> not always available for all platforms
The RPM is cross-platform so that isn't a concern in this case.

sxa555:
> If configured with default options, than make install adds the /usr/local/bin to the PATH (before /usr/bin). Thus no symlinks would be required.
Can you clarify what's leading you to that conclusion? How is it adding it into the PATH? That would be a very strange thing for it to do. Personally I don't think an additional symlink would cause too much trouble and would make it simpler for users to reduce environment setup effort.

sxa555:
`--prefix=/usr` is my "solution 1" and I don't want to do that as it will overwrite anything already there and potentially cause confusion - third party software should ideally be clearly visible as such, especially if it's not from an RPM

jdekonin:
Option 2 from source or rpm, either is good by me.  But with a future inclusion of option 3.  I don't think that needs to block this PR though, unless there that is decided in the next few days.

vsebe:
This PR fixes only the conditional of the existing tasks.
If another installation solution is desired, another PR should be opened.

vsebe:
This change contains:
- conditional fix when autoconf is not installed on the machine (run unix playbooks on a new rhel6 machine)
- update for [version comparison](https://docs.ansible.com/ansible/2.6/user_guide/playbooks_tests.html) 
as  in ansible v2.5 version_compare was renamed to version

Issue #446, opened 2018-08-24
Rhel playbook tries to install i686 package on all architectures


sej-jackson:
As noted by @sxa555 in #396, the AdoptOpenJDK_Unix_Playbook fails for non-x86 Rhel systems with:
```
ok: [build-marist-rhel74-s390x-2] => (item=glibc-devel)
failed: [build-marist-rhel74-s390x-2] (item=glibc-devel.i686) => {"changed": false, "failed": true, "item": "glibc-devel.i686", "msg": "No package matching 'glibc-devel.i686' found available, installed or updated", "rc": 126, "results": ["No package matching 'glibc-devel.i686' found available, installed or updated"]}
```
because glibc-devel.i686 has been added to the common Build_Tool_Packages section rather than an x86_64 section.

PR #449, opened 2018-09-03
don't install glibc-devel.i686 on non-x86 architectures

sej-jackson:
The glibc-devel.i686 package shouldn't be installed in the common Build_Tool_Packages section because it doesn't apply to non-x86 architectures, so have moved it into a new Additional_Build_Tools_RHEL_x86_64 section for x86_64 only.

Fixes [issue #446](https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/446)


sxa555:
(NOTE: This problem was introduced in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/360)

Issue #492, opened 2018-09-12
New OpenJ9 compile requirement for x86: nasm


jdekonin:
This is specific to feature work surrounding https://github.com/eclipse/openj9/issues/2418

@nbhuiyan, can you please advise as to which version of nasm OpenJ9 requires?

nbhuiyan:
@jdekonin The version of nasm we'd like is the current latest release: v2.13.03

sxa555:
@nbhuiyan Does your earlier comment mean that we definitely cannot use the ones supplied with the OS (2.07 in this case) is definitely not an option?

nbhuiyan:
@sxa555 We MAY be able to use nasm v2.07, but I have never tested using that version. The oldest version I have successfully tried building and testing with is v2.11.08. The reason we are setting a newer version as a requirement is that one of the motivations behind making a switch to NASM was to be able to use newer instructions and features available through a modern assembler. NASM v2.07 is [almost a decade old](http://repo.or.cz/nasm.git/tag/6ac396e0714fa3b0b6c0fa476a8f7fea4a3221a3), and a lot has been [fixed/added](https://www.nasm.us/doc/nasmdocc.html) since then.

sxa555:
OK ... just a shame not to use the default on the distribution, but as long as it doesn't introduce any extra runtime dependency I'm happy with using it. Thanks

AdamBrousseau:
Resolved via #494 I believe.

PR #494, opened 2018-09-12
WIP: Add nasm role for x86_64

jdekonin:
* v2.13.03 compiled from sources
* Related to #492


jdekonin:
Marked as WIP as I need to verify : 
- this will work on various operating systems
- requires gcc-7+ to compile (at least 5.4 isn't good enough) so I need a better way to reference CC/CXX << any suggestions anyone?
- likely want to pull this source from jenkins/userContent/ correct @sxa555?

The project takes all of 15 seconds to compile and install, not sure its worth stashing the binary but maybe if that works around the GCC bullet above?

sxa555:
@jdekonin For now I'm ok with pulling it from the official site as you've done in this PR. If/when we get round to hosting more stuff ourselves we can do it all at once later.

sxa555:
Ref 7.3 I think we're going to need some variables set on the jenkins machine definitions to indicate the location of the alternative compiler. Either `CC73` and `CXX73` or a `CC73PATH`. Or possibly all 3 if binutils on CentOS6/RHEL6 would be a concern.

jdekonin:
fyi @AdamBrousseau

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -29,6 +29,7 @@
     - NTP_TIME
     - gcc_48
     - gcc_7
+    - {role: nasm, when: ansible_architecture == 'x86_64'}
```
sxa555: Is this intended to take effect on OSX machines as well?

AdamBrousseau: I think the answer is "not yet". We haven't tried it on OSX. That being said, I know this package is related to building OpenJ9 on OSX.

AdamBrousseau: Afaik only required on x86.

sxa555: To be fair, we currently have plenty of other stuff in the playbooks which won't run on macos/x64 so I'm not too concerned about this one (Will need a major cleanup at some point though!)

PR #500, opened 2018-09-14
ansible: source compile gmake on CentOS/RHEL on ppc64le

### LINKED ISSUE #499: gmake using 100% of CPU for extended periods on CentOS/ppc64le J9 builds
Need to resolve this as it's inhibiting an acceptable solution for #256. The problem does not occur on Ubuntu machines.

Also covered in https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/7

The gmake process has two `<defunct>` children - one `sh` and one `mkdir`.

According to @ymanton:

_Yeah, I think I saw that sort of thing. I attached `strace` to gmake and it was just endlessly `stat`ing object files, hard to tell if it was making forward progress, but my builds on x86 did eventually finish after a couple of hours, it was on ppc that it was running for half a day or something until I killed it_

sxa555:
Fixes #499

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -20,6 +20,7 @@
     - GIT_Source
     - CPAN
     - cmake
+    - gmake
     - Docker
     - NVidia_Cuda_Toolkit
     - Superuser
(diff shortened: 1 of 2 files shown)


sxa555:
Tested on a local machine and seems ok. There is a risk that if the system make package is updated then the `/usr/bin/gmake` symlink will be overwritten

sxa555:
@karianna If you mean the `goodmake_source` lines, most of our playbooks already have something similar e.g. https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_48/tasks/main.yml

sxa555:
Ah yes I had a duplicate tag within the first stanza

PR #501, opened 2018-09-14
Install git to /usr/local to avoid package manager collision

AdamBrousseau:
- Installing to /usr/bin/git seems to be
  causing issues when yum updates and
  overwrites this version back to default.
- Install to /usr/local/bin/git instead.
- This is on the PATH before /usr/bin
- Tested on rhel7 ppc64le.


AdamBrousseau:
@jdekonin and @sxa555 for review

sxa555:
I'm tempted to also suggest that we remove any default git package on the machine if we do this compilation ourselves to avoid any confusion.

AdamBrousseau:
>remove any default git

Probably a good idea. I can't say when I will have the cycles to try and add that. Limited Ansible experience combined with different commands for different OSs.

PR #502, opened 2018-09-16
Downgrade grep to 2.27

### LINKED ISSUE #384: Downgrade Grep to 2.27-2 on windows
jdk8u builds require `grep 2.27-2` but cygwin defaults to `grep 3.0`. There seems to be no easy way to programatically downgrade the package so this may take a bit of working out but ideally needs to be added to the playbook. Installed manually using the cygwin.exe gui application for now.

ali-ince:
This add a couple of steps to Cygwin role to download grep 2.27 and install it (overwrite since cygwin doesn't allow us to downgrade a package unattended).

Fixes #384.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Firefox/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Firefox/tasks/main.yml
@@ -5,7 +5,7 @@
 
 - name: Test if Firefox is already installed
   win_stat:
-    path: 'C:\Program Files (x86)\Mozilla Firefox\firefox.exe'
+    path: 'C:\Program Files\Mozilla Firefox\firefox.exe'
   register: firefox_installed
   tags: Firefox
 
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Firefox/tasks/main.yml
```diff
@@ -5,7 +5,7 @@
 
 - name: Test if Firefox is already installed
   win_stat:
-    path: 'C:\Program Files (x86)\Mozilla Firefox\firefox.exe'
+    path: 'C:\Program Files\Mozilla Firefox\firefox.exe'
```
karianna: Are we sure about this change?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -55,6 +55,17 @@
     - xterm
   tags: cygwin
 
+- name: Download grep-2.27-2
```
karianna: Would be good to add a comment as to why we are using this particular version and warn folks not to upgrade it.

Issue #493, opened 2018-09-12
New OpenJ9 requirement: protobuf and openssl


jdekonin:
This is specific to feature work surrounding eclipse/openj9#2726

@mpirvu @harryyu1994, can you please advise as to which version of protobuf (3.5.1) and openssl OpenJ9 requires?  What platform(s) do these packages need to be rolled out too?

jdekonin:
fyi @vsebe

vsebe:
@mpirvu 
What platforms?
What protobuf package - see https://github.com/protocolbuffers/protobuf/releases/tag/v3.5.1?

mpirvu:
Forgot to mention: we need the cpp version protobuf-cpp-3.5.1.zip

sxa555:
@mpirvu My understanding from the changes going into OpenJ9 is that the build would be downloading an appropriate version of OpenSSL itself. Has that changed since Monday?

mpirvu:
For the time being all our testing has been done on LinuxAMD64. The vast majority of the code to be contributed is platform independent, so in theory with little work this could be extended to other platforms where protobuf is available/compilable. However, we decided to improve the LinuxAMD64 solution as much as possible before moving to other platforms.

For OpenSSL it is the same story as protobuf: building the JITaaS project needs the header files for OpenSSL. Running/testing it needs the OpenSSL shared library. If the test infrastructure can automatically download a development version or OpenSSL (that includes headers), then build JITaaS project and then run the tests, it's perfectly fine with me.

jdekonin:
As small as the openssl package might be, I don't think we want the same package downloaded and compiled every time JITaaS is compiled and tested.  Unless the version of openssl is going to be changing frequently?

@vsebe has taken over what I started and has been working on something to review with you @mpirvu

sxa555:
Looks like `automake` is a prereq for building `protobuf` from source so that will be needed as a prereq too

sxa555:
protobuf 3.5.1 doesn't seem happy when used with the autoconf 2.69 supplied with RHEL6:
```
+ autoreconf -f -i -Wall,no-obsolete
configure.ac:30: error: possibly undefined macro: AC_PROG_LIBTOOL
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1
```
Has anyone following this issue built it from source?

vsebe:
[...]

PR #510, opened 2018-09-18
Add Protobuf and OpenSSL 1.1.0g to Linux x86

vsebe:
Required by OpenJ9 for JITasS (out-of-process JIT compilation)

Issue: #493

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
@@ -1,0 +1,64 @@
+---
+###################
+# OpenSSL v1.1.0g #
+###################
+# Required by OpenJ9 for out-of-process JIT compilation (aka JITaaS)
+# Currently only used by the alternate openj9 branch at https://github.com/eclipse/openj9/tree/jitaas
+
+- name: Test if OpenSSL v1.1.0g is installed
+  shell: openssl version | awk '{print $2}' | cut -c 1-5
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
+    - ansible_architecture == "x86_64"
+  register: openssl_version
+  tags: openssl
+
+- name: Download OpenSSL v1.1.0g
+  get_url:
+    url: https://www.openssl.org/source/old/1.1.0/openssl-1.1.0g.tar.gz
+    dest: /tmp/openssl-1.1.0g.tar.gz
+    force: no
+    mode: 0755
+    validate_certs: no
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
+    - ansible_architecture == "x86_64"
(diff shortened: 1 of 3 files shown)


sxa555:
@vsebe Just to check - is there a reason why you only included RHEL/CentOS6 and Ubu 16.04 in these playbooks, so it won't get picked up on RHEL7/SLES/Later Ubuntu?

Other than that I'm happy with integrating this (Got someone who might test this for me tomorrow :-) )

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -0,0 +1,109 @@
+---
+###################
+# OpenSSL v1.1.0g #
+###################
+# Required by OpenJ9 for out-of-process JIT compilation aka JITasS
```
sxa555: Minor point but should this have been `JITaaS` rather than `JITasS`?

vsebe: right

vsebe:
I haven't yet test it on all platforms (waiting for testing). I'll extend it to all xlinux platforms.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
vsebe: fixed

vsebe:
@sxa555 - extended to centos, rhel, ubuntu and sles

vsebe:
@sxa555 - ready for your review, thank you!
I have tested the playbooks on ub 14 & 16, centos 6, rhel 6 & 7, sles 12.

sxa555:
I'd rather not have openssl 110 installed directly into `/usr/local` since this may not be a permanent requirement and I want to be easily able to delete it later, so can we use `./config --prefix=/usr/local/openssl110` instead please? I'd also prefer not to have the symlink into `/usr/local` since I want absolutely zero risk of anything picking this up by default.

I'm ok with protobuf being in the default location though :-)

sxa555:
FYI openssl 110g compiles ok on Linux for `ppc64le` and `s390x` too

vsebe:
@sxa555 - if opnessl is installed in a custom location (e.g /usr/local/openssl110), this location would have to be added to the PATH for the OpenJ9 for JITaaS builds. As there are multiple openssl installations this would have to be 1st on the PATH. Removing the symlinks would require to add /usr/local/openssl110/lib64 or /usr/local/openssl110/lib to LD_LIBRARY_PATH.

sxa555:
Yep that sounds reasonable to me

vsebe:
@sxa555 - added --prefix configure option and removed symlinks - ready for your review, thank you!

sxa555:
Just one more things based on some extra understanding obtained today:

(1) Since I've just realised this won't be in the mainstream codebase for some time can you add a comment at the start of the two new files saying something like `Currently only available in the alternate openj9 branch at https://github.com/eclipse/openj9/tree/jitaas` to make it clear why it's in there but not currently used by the mainstream build. 

Tempted to suggest renaming openssl to openssl110g just to be clear, but I'm ok to approve without that since at the moment we're not installing another version (111 was a possibility, but we're bundling that instead now)

vsebe:
@sxa555 - files amended (added requested comment) - ready for your review, thank you!

sxa555:
@vsebe Travis check has failed due to trailing whitespace on the line I asked you to add - can you fix please?

PR #514, opened 2018-09-21
Update Ansible script to install Clang on Windows

ChengJin01:
The changes are to install LLVM/Clang so as to 
compile JDK 8 & JDK 11 on Windows.

Signed-off-by: CHENGJin <jincheng@ca.ibm.com>

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -47,3 +47,5 @@
     - NTP_TIME
     - 7-Zip
     - MinGW-W64
+    - Clang_64bit
+    - Clang_32bit
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_32bit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_32bit/tasks/main.yml
@@ -19,7 +19,7 @@
   tags: clang_32bit
 
 - name: Install Clang 32bit
-  win_command: 'c:\temp\LLVM-7.0.0-win32.exe'
+  win_command: 'c:\temp\LLVM-7.0.0-win32.exe /S'
   when: clang_32bit_installed.stat.exists == false
   tags: clang_32bit
 
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_32bit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_32bit/tasks/main.yml
@@ -3,6 +3,8 @@
 # LLVM/Clang 32bit #
 ####################
 # Clang 32bit is an OpenJ9 prerequisite.
+# Silent installer for 32-bit will remove the 64-bit version so
+# we install from a zip file
 
 - name: Test if Clang 32bit is already installed (required by OpenJ9)
   win_stat:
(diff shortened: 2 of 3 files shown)


ChengJin01:
Reviewer: @jdekonin

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+
+- name: Test if Clang is already installed (required by OpenJ9)
+  win_stat:
+    path: 'C:\Program Files\LLVM\bin\clang.exe'
+  register: clang_installed
+  tags: clang
+
+- name: Download Clang
+  win_get_url:
+    url: http://releases.llvm.org/6.0.1/LLVM-6.0.1-win64.exe
```
jdekonin: Please change this to use the https url.

sxa555:
@ChengJin01 For reference when someone comes sees this in the future, can you link back to an issue that covers why this is needed please, as there is no current failure relating to this.

ChengJin01:
@sxa555 , this was originally requested from an internal work item which aims to support /Spectre option on Windows (we have to replace mingw with clang as mingw compiled build crashed for unknown reason after specifying the /Spectre option).

Here's the PR related to the work at https://github.com/eclipse/openj9/pull/2979, which is still in process.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang/tasks/main.yml
ChengJin01: Already fixed.

ChengJin01:
@jdekonin , everything already got updated (32bit and 64bit) here for review again.

The change here need to be merged first to ensure it works good on these build machines before we safely merge the changes with change at https://github.com/eclipse/openj9/pull/2979.

ChengJin01:
Given that Clang 32bit and 64bit share the same executable name: clang.exe,   it needs to export the correct bin path when compiling the build in cygwin
e.g. export PATH="C:\Program Files (x86)\LLVM\bin:$PATH  for 32bit
       export  PATH="C:\Program Files\LLVM\bin:$PATH"   for 64bit

jdekonin:
Seems there is no silent mode or automated command line to install LLVM.  This is problematic.

Suggestions on how to proceed @sxa555?

Manual intervention is required.  The install triggers a popup that asks a few questions (one agreement that will may require review).  I chose all the defaults for 32bit.  When running 64bit it also asks to uninstall the existing, which I say no too and then proceed to install with defaults.

sxa555:
If there is no silent install then we can't install it via the playbooks

karianna:
That seems odd, can we ask the clang maintainers for an option here?

jdekonin:
https://bugs.llvm.org/show_bug.cgi?id=24310 - closed as `RESOLVED WONTFIX`

This isn't the first windows requirement that cannot be installed via ansible.  Fix packs on JDK8 toolchain requirements come to mind.  Is there a location for steps that need to be performed manually for any given platform?

karianna:
So this needs a one-off intervention in the UI when we run this?

jdekonin:
Sort of, you need to chose the default options at 4 or 5 screens.  Its a 30 second install, but not convenient or unattended at all. 

This is why I ask if there is location for doc for steps that need to be performed manually for any given platform.

sxa555:
> Fix packs on JDK8 toolchain requirements

Do you have the details @jdekonin

sxa555:
I'm going to check whether we can install from a zip file created from the installed directory. Will likely depend on whether there are any significant registry entries set by the installer.

sxa555:
As per the issue referenced on the LLVM project, the `/S` option appears to work. We will need to verify whether the installer gets confused if both 64-bit and 32-bit versions are installed on the same machine (Invoking it manually did suggest it tried to prompt to uninstall the "previous" version)

sxa555:
Installing both versions silently appears to work ok from my testing. Have added the `/S` into this PR

jdekonin:
Ah, I didn't see that.  Awesome, I really didn't want to have manual steps for this.

sxa555:
Rather infuriatingly when triggered with `/S` it uninstalls the other "bitness" version that's already installed ...

sxa555:
Going back to liking the idea of installing from a `.zip` file again ...

sxa555:
I have `zip` files of both available if we don't find another solution by the time we need to go live on this

sxa555:
Now installed on both softlayer build machines (The ones tagged for J9)

sxa555:
I think installing 64-bit using the `/S` parameter and unzipping the 32-bit one is going to be the sane route to take going forward

karianna:
@sxa555 conflicts.

sxa555:
Conflicts resolved

karianna:
Is this still a WIP or?

sxa555:
Still a WIP as it'll need to be adjusted to install the 64 and 32 bit ones separately. I've got a 32-bit zipball at https://ci.adoptopenjdk.net/userContent/llvm700/llvm-7.0.0-win32.zip but playbook will need changed to download/extract that instead of running the 32-bit installer (Although needs to be verified as I got a rather odd directory name out of it when I tried to unzip that earlier this week ...)

jdekonin:
Make or makefiles have issues with spaces or special characters in the path.

Creating links fixes this issue.
mklink /D "C:\openjdk\LLVM64\" "C:\Program Files\LLVM\"
mklink /D "C:\openjdk\LLVM32\" "C:\Program Files (x86)\LLVM\"

sxa555:
@ChengJin01 Are you still owning this and able to address the last two comments or do you need me to take it over?

ChengJin01:
@sxa555 , please go ahead to take it over as there is nothing left on my side in terms of script changes.

PR #541, opened 2018-10-19
Add opensuse repo for git installation through package manager

jdekonin:


jdekonin:
I don't see much for SLES wrt 390.  Not sure if 390 and SLES is a combination anyone has tried as of yet??

The git 2.15 compiled from source in the `GIT_Source` role doesn't have all the dependencies.  Clones using `https` links fails.  The same general location would work for 390 as well as sles11, if those are wanted in the future.

Marked as WIP as I need to test and verify that this works.

karianna:
@jdekonin Safe to merge now?

jdekonin:
Yes this is good to go.

karianna:
@sxa555 I'll wait for your approval

PR #545, opened 2018-10-23
Make java7/8 sdk identifier generic on Cent/Redhat

jdekonin:
* intel - installs java7 to /usr/lib/jvm/java-1.7.0-openjdk.x86_64
* amd - installs java7 to /usr/lib/jvm/java-1.7.0-openjdk.amd64

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
@@ -167,3 +167,30 @@
   when:
     - ansible_architecture == "x86_64"
   tags: expat
+
+########
+# Java #
+########
+- name: Check if java8 directory exists
+  shell: ls -d /usr/lib/jvm/java-1.8.0-openjdk.* >/dev/null 2>&1
+  ignore_errors: yes
+  register: java8
+  tags: java
+
+- name: Create /usr/lib/jvm/java-1.8.0-openjdk symlink
+  shell: ln -s /usr/lib/jvm/java-1.8.0-openjdk.* /usr/lib/jvm/java-1.8.0-openjdk
+  when:
+    - java8.rc == 0
+  tags: java
+
+- name: Check if java7 directory exists
+  shell: ls -d /usr/lib/jvm/java-1.7.0-openjdk.* >/dev/null 2>&1
+  ignore_errors: yes
+  register: java7
+  tags: java
+
+- name: Create /usr/lib/jvm/java-1.7.0-openjdk symlink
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
```diff
@@ -158,3 +158,42 @@
   when:
     - ansible_architecture == "x86_64"
   tags: expat
+
+########
+# Java #
+########
+- name: Check if x86_64 java8 directory exists
```
vsebe: Instead of checking for both amd64 and x86-64 packages for  java 7 & 8 , couldn't you just check for if a package is available (repeat for java 7 and 8)?
Something like this:
```
- name: Check if java8 directory exists
   shell:  ls -d /usr/lib/jvm/java-1.8.0-openjdk.* >/dev/null 2>&1
   ignore_errors: yes
   register: java8
   tags: java
```
if `java8.rc == 0` then create symlink.

jdekonin:
I removed the portion of the changes that modified the alternatives for java.

This change is to make the sdk reference common across amd and intel.  By adding a symlink, build scripts can simply reference `/usr/lib/jvm/java-1.8.0-openjdk` regardless if the underlying hardware is amd or intel.  

- intel - installs java7 to /usr/lib/jvm/java-1.7.0-openjdk.x86_64
- amd - installs java7 to /usr/lib/jvm/java-1.7.0-openjdk.amd64

sxa555:
> Regardless if the underlying hardware is amd or intel.

Whoa really? Is that the zulu packages doing that? That's realy surprising - Didn't think anyone would do detection on that any more!

jdekonin:
java-1.7.0-openjdk-devel and java-1.8.0-openjdk-devel, ya I thought that was odd too.  I didn't check zulu, as I don't use that for my Java9 bootjdk, I use the adopt-openj9 one installed.

sxa555:
I suppose another option would be to force the `x86_64` one regardless of whether it's an AMD64/Intel64 system ... I would imagine they're compatible but have different optimisations

sxa555:
@jdekonin I've just had a look on an AMD64 box with RHEL7.5 and the installed openjdk has gone into a directory suffixed `x86_64` ... Which distro is it that's got the `amd64` suffix in the default repos?

jdekonin:
I've actually been attempting to generate a new instance one to prove.  But I haven't been able to, the cloud we use is only providing me x86_64, not amd64.  

My notes said cent/rh6, which now I am starting to doubt, I don't have an actual rh6 and amd non-virtual installs to check so I am at bit confused here.

I think the best path forward is to close and revisit, if/when I can prove the situation otherwise its a change for no purpose.

Issue #467, opened 2018-09-04
RHEL7 ppc64le NVidia_Cuda_Toolkit DKMS download failure


sej-jackson:
```
TASK [NVidia_Cuda_Toolkit : Download DKMS - RHEL 7 PPC64LE] ********************
fatal: [rtsvt-rhel74-ppc64le-1.fyre.ibm.com]: FAILED! => {"changed": false, "dest": "/tmp/dkms.noarch.rpm", "msg": "Request failed", "response": "HTTP Error 404: Not Found", "state": "absent", "status_code": 404, "url": "https://dl.fedoraproject.org/pub/epel/7/ppc64le/Packages/d/dkms-2.4.0-1.20170926git959bd74.el7.noarch.rpm"}
```
Checked `https://dl.fedoraproject.org/pub/epel/7/ppc64le/Packages/d/` to find that `dkms-2.6.1-1.el7.noarch.rpm` exists, but not `dkms-2.4.0-1.20170926git959bd74.el7.noarch.rpm`

sxa555:
Relevant line: https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml#L40

sxa555:
(Also, for reference, this PR is trying to add CUDA to RHEL6/x86-64 systems as well: https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/362)

keithc-ca:
I just tested this on a new system:
```
# cat redhat-release 
Red Hat Enterprise Linux Server release 6.10 (Santiago)
```
These two commands install the CUDA toolkit (`--override` can probably be omitted if gcc is installed first):
```
# wget https://developer.nvidia.com/compute/cuda/9.2/Prod2/local_installers/cuda_9.2.148_396.37_linux
# bash cuda_9.2.148_396.37_linux --silent --toolkit --override
```

sxa555:
In light of the above I'm happy to remove the DKMS package from the playbooks unless we find a subsequent reason for reinstating them.

sxa555:
@CJKwork One for you :-)

CJKwork:
Did some testing, and in playbook 
[...]

PR #548, opened 2018-10-29
Ansible: Removed the install DKMS steps, and updated the Install NVid…

CJKwork:
…ia CUDA toolkit step to install cuda-toolkit-9-0

Fix for issue https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/467

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
@@ -37,27 +37,6 @@
     - ansible_architecture == "ppc64le"
   tags: nvidia_cuda_toolkit
 
-- name: Download DKMS - RHEL 7 PPC64LE
-  get_url:
-    url: https://dl.fedoraproject.org/pub/epel/7/ppc64le/Packages/d/dkms-2.4.0-1.20170926git959bd74.el7.noarch.rpm
-    dest: /tmp/dkms.noarch.rpm
-  when:
-    - cuda_installed.stat.isdir is not defined
-    - ansible_distribution == "RedHat"
-    - ansible_distribution_major_version == "7"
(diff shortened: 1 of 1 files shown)


sxa555:
I worked with Clive to adjust the commit message :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
```diff
@@ -37,27 +37,6 @@
...
-  when:
-    - cuda_installed.stat.isdir is not defined
-    - ansible_distribution == "RedHat"
-    - ansible_distribution_major_version == "7"
-    - ansible_architecture == "ppc64le"
-  tags: nvidia_cuda_toolkit
-
 - name: Download NVidia CUDA Repo RPM - RHEL 7 on PPC64LE
   get_url:
     url: https://developer.download.nvidia.com/compute/cuda/repos/rhel7/ppc64le/cuda-repo-rhel7-9.0.176-1.ppc64le.rpm
```
jdekonin: Walking through the issue to #467, I see @keithc-ca mentions v9.2.148 where this installs v9.0.176.  @keithc-ca is there an OpenJ9 support statement for the version of cuda on PPC64LE?

sxa555:
Cheers for checking up on that one @jdekonin :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
keithc-ca: We will need to update to 9.2 to get a version that is compatible with gcc 7. CUDA 9.1 `crt/host_config.h` says:
```c
#if defined(__GNUC__)
#if __GNUC__ > 6
#error -- unsupported GNU version! gcc versions later than 6 are not supported!
```
Older versions have even lower limits on the version of gcc.

PR #552, opened 2018-10-30
Add libcurl-devel package for CentOS

vsebe:
The library is used by git-http-fetch, git-fetch.


sxa555:
@vsebe Is this really required by git fetch operations? I'd have thought that if it was a hard dependency it would be causing us problems prior to now. What were the symptoms?

PR #555, opened 2018-11-01
Adoptopenjdk10 lint

sxa555:
Should fix the linter errors affecting PR427

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk10/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk10/tasks/main.yml
@@ -21,7 +21,9 @@
   shell: ls -ld /usr/lib/jvm/jdk-10* >/dev/null 2>&1
   ignore_errors: yes
   register: adoptopenjdk10_installed
-  tags: adoptopenjdk10
+  tags:
+    - adoptopenjdk10
+    - skip_ansible_lint
 
 - name: Install latest release if one not already installed (Linux/x64)
   unarchive:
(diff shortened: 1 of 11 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk10/tasks/main.yml
```diff
@@ -21,7 +21,9 @@
   shell: ls -ld /usr/lib/jvm/jdk-10* >/dev/null 2>&1
   ignore_errors: yes
   register: adoptopenjdk10_installed
-  tags: adoptopenjdk10
+  tags:
+    - adoptopenjdk10
+    - skip_ansible_lint
```
karianna: Does that mean skip the linter?

sxa555: I'm assuming so ... I just copied it from the adoptopenjdk9 one ... As mentioned elswehere recently we need to try and implement some ansible best practices instead of doing this stuff but that should probably be part of a general overhaul

jdekonin:
Can you fix up what is being committed?  Looks like you are resubmitting the same change already delivered under #546.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk10/tasks/main.yml
karianna: yeah - we shouldn't skip the linting really :-)

PR #560, opened 2018-11-12
Install OpenSSL in windows playbook

cwesMills:
OpenSSL is required to compile java on windows, so the OpenSSL role will
ensure the 32bit and 64bit versions are both installed into C:\openjdk

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
@@ -1,0 +1,56 @@
+---
+###########
+# OpenSSL #
+###########
+- name: Check if OpenSSL 32bit installed
+  win_shell: Test-Path -Path C:\openjdk\OpenSSL-1.1.1-x86_32
+  register: openssl32_installed
+  tags: openssl
+
+- name: Check if OpenSSL 64bit installed
+  win_shell: Test-Path -Path C:\openjdk\OpenSSL-1.1.1-x86_64
+  register: openssl64_installed
+  tags: openssl
+
+- name: Download OpenSSL-1.1.1
+  win_get_url:
+    url: https://www.openssl.org/source/openssl-1.1.1.tar.gz
+    dest: C:\temp\openssl-1.1.1.tar.gz
+  when: (openssl32_installed == False) or (openssl64_installed == False)
+  tags: openssl
+
+- name: Unpack OpenSSL-1.1.1
+  win_unzip:
+    src: C:\temp\openssl-1.1.1.tar.gz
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+###########
+# OpenSSL #
+###########
+- name: Check if OpenSSL 32bit installed
+  win_stat:
+    path: 'C:\openjdk\OpenSSL-1.1.1-x86_32'
+  register: openssl32_installed
+  tags: openssl
+
+- name: Check if OpenSSL 64bit installed
```
jdekonin: You could combine this check into one statement (possibly).  On linux this would work,
` shell: ls -d /cygdrive/c/openjdk/OpenSSL* >/dev/null 2>&1`

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+  win_stat:
+    path: 'C:\openjdk\OpenSSL-1.1.1-x86_64'
+  register: openssl64_installed
+  tags: openssl
+
+- name: Check if OpenSSL downloaded
```
jdekonin: If its not installed, determined by the previous check, do everything.  I don't think this is needed.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+    nmake install
+  args:
+    executable: cmd
+  when: (openssl64_installed.stat.exists == false)
+  tags: openssl
+
+- name: Cleanup OpenSSL
+  win_file:
+    path: C:\temp\openssl-1.1.1
+    state: absent
```
jdekonin: Is there a download file from the unpack step?

cwesMills: That step uses the delete_archive option to delete the archive

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
cwesMills: I've found a way to check if a directory/file exists with Test-Path in powershell. It outputs either true or false which gets saved to the variable on the register line.

PR #566, opened 2018-11-13
Adding to Freetype windows role

### LINKED ISSUE #367: Add building of freetype (32+64 versions) to the Windows playbook
Current freetype role at https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freetype/tasks/main.yml seems to download the source but not build it. The built 64-bit version is in `C:\openjdk\freetype\{include,lib}` on the build machines - unclear of the source. We need a 32-bit version to satisfy https://github.com/AdoptOpenJDK/openjdk-build/issues/342, and get both versions into the ansible playbooks for setting up windows machines

cwesMills:
JDK8 compile will fail unless freetype is built beforehand, so these changes
will build freetype 32bit and 64bit versions inside C:\openjdk\freetype-2.5.3
Fixes #367 

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freetype/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freetype/tasks/main.yml
@@ -39,3 +39,31 @@
     state: present
   when: (freetype_installed.stat.exists == false)
   tags: Freetype
+
+- name: Check if freetype 32bit has been built
+  win_shell: Test-Path -Path C:\openjdk\freetype-2.5.3\lib32
+  register: freetype32_built
+  tags: Freetype
+
+- name: Check if freetype 64bit has been built
+  win_shell: Test-Path -Path C:\openjdk\freetype-2.5.3\lib64
+  register: freetype64_built
+  tags: Freetype
+
+- name: Build freetype-2.5.3 32bit
+  win_shell: |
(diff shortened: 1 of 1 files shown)


karianna:
How are the 64bit versions actually 32bit?  That would probably be an issue.  Is this change to support OpenJ9 builds only?  I'd probably get @ali-ince and @johnoliver to review this as well if the builds are relying on this...

jdekonin:
Its a requirement for openjdk as I understand it.  Microsoft Visual Studio 10.0 is 32bit only.  @cwesMills could you test compiling these like you did with the OpenSSL role using Microsoft Visual Studio 12 which has for 32 & 64 bit support?

cwesMills:
Visual Studio 12 gives the same result. When running `file lib32/freetype.dll`, the output is `lib32/freetype.dll: PE32 executable (DLL) (GUI) Intel 80386, for MS Windows`, and `file lib64/freetype.dll` outputs `lib64/freetype.dll: PE32+ executable (DLL) (GUI) x86-64, for MS Windows`. It seems we have the same problem with the [OpenSSL install.](https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/560)

cwesMills:
It turns out the `PE32+ executable` is actually 64bit, we were just confused by the naming convention.

jdekonin:
Okay, I'm good with this solution.

PR #568, opened 2018-11-15
Ansible: Updates to AIX scripts

CJKwork:
Fixes for issues:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/182
and
https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/562

### DIFF
--- a/ansible/playbooks/aix.yml
+++ b/ansible/playbooks/aix.yml
@@ -90,18 +90,18 @@
 
         - name: Transfer Openssl to remote host
           copy:
-            src: /Vendor_Files/aix/openssl-1.0.2.1300.tar.gz
-            dest: /tmp/openssl-1.0.2.1300.tar.gz
+            src: /Vendor_Files/aix/openssl/openssl-1.0.2.1600.tar.Z
+            dest: /tmp/openssl-1.0.2.1600.tar.Z
           when: does_openssl_exist.rc != 0
           tags: openssl
 
         - name: Extract Openssl
-          shell: gzip -cd /tmp/openssl-1.0.2.1300.tar.gz | tar xf - -C /tmp/
+          shell: uncompress /tmp/openssl-1.0.2.1600.tar.Z | tar xf - -C /tmp/
           when: does_openssl_exist.rc != 0
           tags: openssl
 
         - name: Install IBM Openssl - installp
-          command: installp -aXYgd /tmp/openssl-1.0.2.1300 openssl.base
+          command: installp -aXYgd /tmp/openssl-1.0.2.1600 openssl.base
           register: result.openssl
           ignore_errors: yes
           when: does_openssl_exist.rc != 0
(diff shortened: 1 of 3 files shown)


ansible/playbooks/aix.yml
```diff
@@ -14,9 +14,23 @@
...
+      # Set standard variables
+        - name: Load AdoptOpenJDKs variable file
+          include_vars: adoptopenjdk_variables.yml
+      # Install bash if it's not already there
+        - name: Check if bash is installed
+          shell: bash --version >/dev/null 2>&1
+          register: does_bash_exist
+          ignore_errors: yes
+        - name: install bash
+          yum: name=bash state=present update_cache=yes
```
jdekonin: Can bash not be added to the yum installer section?  Or is it needed for the filesystem configuration?

sxa555: @jdekonin - `AIX_filesystem_config.sh` has `#!/bin/bash` at the top, so this would be a prereq for that (although that's an existing piece of code). I'd be tempted to re-engineer it in `sh`/`ksh` if that really is the only reason for having this extra section to install `bash`

jdekonin: There should be filesystem configuration run every time this playbook is run, its a one time setup action, much like the openssl and initial yum install.  I wonder if these steps/actions would be better broken out and added as a basic setup document before the playbook is thrown at it.

Either way, I guess this is a discussion for a improvement beyond this commit and can be taken to a new issue, same with my suggestion to cpan below.

ansible/playbooks/aix.yml
```diff
@@ -596,6 +638,23 @@
             - /tmp/freemarker-2.3.8.tar.gz
           tags: freemarker
 
+      ########
+      # cpan #
+      ########
+        - name: Ensure memory limits for root are unlimited
```
jdekonin: We should add a check to see if the module is installed to avoid doing a work thats already been performed.  `perldoc -l Text::CSV`

ansible/playbooks/aix.yml
sxa555: Yep let's separate that out into a separate issue.

ansible/playbooks/aix.yml
sxa555: Fair comment - are you able to address this @CJKwork ?

sxa555: Now done

ansible/playbooks/aix.yml
sxa555: Done: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/655

jdekonin:
Can we get this rebased, reviewed or polished to get it committed please?  Its blocking other PR work for this playbook.

sxa555:
Will need to update this to openssl 1601 which sorts out some issues with yum but integrating this for now as it's blocking integration of other PRs

sxa555:
No I'm not as it's falling foul of the Travis lint issue that was causing everything to fail a while back so I'll change to 1601 to force the rescan

PR #581, opened 2018-11-19
Dont replace usrbingmake

sxa555:
Same reasoning as https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/576 - it causes the playbooks to fail :-\


sxa555:
@jdekonin done

PR #583, opened 2018-11-20
ansible: Install gmake 4.1 on zLinux

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/357 (or at least fixes the current problem!)


sxa555:
Done. Resolving it will revert the change made in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/576 ... On further investigation it looks like the "real" file `/usr/bin/gmake` was actually a make 4.1 that had been copied there instead of symlinked, despite the system `make` package claiming to own it (presumably because it normally creates a gmake -> make symlink)

I'll convert `/usr/bin/gmake` back to a symlink on the machines I know currently have a real file so re-introducing the symlink creation doesn't cause a problem.

PR #584, opened 2018-11-20
Use common ansible role for download/extract/configure/make install

sxa555:
Many source packages that we're installing into /usr/local on the machine have a similar pattern for the installation. This PR creates a single role which can be called with the parameters to do those source compiles.

NOTE 1: I had a quick look on Galaxy and couldn't find anything similar, but I didn't look check everything, and regardless putting this role together was educational even if we don't use it :-) I suggest we do a proper search before choosing to merge this.
NOTE 2: It doesn't currently check the version of product installed in the target directory which some of our other source compile roles do (`autoconf`, `cmake`, `curl`, `nasm`, `GIT_Source`)
NOTE 3: This could, of course, do with using a properly generated temp dir/file for the stuff in /tmp ...
NOTE 4: I'd also like to use /dev/shm if available on Linux which should speed things up a little.

Any other feedback welcome!


karianna:
conflicts!

sxa555:
Not sure which conflicts they were @karianna but I'm not seeing them now - Perhaps they disappeared when the zlinux gmake 4.1 one was merged as I'd based this off my branch that had that change already in it

sxa555:
I'm getting a bit more ambitious with this and have managed to replace some more roles with it now that I can do version checking. Also at https://github.com/sxa555/local_srcinstall and I'm copying in the current README.md from there to explain what I've got so far ... If anyone can test this on a "safe" system that would be appreciated ;-)

# local_srcinstall
ansible role for downloading and installing from source tarballs on UNIX platforms

Sample usage from within a playbook (short version)
```
    - role: local_srcinstall
      src_tarball: https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz 
      installed_target: /usr/local/bin/ccache
```

Sample usage from within a playbook (long version)
```
    - role: local_srcinstall
      src_tarball: https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz 
      installed_target: /usr/local/ccache/bin/ccache
      srcinstall_configure_params: --prefix=/usr/local/ccache
      srcinstall_version_required: 3.4.2
```
The first version is fairly straightforward. It will check for the presence
of /usr/local/bin/ccache and if it's not there then the source will be
downloaded and then `./configure --prefix=/usr/local && make && make install`
will be performed.

The long version causes various different things to happen.
- Firstly we are choosing to install into /usr/local/ccache as a prefix instead of /usr/local, so `srcinstall_configure_params` is added to the configure line
- Secondly a minimum version is specifed. In this case, a check will be run by parsing the last "word" of the first line obtained from running `--version` against the `installed_target`
- If `installed_target` does not exist, then a --version check will be done against just running the basename of the `installed_target` i.e. `ccache --version`
- The version will be checked against `srcinstall_version_required` and it will only be downloaded and built if it is backlevel

Gotchas:

- If you don't specify `srcinstall_version_required` then there is no check to see if the product is already on your machine, it ONLY checks the absolute path from `installed_target`
  This is fine for things that aren't likely to be installed via your package manager elsewhere, but may cause confusion otherwise

sxa555:
Current conflict is because of the cmake entry being moved ... If this is landed then the cmake section should be moved to match and replace the cmake line

karianna:
@sxa555 rebase required.

sxa555:
Removing `DO NOT MERGE` has this has been kicking around for a while and no-one seems to have any any fundamental objections. I think it's ready to go now ...

PR #587, opened 2018-11-20
Switch Vagrant file from using ubuntu/trusty64 to generic/ubuntu1604

sxa555:
This will move the image up from 14.04 (which isn't our main platform) to 16.04 and also use an image that will work with providers other than VirtualBox (generic/ubuntu1604 supports):
 - hyperv
 - libvirt
 - parallels
 - virtualbox
 - vmware_desktop

Explicit entry for /vagrant is because it wasn't working for me by default

Reviewers: PLEASE test this on your setup and let me know if it works (especially the `/vagrant` mount) and whether it causes regressions compared to the previous image - I'm going to leave this as WIP for at least a few days until I'm confident it doesn't break anything.

### DIFF
--- a/ansible/Vagrantfile
+++ b/ansible/Vagrantfile
@@ -30,7 +30,8 @@ SCRIPT
 Vagrant.configure("2") do |config|
 
   config.vm.define :adoptopenjdk do |adoptopenjdk|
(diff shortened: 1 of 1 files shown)


sxa555:
Been working with this again today on my laptop (RHEL+qemu) and it seems ok

PR #589, opened 2018-11-20
Added nasm role to Windows playbook

cwesMills:
Windows playbook will now install nasm-2.13.03 into C:\openjdk

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/nasm/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/nasm/tasks/main.yml
@@ -1,0 +1,24 @@
+---
+########
+# nasm #
+########
+- name: Check if nasm is installed
+  win_stat:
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/nasm/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+########
+# nasm #
+########
+- name: Check if nasm is installed
+  win_shell: Test-Path -Path C:\openjdk\nasm-2.13.03
```
sxa555: Can you modify this to use `win_stat` to check for the presence instead of calling the shell? Something like this ought to work (please test it first though!)
```
  win_stat: path= 'C:\openjdk\nasm-2.13.03'  
  register: nasm_stat_file

- name: Download nasmCreate DEF.txt file if ABC.txt exists 
  [...]
  when: nasm_stat_file.stat.exists == False
```

PR #590, opened 2018-11-21
ansible: Do not install EPEL on s390x

sxa555:
Was added in https://github.com/AdoptOpenJDK/openjdk-infrastructure/commit/86bccd2e3625e64eecd2ae22619906cb08a8f820 but is causing subsequent yum operations in the playbooks to fail. If it was only needed to support an extra package on ppc64le then we might get away with just not enabling this. If it's needed for other things then this is a once in a lifetime opportunity change to document it ;-)

Any comments / suggestions / experience welcome ;-)


lasombra:
There's no EPEL repository for s390x, https://fedoraproject.org/wiki/EPEL#What_packages_and_versions_are_available_in_EPEL.3F, so removing it makes sense.

PR #596, opened 2018-11-22
ansible: Add libpng-devel to CentOS/RHEL (required for freetype build)

sxa555:
Required for https://github.com/AdoptOpenJDK/openjdk-build/pull/758

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
@@ -28,6 +28,7 @@ Build_Tool_Packages:
   - java-1.8.0-openjdk-devel
   - libcurl-devel
   - libdwarf-devel
+  - libpng-devel
   - libXext-devel
   - libXrender-devel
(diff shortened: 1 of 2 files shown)


karianna:
> @karianna
> 
> Short answer: Yes
> 
> Long answer: I didn't do it this time because
> 
> 1. I didn't want to just have one line with a comment beside it when all the others don't as yet and I'd prefer to do a one-off overhaul
> 2. We've got someone asking for separate documentation on what is required to build and perhaps that is the place for it
> 3. At the moment unless someone refactors things it's likely that "blame" on this source file will show the commit comment that makes it clear why it was added
> 4. There was a question asked recently of whether we should auto-download and install freetype or do what the OpenJ9 machines do (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/freemarker/tasks/main.yml) and use a pre-installed version to remove an external dependency on the build process (~jenkins doesn't seem like the right place for that, but that's a separate issue)
> 
> But if you still want o me to I can add a comment for this one ;-)

OK, let's just clean up the whole lot in a big comment block I reckon.

PR #598, opened 2018-11-22
Install binutils 2.31 from source on RHEL

vsebe:
Add role binutils to install a newer version of the GNU Binutils tools.
Required by gcc-7.3

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
@@ -1,0 +1,79 @@
+---
+############
+# binutils #
+############
+
+# gcc-7.3 requires binutils v2.24 or newer
+# Note the installation on CentOS x86_64 is done via Developer Toolset 7 from
+# Software Collections (see the Common role for CentOS).
+
+# Note:
+# In order to use the binutils 2.31 tools (ld, as, ar) add:
+# /usr/local/binutils/bin to PATH
+# /usr/local/binutils/lib to LD_LIBRARY_PATH
+
+- name: Get the binutils version on RHEL/CentOS
+  shell: (/usr/local/binutils/bin/ld -v 2>/dev/null || ld -v 2>/dev/null ) | sed -n 's/^.*\( [0-9]\.[0-9][0-9]\).*/\1/p'| awk '{print $1}'
+  ignore_errors: yes
+  register: binutils_version
+  tags: binutils
+  when: ansible_distribution == "RedHat"
+
+- debug: var=binutils_version.stdout
+  tags: binutils
+
+# should install the same version used to build gcc-7.3
+- name: Download binutils-2.31
+  get_url:
+    url: https://ftp.gnu.org/gnu/binutils/binutils-2.31.1.tar.xz
+    dest: /tmp/binutils-2.31.1.tar.xz
+    force: no
+    mode: 0644
+  when:
+    - ansible_distribution == "RedHat"
+    - (binutils_version.stdout == '') or (binutils_version.stdout | version_compare('2.24', operator='lt'))
+  tags: binutils
+
+- name: Extract binutils-2.31
+  unarchive:
+    src: /tmp/binutils-2.31.1.tar.xz
+    dest: /tmp/
+    copy: False
+  when:
+    - ansible_distribution == "RedHat"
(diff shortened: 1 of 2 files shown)


vsebe:
@sxa555 fyi

sxa555:
We've already been building with gcc-7.3 on our RHEL/CentOS boxes ... Can you clarify what problem is currently visible that this solves?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+    - ansible_architecture != "CentOS" and ansible_architecture != "x86-64"
+    - (binutils_version.stdout == '') or (binutils_version.stdout | version_compare('2.24', operator='lt'))
+  tags: binutils
+
+# Note: Do not use --suffix option!
+# No suffix tools - ld, as, ar - are required by OpenJ9.
+- name: Build and install binutils-2.31 in /usr/local
+  shell: cd /tmp/binutils-2.31.1 && ./configure --prefix=/usr/local/binutils-2.31.1 && make -j {{ ansible_processor_vcpus }} && make install
+  when:
+    - ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
```
sxa555: Unless I'm mistaken these two lines end up as just `ansible architecture == "RedHat and ansible acrchitecture != "x86-64"` which is simpler

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+# Note the installation on CentOS x86_64 is done via Developer Toolset 7 from
+# Software Collections (see the Common role for CentOS).
+
+# Note:
+# In order to use the binutils 2.31 tools (ld, as, ar) add:
+# /usr/local/binutils/bin to PATH
+# /usr/local/binutils/lib to LD_LIBRARY_PATH
+
+- name: Get the binutils version on RHEL/CentOS
+  shell: (/usr/local/binutils/bin/ld -v 2>/dev/null || ld -v 2>/dev/null ) | sed -n 's/^.*\( [0-9]\.[0-9][0-9]\).*/\1/p'| awk '{print $1}'
```
sxa555: This appears to ignore the possibility that there will be a suitable version installed in /usr/bin (RHEL7)

vsebe: `... || ld -v` should take care of the one installed in /usr/bin.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
vsebe: RHEL x86 is a valid platform. CentOS x86 is the exception as it would use devtoolset-7.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
sxa555: Fair point :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/binutils/tasks/main.yml
sxa555: Not sure I understand "RHEL x86 is a valid playform" relative to my coment - your logic is saying **don't** do this on RHEL/x64. And doesn't special case CenrOS only on x64 as far as I can tell

sxa555: Hmmm so I think you're saying you want it done on RHEL or centos on everything other than x86, in which case the second line should have an `or` in the middle instead of an `and` I believe i.e.:
```
     - ansible_architecture != "CentOS" or ansible_architecture != "x86-64" 
```
otherwise when you currently have resolves to what I said in my earlier comment which excludes CentOS entirely.

vsebe: @sxa555 - removed centos entirely

PR #599, opened 2018-11-22
ccache should be installed after all compilers are installed

jdekonin:


jdekonin:
Looking to see if there is some way to refresh the configuration.  At a minimum I was asked to install ccache after gcc7 because it wasn't working in this format.

sxa555:
@jdekonin @karianna On what platforms are you seeing this being a problem - I believe `gcc` is installed an all Linux platforms from the default repositories in the Common section, therefore a suitable compiler should always be available even before our own compiled ones

Issue #347, opened 2018-04-20
Add Docker to all linux / mac playbooks 


smlambert:
We can/should increase the number of test machines that have Docker available on them, to optimize our machine usage (to have the flexibility to add more or shift some testing into Docker containers.

Mesbah-Alam:
The more machines we'd have with Docker installed, the better will shorter it will take for the External test builds (https://ci.adoptopenjdk.net/view/External%20tests/) to run, as we are going to parallellize the execution process for each different type of test under the External test builds. Currently they are running serially and taking somewhere betteen 2-4 hours on an average to finish. Moreover, we are actively working on adding new external tests (e.g. OpenLiberty will be added soon) - which will slow down the builds even further and may force us to make these builds run only weekly as opposed to nightly. So, having more docker-enabled machines will greatly help the cause going forward.

sxa555:
Running the current playbooks on the following machines so will see what we end up with - current list with the tag are as follows: https://ci.adoptopenjdk.net/label/sw.tool.docker/ ;-)
- [test-osuosl-ubuntu1604-ppc64le-2](https://ci.adoptopenjdk.net/computer/test-osuosl-ppc64le-ubuntu-16-04-2/)
- [test-packet-ubuntu1604-x64-2](https://ci.adoptopenjdk.net/computer/test-packet-ubuntu1604-x64-2/)

sxa555:
Now installed on both machines - if you can test and let me know if any further setup is needed that isn't in the playbooks that would be good

sxa555:
@chandrams looks like there were alrady one ppc64le and s390x machine available - are they working ok for you now?

sxa555:
Will assume this is ok now - please reopen if not

PR #600, opened 2018-11-23
Added s390x and ppc64le arch to docker role

cwesMills:
Docker will now be installed on ubuntu (x86_64/s390x/ppc64le), RHEL7 (x86_64/s390x/
ppc64le), cent7 (x86_64/s390x/ppc64le), and SLES12 (x86_64). Formerly docker was
only installed on these operating systems on x86_64.

Related to #347 


cwesMills:
I've tested these changes against a s390x ubuntu machine, and it successfully installed docker.

PR #606, opened 2018-11-28
ansible: Switch adoptopenjdk11 role to install OpenJ9 on s390x

sxa555:
Puts correct change into s390x section of the adoptopenjdk11 setup (Now matches the comment)
Fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/594


jdekonin:
Good catch.  I read the comment and didn't confirm the link location.  I had copied from adoptopenjdk10 directory, so this is wrong there as well.  Might as well get that change in this commit?

sxa555:
@jdekonin Ah yes thanks - changed both now :-)

PR #608, opened 2018-11-29
Add SLES12/s390x box into inventory.yml

sxa555:
Not sure why this was missing but since I'm handing this to the test team it makes sense to be able to put it back in so ansible can run against it. May rename it to have `test-` instead of `build-` as a prefix later


sxa555:
@karianna What needs to be the other way round?

sxa555:
Oh the `sles` vs `s390x` ... Yeah could do - it was the wrong way round in jenkins too but the hostname on the box itself looks "correct". Will update both ...

sxa555:
Changed

Issue #594, opened 2018-11-21
Add Bootjdk11 (for jdk12) to Ansible playbooks and run/install on all build machines


AdamBrousseau:
Do we have a flag so we can choose whether to install hs vs openj9?
Do we have flag(s) so we only install bootjdks on build machines and not test machines?

Platforms:
- [x] Linux x64 #604 
- [x] Linux ppc64le #604 
- [x] Linux s390x #604 #606
- [x] Windows #605 
- [ ] Mac #614
- [ ] Arm/Aarch?
- [x] AIX #632

AdamBrousseau:
Is this finished for all platforms?

jdekonin:
I didn't catch the comment in the description of #606.  This isn't complete no.

sxa555:
Completed AFAIK (since we've shipped two releases of JDK12 now!) If anyone knows otherwise feel free to re-open

PR #614, opened 2018-12-03
[WIP] Rough MacOS Playbook

cwillhelm:
Currently running locally against an internal host with:

`ansible-playbook --user [redacted]  MacOS.yml --ask-become-pass`

In reference to: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/212

Signed-off-by: Connor Willhelm <connor.willhelm@ibm.com>

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOS.yml
@@ -1,0 +1,563 @@
+---
+#########
+# MacOS #
+#########
+
+- name: MacOS Playbook
+  hosts: all
+  environment:
+    TOOLS_941: /tmp/Command_Line_Tools_macOS_10.13_for_Xcode_9.4.1.dmg
+    TMPMOUNT_941: /tmp/xcode941
+    DMG_FILE_7: /tmp/Xcode_7.2.1.dmg
+    DMG_FILE_4: /tmp/Xcode_4.6.3.dmg
+    TMPMOUNT_7: /tmp/xcode721
+    TMPMOUNT_4: /tmp/xcode463
+  tasks:
+  
+##############
+# Brew Tasks #
+##############
+
+  - name: Check if Homebrew is already installed
+    stat:
+      path: /usr/local/bin/brew
+    register: brew
+
+  - name: Install and Setup Brew
+    shell: |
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOS.yml
```diff
@@ -0,0 +1,563 @@
...
+    #  - ant-contrib
+      state: present
+    tags: brew_packages
+  
+  ##############
+  # JDK11 Only #
+  ##############
+  - name: Download and extract JDK11 
+    unarchive:
+      src: https://github.com/AdoptOpenJDK/openjdk10-releases/releases/download/jdk-10.0.2%2B13/OpenJDK10_x64_Mac_jdk-10.0.2.13.tar.gz
```
jdekonin: Try how the unix playbook role does this.
https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/8be6072a398346ab2d97a3dc352fe3c4ee44ce62/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk11/tasks/main.yml#L50

AdamBrousseau:
@cwillhelm can you add bootjdk11 (ref #594) since this is still WIP?

cwillhelm:
@AdamBrousseau will do, I've been without a machine for testing until we get the Macs back from UNB, so the playbook has been on the backburner a bit. Once we get them back I'll start making more progress.

gdams:
Can we not use roles here rather than pushing all of the mac stuff into a single area?

sxa555:
I'm ok with an initial version being done as one file then splitting out into roles later to match the Linux platforms. More critical that we get "something" in place just now I think.

sxa555:
I'm ok with an initial version being done as one file then splitting out into roles later to match the Linux platforms. More critical that we get "something" in place just now I think.

sxa555:
@cwillhelm What is the purpose of the cache here? We don't have it on other platforms, and I don't see the reasoning documented anywhere.

AdamBrousseau:
Repo cache is something we use in OpenJ9 builds to make clones faster. Adopt builds do not use this at the moment. Since we have a job that will automatically add & update the cache, I don't think it's necessary for the playbooks at this point.

jdekonin:
Taken from discussion with @cwillhelm, `pmset -a sleep 0` might be of use in stopping the mac from going to sleep and becoming disabled in jenkins.

gdams:
closing in favor of https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/826

PR #616, opened 2018-12-04
Updated docker install in unix playbook

### LINKED ISSUE #615: Docker role failing on CentOS/ppc64le
Presumably caused by the addition if docker in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/600:

```
TASK [Docker : Install Docker for ALL] *****************************************
11:52:21
fatal: [build-osuosl-centos74-ppc64le-2]: FAILED! => {"changed": false, "failed": true, "msg": "https://download.docker.com/linux/centos/7/ppc64le/stable/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found\nTrying other mirror.\nTo address this issue please refer to the below wiki article \n\nhttps://wiki.centos.org/yum-errors\n\nIf above article doesn't help to resolve this issue please use https://bugs.centos.org/.\n\n\n\n One of the configured repositories failed (Docker CE Stable - ppc64le),\n and yum doesn't have enough cached data to continue. At this point the only\n safe thing yum can do is fail. There are a few ways to work \"fix\" this:\n\n     1. Contact the upstream for the repository and get them to fix the problem.\n\n     2. Reconfigure the baseurl/etc. for the repository, to point to a working\n        upstream. This is most often useful if you are using a newer\n        distribution release than is supported by the repository (and the\n        packages for the previous distrib…
```

cwesMills:
Docker will not be installed on cent/rhel 7 for s390x and
ppc64le, because on these operating systems, docker is only
supported for x86_64.
Fixes #615

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
@@ -93,7 +93,7 @@
     - docker_installed.rc != 0
     - ansible_distribution == "RedHat"
     - ansible_distribution_major_version == "7"
-    - (ansible_architecture == "x86_64" or ansible_architecture == "s390x" or ansible_architecture == "ppc64le")
+    - ansible_architecture == "x86_64"
   tags: docker
 
 - name: Add Docker Repo for RedHat 7
@@ -108,7 +108,7 @@
     - docker_installed.rc != 0
     - ansible_distribution == "RedHat"
     - ansible_distribution_major_version == "7"
-    - (ansible_architecture == "x86_64" or ansible_architecture == "s390x" or ansible_architecture == "ppc64le")
+    - ansible_architecture == "x86_64"
   tags: docker
 
 - name: Import Docker Repo key for RedHat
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
```diff
@@ -187,55 +187,66 @@
...
-    - (ansible_distribution == "RedHat" and ansible_distribution_major_version == "7" ) or (ansible_distribution == "Ubuntu") or (ansible_distribution == "CentOS" and ansible_distribution_major_version == "7" )
+    - ansible_distribution == "Ubuntu"
     - (ansible_architecture == "x86_64" or ansible_architecture == "s390x" or ansible_architecture == "ppc64le")
   tags:
     - docker
     # TODO: Package installs should not use latest
     - skip_ansible_lint
 
-- name: Add Jenkins user to the docker group for RHEL7, cent7, and ubuntu
+- name: Install Docker for Cent/Rhel 7
```
karianna: Is It really CentOS 7 or are we conflating that it's Ansible 7?

cwesMills: I'm not sure, I just used the same syntax as what was in the file before I made my changes

sxa555: This is ok @karianna - `ansible_distribution_version` is checking the version of the OS distribution on the target systemm, not the ansible version (Compare with https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/573/files which is the ansible version check)

PR #617, opened 2018-12-04
Updated Windows_Updates role

cwesMills:
Ansible thinks the when statement is a part of the
win_reboot module, and the register statement is a
part of the win_updates module causing errors. Removed
spaces so when and register statements will work. Removed
shutdown_timeout as it is deprecated

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Windows_Updates/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Windows_Updates/tasks/main.yml
@@ -4,13 +4,12 @@
 ###################
 - name: Download and Install Windows Updates
   win_updates:
-    register: update_result
+  register: update_result
   ignore_errors: yes
   tags: Windows_Updates
 
 - name: Reboot machine if necessary
   win_reboot:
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Windows_Updates/tasks/main.yml
```diff
@@ -4,13 +4,12 @@
...
 - name: Download and Install Windows Updates
   win_updates:
-    register: update_result
+  register: update_result
   ignore_errors: yes
   tags: Windows_Updates
 
 - name: Reboot machine if necessary
   win_reboot:
-    shutdown_timeout: 3600
```
karianna: If this was deprecated, what's replaced it?

cwesMills: Nothing has directly replaced it. The shutdown_timeout option is the max amount of time to wait for shutdown to begin, and the reboot_timeout is the max amount of time to wait for machine to re-appear on the network and respond to a test command. It should work using only the reboot_timeout option.

PR #620, opened 2018-12-05
Remove any backlevel distribution-supplied git

sxa555:
Current RHEL and SLES are all shipped with a backlevel git compared to the one we need (2.15.0 which supports the `-C` parameter which the OpenJ9 builds use) so let's remove it when we install our own, and avoid installing it from the default repos in the first place. Added "dont_remove_system" tag so that people can skip it if they want to avoid adjusting up their existing one on their machines.

NOTE: This won't automatically remove it from machines that already have the distro-supplied package and have an updated version in /usr/local/bin, but if they're causing a problem they can be manually fixed up ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ addons:
 install:
   # Install ansible
   - pip install ansible
-  - pip install yamllint
-  - pip install ansible-lint
+  - pip install yamllint==1.13.0
+  - pip install ansible-lint==3.5.1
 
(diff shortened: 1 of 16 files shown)


sxa555:
Closing as this inadvertently sneaked itself in under https://github.com/AdoptOpenJDK/openjdk-infrastructure/commit/ac71ca8beb8f5e8bb1f6904bf36b44a6455b4e55

PR #623, opened 2018-12-05
WIP: Add XML::parser to AIX playbook

jdekonin:
* related AdoptOpenJDK/openjdk-infrastructure#622


jdekonin:
I think #568 should go in first.  Maybe I'll work on adding a cpan module check if @CJKwork hasn't already done that under #568.

sxa555:
Suspect there's a load more test prereqs missing from the AIX playbook too ;-)

sxa555:
We should probably add everything that's in the UNIX CPAN role ... Also not sure why we seem to be trying to install CPAN modules on AIX into the jenkins user's home directory instead of globally as we do elsewhere. Probably something we could change if we don't know why it was done, and if it doesn't work we can document it in the playbook ;-)

jdekonin:
I also moved the environment settings to the root of the playbook to avoid duplication.  I think this means that the lint stage will pass now??  I don't have the ability to test this via AWX currently.

karianna:
> I also moved the environment settings to the root of the playbook to avoid duplication. I think this means that the lint stage will pass now?? I don't have the ability to test this via AWX currently.

Travis seems to think it's OK so it's passed the linter.

karianna:
@sxa555 I'll let you review as my AIX is..... non existent :-)

jdekonin:
This PR is ready to merge.

ansible/playbooks/aix.yml
```diff
@@ -8,7 +8,9 @@
   remote_user: root
   become: yes
   environment:
-    PATH: "/opt/freeware/bin:{{ ansible_env.PATH }}"
+    PATH: "/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:{{ ansible_env.PATH }}"
```
karianna: Is this IBM path common for generic AIX hosts?

jdekonin:
Yes, that is the location the IBM xlc compiler installs too by default via installp.

Issue #618, opened 2018-12-05
Update playbooks with locales to support MBCS testing


smlambert:
Related to new test contributions: https://github.com/AdoptOpenJDK/openjdk-tests/issues/726

Note: A PR for this issue will follow shortly

toshiona:
Current Status
Linux: PR was merged.
AIX: Still uncertain how we can install AIX base OS's filesets.

sxa555:
@sej-jackson Are you able to schedule in some time to look at addressing this requirement in the next two weeks?

toshiona:
Sorry if I missed something, but in my understanding, AIX configuration changes had proceeded in #704. No request for playbook.

sxa555:
Unless this is done in the ansible playbook, this requirement will not be captured and any new machines will not have it configured in the same way. We shouldn't be making any configuration changes on machines without replicating the installations in the playbooks

Haroon-Khel:
@sxa @sej-jackson Whats left to do on this issue?

aixtools:
@sxa @sej-jackson @Haroon-Khel 

* I would like to rename the X11 role to BOS (or bos) - as these are all fileesets, like X11, that need to be verified as installed.
* as each OS level may have different versions - I also propose to drop X11 vendor files.
* Comments - especially on the role name.

sxa:
This is MBCS, not X11. @Haroon-Khel @aixtools Have we got the localtes required to run our MBCS tests on the AIX systems now? We should verify that these run correctly on our AIX test systems now.

PR #624, opened 2018-12-06
Update Linux playbooks with locales to support MBCS testing

toshiona:
Create Asian locales on Linux distributions which didn't create them.
Install locale filesets to AIX. "/Vendor_Files/aix" should contain the locale filesets of AIX OS.

#618

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
@@ -167,3 +167,32 @@
   when:
     - ansible_architecture == "x86_64"
   tags: expat
+
+###########
+# Locales #
+###########
+- name: Get locale list
+  shell: locale -a
+  register: localeList
+  changed_when: False
+  tags: locales
+
+- name: Create Japanese locale
+  shell: localedef -i ja_JP -c -f UTF-8 ja_JP.UTF-8
+  when: localeList.stdout|lower is not search("ja_jp\.utf8")
+  tags: locales
+
+- name: Create Korean locale
+  shell: localedef -i ko_KR -c -f UTF-8 ko_KR.UTF-8
+  when: localeList.stdout|lower is not search("ko_kr\.utf8")
+  tags: locales
+
+- name: Create Chinese locale
+  shell: localedef -i zh_CN -c -f UTF-8 zh_CN.UTF-8
+  when: localeList.stdout|lower is not search("zh_cn\.utf8")
+  tags: locales
+
+- name: Create Taiwanese locale
+  shell: localedef -i zh_TW -c -f UTF-8 zh_TW.UTF-8
+  when: localeList.stdout|lower is not search("zh_tw\.utf8")
+  tags: locales
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
@@ -47,5 +47,6 @@ Test_Tool_Packages:
   - xorg-x11
   - xorg-x11-server
   - xorg-x11-server-extra
+  - glibc-locale                  # Internationalization tests
 
 crontab_Patching: "/usr/bin/zypper refresh && /usr/bin/zypper -n up"
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
@@ -135,3 +135,32 @@
     - ansible_distribution_major_version == "7"
     - ansible_architecture == "s390x"
   tags: default_java
+
+###########
+# Locales #
+###########
+- name: Get locale list
+  shell: locale -a
+  register: localeList
+  changed_when: False
+  tags: locales
+
+- name: Create Japanese locale
+  shell: localedef -i ja_JP -c -f UTF-8 ja_JP.UTF-8
+  when: localeList.stdout|lower is not search("ja_jp\.utf8")
+  tags: locales
+
+- name: Create Korean locale
+  shell: localedef -i ko_KR -c -f UTF-8 ko_KR.UTF-8
+  when: localeList.stdout|lower is not search("ko_kr\.utf8")
+  tags: locales
+
+- name: Create Chinese locale
+  shell: localedef -i zh_CN -c -f UTF-8 zh_CN.UTF-8
+  when: localeList.stdout|lower is not search("zh_cn\.utf8")
+  tags: locales
+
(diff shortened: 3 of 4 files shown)


toshiona:
"/Vendor_Files/aix/asian.tar.gz" should contains the following filesets of AIX OS.

bos.iconv
bos.loc.com.CN
bos.loc.com.utf
bos.loc.iso.ja
bos.loc.iso.ko
bos.loc.iso.zh
bos.loc.pc
bos.loc.utf.JA_JP
bos.loc.utf.KO_KR
bos.loc.utf.ZH_CN
bos.loc.utf.ZH_TW

ansible/playbooks/aix.yml
```diff
@@ -717,6 +717,30 @@
             # Environment variables don't work as part of command
             - skip_ansible_lint
 
+      #########################
+      # Install Asian Locales #
+      #########################
+        - name: Transfer locales installp to remote host
+          copy:
+            src: /Vendor_Files/aix/asian.tar.gz
```
jdekonin: Can you document where/how to obtain these files for future reference when updating is required.

ansible/playbooks/aix.yml
```diff
@@ -717,6 +717,30 @@
...
+            dest: /tmp/locale/
+          tags: locale
+
+        - name: Extract asian.tar
+          shell: gzip -dc /tmp/locale/asian.tar.gz | tar xf - -C /tmp/locale
+          tags: locale
+
+        - name: Install locales - installp
+          command: installp -aXYgd /tmp/locale all
+          register: result.locale
```
jdekonin: What will installp do if this is already installed on the system?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/OracleLinux.yml
```diff
@@ -59,3 +59,23 @@
   shell: cd /tmp/expat-2.2.5 && ./configure && make -j {{ ansible_processor_vcpus }} && sudo make install
   become: yes
   tags: expat
+
+############################
+# Test tools               #
+############################
+- name: Create Japanese locale
+  shell: localedef -i ja_JP -c -f UTF-8 ja_JP.UTF-8
+  tags: test_tools
```
jdekonin: What will localedef do if the locale is already defined?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
```diff
@@ -167,3 +167,24 @@
...
     - ansible_architecture == "x86_64"
   tags: expat
+
+
+############################
+# Test tools               #
+############################
+- name: Create Japanese locale
+  shell: localedef -i ja_JP -c -f UTF-8 ja_JP.UTF-8
+  tags: test_tools
```
jdekonin: Would this likely be the same on RHEL?

ansible/playbooks/aix.yml
toshiona: Thank you for your review. Yes, I added fileset list which should contain the tar file.

ansible/playbooks/aix.yml
toshiona: The result is success with "already installed" message. I added to check available locales, anyway.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/OracleLinux.yml
toshiona: It will overwrite with the same data. But, I added to check available locale to avoid the situation.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
toshiona: RHEL set up these locales with default installation, but I added the same code with condition check.

ansible/playbooks/aix.yml
toshiona: I expect these filesets are picked from AIX base OS media. I can create the file if the target AIX's version (TL/SP) is identified.
I appreciate any advice.

jdekonin: For what version is this file generated?

toshiona: Thank you for the comment. 
My test environment is AIX7.1TL5.

I think it's better the target filesets are stored in /Vendor_Files/aix without packing. I updated this pull request.

I don't know how I can request to store files in /Vendor_Files. Could you kindly advice that?

I don't think the AIX filesets can be downloaded (We can download update images though). I believe they should be copied from AIX base OS media.

AdamBrousseau:
I don't think either of us feels confident enough with the content to provide a review here, sorry.

jdekonin:
Please squash the commits.  For AIX the files are in an unreachable location for anyone not running off the AdoptOpenJDK AWX server.  I understand that these files are something that a user would need a license to obtain, so I think there needs to be an option specified to enable installing these files.  Without this option the playbook will fail in all other environments.  Detailing where the files can be obtained could also be added.

ansible/playbooks/aix.yml
sxa555: @toshiona We'd need to be 100% certain we're not violating any license agreements if we do that. I'm nervous about copying stuff from the AIX install media across to our servers (this redistributing them)
Can you separate the AIX bits into a separate PR so we can get the Linux stuff merged ASAP.

toshiona:
@sxa555 @jdekonin 
Thank you for your suggestion. OK, I've removed AIX part from this PR.

toshiona:
@sxa555 @jdekonin 
Sorry, I overlooked a conflict due to other update. Could you re-review this? Please let me know if I did something wrong.

jdekonin:
Please squash the commits, looks fine to me otherwise.

toshiona:
@jdekonin Yes, it's done. I hope my operation is correct. Could you review again, please?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
```diff
@@ -47,5 +47,6 @@ Test_Tool_Packages:
   - xorg-x11
   - xorg-x11-server
   - xorg-x11-server-extra
+  - glibc-locale
```
karianna: Can you add a comment as to why this was added

toshiona: Thank you for your support, @karianna . I've updated the line.

toshiona:
It seems this change has not been applied to test systems.
Should I post a request to run the updated playbook?

karianna:
> It seems this change has not been applied to test systems.
> Should I post a request to run the updated playbook?

Please drop a note to the #infrastructure channel on slack, I thought they should have auto run by now.

Issue #553, opened 2018-10-31
Get Windows playbooks production ready


sxa555:
General issue to cover work being done to make the Windows playbooks "production ready". Should also result in some setup doc.

Info on setting up the communication to windows boxes (since it uses WinRM rather than ssh generally, and the `pywinrm` package to communicate with it) are at https://docs.ansible.com/ansible/latest/user_guide/windows.html

See also https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/300 which has been pending for a while.

sxa555:
Related: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/378

karianna:
Build machines should always have rsync installed via Cygwin.  I'm not sure if/how Ansible can cover that.

sxa555:
Yep we can cover by ansible - it's just another dependency we need to call out explicitly :-)

sxa555:
@cwesMills Has your work/testing been based on PR300 as a starting point?

cwesMills:
I did not use PR300 as a reference, but the changes I had to make were:
- Install nasm by downloading `https://www.nasm.us/pub/nasm/releasebuilds/2.13.03/win64/nasm-2.13.03-win64.zip` and extracting it to `C:\openjdk\nasm-2.13.03`
- Install clang based on steps provided in [this PR](https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/514)
- Run `ssh-host-config` based on [these steps](https://go.cloudbees.com/docs/cloudbees-documentation/admin-cje/agents/#_microsoft_windows). With the exception, when prompted `Enter the value of CYGWIN for the daemon:` the value should be nothing.
- The jdk I used didn't have the proper cacerts ([known issue](https://github.com/AdoptOpenJDK/openjdk-build/issues/555)), so I had to copy it from an existing machine

sxa555:
What do we need an ssh server config for?

cwesMills:
It is needed to setup cygwin ssh.

cwesMills:
I've just added #589 so windows playbooks will install nasm.

sxa555:
> It is needed to setup cygwin ssh, so the jenkins master can ssh into the machine.
At the moment the adoptopenjdk machines start the jenkins slave as a service, not via ssh so I'm not sure at the moment that I want another port exposed on the adoptopenjdk windows servers.

jdekonin:
[...]

PR #625, opened 2018-12-06
Updated cygwin role

cwesMills:
Cygwin will now be added to the system path instead of
the User path.


cwesMills:
Related #553

jdekonin:
The user path being Administrator on windows at the time the playbook is applied.

PR #628, opened 2018-12-07
Fixed 7-zip role

cwesMills:
7-zip role was not being executed by ansible, so tasks
directory was created and main.yml moved inside. Added
the /S option to the 7-zip installation so it wouldn't
hang waiting for input.


cwesMills:
Related #553

PR #633, opened 2018-12-11
Updated Freetype role

cwesMills:
Changed checks for when to build freetype as they were
skipping tasks constantly. Also Changed syntax of win_shell
command for building freetype so it will work properly.


cwesMills:
Related #553

AdamBrousseau:
@karianna I'm confused. Admittedly I'm not an expert on the Travis build. My understanding was it's a syntax checker. So I don't understand why changing the freetype yml causes errors to be thrown in all the other files when they were previously green.

karianna:
Yeah I'm also confused, it should not have behaved that way.  I'm happy to let it slide and we can sort it out on the next PR

Issue #619, opened 2018-12-05
Synchronise hostnames across jenkins/inventory.yml/anywhere else


sxa:
Example: [test-osuosl-ppc64le-ubuntu-16-04-1](https://ci.adoptopenjdk.net/computer/test-osuosl-ppc64le-ubuntu-16-04-1/) is shown as `test-osuosl-ubuntu1604-ppc64le-1` in [inventory.yml](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/inventory.yml)

There are likely more examples. Makes it awkward to run the inventories against a specific machine when someone requests it (generally with a name from Jenkins)

HusainYusufali:
Pull request opened to update the inventory.yml file to match the jenkins hosts. https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/634

HusainYusufali:
Below is a list of changes that need to be created in order to synchronize the hosts across Jenkins and the Inventory.yml. @sxa555 can you please review this before I go ahead and make the changes

| Current Name In Jenkins  | New Name in Jenkins |
| ------------- | ------------- |
| build-macstadium-macos1010-1 | build-macstadium-macos1010-x64-1 |
| build-macstadium-macos1010-2 | build-macstadium-macos1010-x64-2 |
| test-macincloud-macos1010-1 | test-macincloud-macos1010-x64-1 |
| test-macincloud-macos1010-2 | test-macincloud-macos1010-x64-2 |
| build-osuosl-ppc64-aix-71-1 | build-osuosl-aix71-ppc64-1 |
| build-osuosl-ppc64-aix-71-2 | build-osuosl-aix71-ppc64-2 |
| test-osuosl-ppc64le-ubuntu-16-04-1 | test-osuosl-ubuntu1604-ppc64le-1 |
| test-osuosl-ppc64le-ubuntu-16-04-2 | test-osuosl-ubuntu1604-ppc64le-2 |
| test-packet-armv8-ubuntu-16-04 | test-packet-ubuntu1604-armv8-1 |
| test-packet-x64-windows-2012r2-1 | test-packet-win2012r2-x64-1 |
| build-marist-s390x-sles-12 | build-marist-sles12-s390x-1  |
| build-marist-s390x-zOS-1 | build-marist-zos21-s390x-1 |
| build-marist-s390x-zOS-2 | build-marist-zos21-s390x-2 |
[...]

PR #634, opened 2018-12-11
Synchronise hostnames in inventory.yml to jenkins host

HusainYusufali:
The hostnames in the inventory.yml have been changed to match the hostnames in the jenkins inventory. 

Fixes: #619


gdams:
closing as this is the wrong approach, we should work to update the hosts to all use the schema defined here

Issue #635, opened 2018-12-14
Update Windows playbooks with locales to support MBCS testing


toshiona:
Related to new test cases for MBCS testing. The new playbook to add Windows Language pack (Japanese or Korean) if host variable has lang=ja or lang=ko.
 
Also related to infra request #627

toshiona:
Withdrawing the request.

We tried several ways to use non-English environment without changing default, but we couldn't.
So, we will try to change test case side to cover the area. They couldn't cover 100%, but can be 60% at the current assumption.
Thank you for your supports, anyway.

PR #636, opened 2018-12-14
Windows LanguagePack installation

toshiona:
For new internationalization tests, install Windows Language pack and set the language.
In this phase, Japanese or Korean will be set up if Host variable has "lang=ja" or "lang=ko".

Reboot is required twice; after Language Pack install and after switching to new language.

/Vendor_Files/langpack/Win2012R2/jp/lp.cab
/Vendor_Files/langpack/Win2012R2/kr/lp.cab
They need to be extracted from Windows Language Pack media.

Reference: #635

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -52,3 +52,4 @@
     - Clang_32bit                 # OpenJ9
     - OpenSSL                     # OpenJ9
     - nasm                        # OpenJ9
+    - LangPack                    # Internationalization tests
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/LangPack/tasks/ja.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/LangPack/tasks/ja.yml
@@ -1,0 +1,35 @@
+---
+###############################
+# Switch language to Japanese #
+###############################
+
+# Prerequisite: Japanese language pack was installed
+- name: Test if language pack ja is installed
+  win_shell: |
+      DISM.exe /Online /Get-Intl /English |
+      Select-String -SimpleMatch 'Installed language(s)'|
+      ForEach-Object { if($_ -match 'ja-JP'){ $Matches[0] }}
+  register: lang_installed
+  failed_when: lang_installed.stdout | length == 0
+  tags: langpack
+
+- name: Set Languagelist
+  win_shell: Set-WinUserLanguageList -LanguageList ja-JP,en-US -Force
+  tags: langpack
+
+- name: Overwrite UI language
+  win_shell: Set-WinUILanguageOverride -Language ja-JP
+  tags: langpack
+
+- name: Region Setting
+  win_region:
+    copy_settings: "true"
+    format: ja-JP
+    unicode_language: ja-JP
+  register: result
+  tags: langpack
+
+- name: Change Region - Reboot
+  win_reboot:
+  when: result.restart_required
+  tags: langpack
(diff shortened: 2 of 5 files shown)


sxa555:
Can you add something (a tag most likely!) to ensure that this can be skipped for people who don't need it and don't have easy access to the appropriate files.

toshiona:
Thank you for the comment. I added tags, "langpack".

sxa555:
@toshiona Does this still need to be marked WIP?

toshiona:
@sxa555 Thank you for your review. I've updated my patch to adjust lint. It seems ansible-lint was updated recently and lots of errors are reporting. I believe they are not related to my latest patch.

toshiona:
I reported #641 for new ansible-lint's new errors.

sxa555:
Need to think about this ... It looks like the language applet in the control panel is able to auto-download the right thing when you install the extra languages so that might be a better option than trying to grab an ISO. Not a clue at the moment how you automate that though ;-)

toshiona:
@sxa555 Yes, GUI control panel can download the language pack. However, I couldn't find the way by command or ansible module, either.
(Windows Server 2016 can do that by "DISM /Online /Add-Capability" option, but it's not available on 2012R2.)

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/LangPack/tasks/ja.yml
```diff
@@ -0,0 +1,77 @@
...
+  win_shell: |
+      DISM.exe /Online /Get-Intl /English |
+      Select-String -SimpleMatch 'Installed language(s)'|
+      ForEach-Object { if($_ -match 'ja-JP'){ $Matches[0] }}
+  register: jp_installed
+  tags: langpack
+
+- name: install language pack ja
+  win_shell: "Dism /online /Add-Package /PackagePath:{{ JP_cab }}"
+  when: (jp_installed.stdout | length == 0) and JP_cab is defined
```
jdekonin: If JP_cab is undefined at this point, the playbook fails, example my windows8 install.  Please add a when JP_cab is defined here.

toshiona: Thanks you for the comment, but I'm afraid the last condition is the one.

jdekonin: I have no idea how I missed that!  Sorry.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
```diff
@@ -53,3 +53,4 @@
     - Clang_32bit                 # OpenJ9
     - OpenSSL                     # OpenJ9
     - nasm                        # OpenJ9
+    - LangPack
```
karianna: Can we add a comment as to why this is needed please.

toshiona: @karianna Yes, updated.

toshiona:
@sxa555 
How about this way for non-English environment of Windows?
- Remove the language pack installation part from this PR.  (assuming language pack has been installed separately.)
- Adding another playbook to back to English environment.
Then, we can switch language by the ansible variable.

sxa555:
@toshiona As long as we can make English the default that would be a good solution :-)

toshiona:
@sxa555 Unfortunately, Windows can have only one language for testing...
I think all I can do now is to prepare language switching playbook, which is the latest committing in the PR.

With the playbook, we can run the following steps when the specific language is required.
1. Run ansible-playbook with the variable lang=ja (or lang=ko)
2. Run some internationalization tests
3. Run ansible-playbook with the variable lang=en

I believe this PR has no effect without lang variable.

toshiona:
Withdrawing the request.

We tried several ways to use non-English environment without changing default, but we couldn't.
So, we will try to change test case side to cover the area. They couldn't cover 100%, but can be 60% at the current assumption.
Thank you for your supports, anyway.

PR #640, opened 2018-12-17
Add JDK12 prereqs libXi libXrandr to linux playbooks

### LINKED ISSUE #639: Add JDK12 prereqs to Ansible
Need some more header files on the build systems. These files were previously available in the OpenJDK source but from the latest JDK 12 code these files are not provided in their source, so they need to be available on the machine that is building OpenJDK.

Example failed configure output from a build on Cent6 x86-64
```
00:06:35 checking for X11/Intrinsic.h... yes
00:06:35 checking for X11/extensions/Xrandr.h... no
00:06:35 configure: error: Could not find all X11 headers (shape.h Xrender.h Xrander.h XTest.h Intrinsic.h). You might be able to fix this by running 'sudo yum install libXtst-devel libXt-devel libXrender-devel libXrandr-devel libXi-devel'.
00:06:35 configure exiting with result code 1
```

AdamBrousseau:
Fixes #639

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
@@ -30,6 +30,8 @@ Build_Tool_Packages:
   - libdwarf-devel
   - libpng-devel
   - libXext-devel
+  - libXi-devel # JDK12+ compilation
+  - libXrandr-devel # JDK12+ compilations
   - libXrender-devel
   - libXt-devel
   - libXtst-devel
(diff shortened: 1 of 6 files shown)


AdamBrousseau:
@sxa555 for review

karianna:
@AdamBrousseau I'm happy with this, but the old question of documenting "why we are adding these" still remain.  Can you start a comment block at the top of the `Build_Tool_Packages:` section?  Let's start boy/girl scouting this as we go.

sxa555:
@karianna I think adding a block at the start is likely to just become a headache - IMHO we should add a (lined up) comment to the end of each line which I think will be preferable e.g.
```
  - libXrandr-devel                    # JDK12 build PR640
```

AdamBrousseau:
Added the comments. I had this version drafted before Stewart's suggestion was made. Let me know which you prefer and I can update if necessary.

PR #644, opened 2018-12-21
[WIP] ansible: Reference correct default PATH to 7-zip when extracting mingw

sxa555:

WIP as needs to be tested.
Alternative solution is to adjust the 7-Zip role to install to `C:\7-Zip` (probably the better solution ... This is really just a placeholder to ensure someone comes up with a solution!


HusainYusufali:
I will be changing the 7-Zip role to install 7-Zip into the `C:\7-Zip` directory instead of modifying mingw

HusainYusufali:
Pull request #653 has changed the 7-Zip role to make the install directory `C:\7-Zip`

karianna:
#653 has landed, does this PR now need to change?

sxa555:
Yep just needs to be canceled :-)

PR #649, opened 2019-01-04
ansible: Add JDK10 boot JVM to aarch64 machines

sxa555:

Currently using an automatically downloaded version into the workspace which is just silly.


sxa555:
Lint errors are not as a result of this update so to any reviewers: please ignore those (Will fix them under separate PR)

sxa555:
Merging as this fixes the final lint problems

PR #657, opened 2019-01-09
Update AIX playbooks with locales to support MBCS testing

toshiona:
Separated from Linux part (#624).

ref #618

### DIFF
--- a/ansible/playbooks/adoptopenjdk_variables.yml
+++ b/ansible/playbooks/adoptopenjdk_variables.yml
@@ -9,6 +9,7 @@ Slack_Notification: Disabled
 Superuser_Account: Enabled
 jckftp_Username: jckftp
 jckftp_Passwd: /Vendor_Files/passwords/ftp
+Asian_Locales: Disabled
 # Jcktestr User Variables:
 Jcktestr_Username: jcktestr
 Jcktestr_User_SSHKey: /Vendor_Files/keys/keybox.key
--- a/ansible/playbooks/aix.yml
+++ b/ansible/playbooks/aix.yml
@@ -759,6 +759,72 @@
             # Environment variables don't work as part of command
             - skip_ansible_lint
 
+      #######################################
+      # Install Asian Locales               #
+      # Locale filesets are in AIX OS media #
+      #######################################
+        - name: Get locale list
+          command: locale -a
+          register: localeList
+          changed_when: False
+          when: Asian_Locales == "Enabled"
+          tags: locales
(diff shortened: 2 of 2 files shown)


toshiona:
@jdekonin Thank you for the comment.
OK, "Asian_Locales" variable was introduced. 
I'm not sure only aix.yml and adoptopenjdk_variables.yml were enough to update.

The following filesets are from AIX Base OS media (I don't think they are downloadable.)
I expected "/Vendor_Files/aix" folder can be used for that purpose, but it's still unclear that how we can place them.
- bos.iconv
- bos.loc.com.CN
- bos.loc.com.utf
- bos.loc.iso.ja
- bos.loc.iso.ko
- bos.loc.iso.zh
- bos.loc.pc
- bos.loc.utf.JA_JP
- bos.loc.utf.KO_KR
- bos.loc.utf.ZH_CN
- bos.loc.utf.ZH_TW

Due to my operation mistake, other change were listed, but the diff result is correct.

sxa555:
@jdekonin @karianna I am *NOT* comfortable with this going in since it requires copies of licensed packages from the AIX media to be distributed via our ansible server and I'm not clear that we can do that. Ref comment in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/624#discussion_r246078852 which is why I had this split out from the Linux PR

karianna:
> @jdekonin @karianna I am _NOT_ comfortable with this going in since it requires copies of licensed packages from the AIX media to be distributed via our ansible server and I'm not clear that we can do that. Ref comment in [#624 (comment)](https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/624#discussion_r246078852) which is why I had this split out from the Linux PR
> (And since they're not currently there, this new function will not work)

Hmm, that's less than ideal.

@toshiona - Really sorry to ask this, but can you create a reverse PR to pull this back out for now please?

toshiona:
Oh, OK. I created PR #688 to back out this.

karianna:
@toshiona - Thank you!  @sxa555 Are you able to work with @toshiona to get this change back into AIX in the appropriate manner?

sxa555:
@karianna Unfortunately I suspect the answer needs to be that whoever does the OS installation needs to handle it on request so it may not be possible to easily add it to the playbooks in the way we'd like, but I will continue to investigate options.

toshiona:
@sxa555 Is it possible to request administrators to install these internationalization filesets manually?

PR #660, opened 2019-01-09
[WIP] Split out AIX playbooks to use individual roles for operations like Linux

sxa:


sxa555:
`Assigning to @sej-jackson as she might be able to make use of this in order to finalise it properly :-)

Haroon-Khel:
@sej-jackson Has there been any progress on resolving this? There also seems to be some minor syntax errors which is causing the travis check to fail

sxa:
This has suffered too much code rot in the past 18 months ... Closing so someone else can start again ;-)

PR #662, opened 2019-01-09
ansible: Remove OracleLinux from Unix 'Common' role

sxa555:

We don't have any Oracle Linux systems in the adoptopenjdk machine farm and so having this just results in duplication being required between RHEL/CentOS/Oracle with a risk they are getting out of sync due to not being tested, therefore removing.


karianna:
We can put these back in if someone outside of our build farm is using them, but since they were bitrotting, I doubt it.

PR #670, opened 2019-01-11
ansible: Remove deprecation warnings regarding version_compare

sxa555:


sxa555:
Seems ok - now ready for approval

jdekonin:
What version will this be compatible with?  What version of AWX does Adopt use?

sxa555:
It's based on ansible 2.4.0.0 - haven't explicitly checked with that version yet though.

karianna:
> It's based on ansible 2.4.0.0 - haven't explicitly checked with that version yet though.

What level is our linter set at? If it's 2.4.x then we can land this...

sxa555:
I'm not sure it has a specific version defined, although the fact the checks haven't tripped up suggest it's ok unless you have any specific concerns I'm missing @karianna

PR #675, opened 2019-01-15
ansible: Create /usr/bin/cpanm link on CentOS6/RHEL6/SLES

sxa555:

Without this cpanm isn't necessarily in the PATH when the playbooks are run on those platforms
Part of enabling CentOS/Vagrant support as per https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/673


jdekonin:
That sounds like we should be setup in the playbook environment then??  Something like whats been done in the AIX playbook.

```
  environment:
    PATH: "/usr/local/bin:{{ ansible_env.PATH }}"
``

sxa555:
That would be the other option. I'm in two minds ...

sxa555:
Have decided that's the better option to avoid any potential to disrupt things outside `/usr/local` - adjusted accordingly

PR #677, opened 2019-01-15
WIP: Add ppcle and s390 openssl 1.1.0g support

jdekonin:


jdekonin:
Further discussion with OpenJ9 dev on the requirement and this isn't what is wanted.  They would like the default openssl version installed via package manager for each o/s release, so I am closing this down.

Issue #673, opened 2019-01-11
Allow Vagrantfile to test playbooks on CentOS6


sxa555:
Since this is our production build environment it would be useful for end users to be able to test changes before committing on CentOS6.0 (provider `generic/CentOS6`

PR #679, opened 2019-01-16
ansible: Use installed libcurl on CentOS/RHEL6 to compile git

### LINKED ISSUE #676: git being built on CentOS without http/https support
Tried in my vagrant system as a result of #673 - the libcurl-devel package on the system does not appear adequate for building git with https support
```
checking for curl_global_init in -lcurl... no
```
by explicitly running configure with `--with-curl=/usr/local/curl-7.61.1` we get the following which results in a version that works. The standard libcurl in the CentOS6 packages is 7.19.7-53:
```
configure: Setting CURLDIR to /usr/local/curl-7.61.1
checking for curl_global_init in -lcurl... yes
```
Logging this at the moment for more investigation later e.g. do we need curl into /usr/local/curl-7.61.1 specifically instead of straight into a prefix of /usr/local?

sxa555:

Fixes #676 ... part of #673 
Need to verify that the ansible logic is correct before merging

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
@@ -54,13 +54,23 @@
     # TODO: Package installs should not use latest
     - skip_ansible_lint
 
-- name: Compile and install git from source
+- name: Compile and install git from source on RHEL/CentOS6
+  shell: cd /tmp/git-2.15.0 && ./configure --with-curl=/usr/local/curl-7.61.1 --prefix=/usr/local --without-tcltk && make clean && make -j {{ ansible_processor_vcpus }} && sudo make install
+  become: yes
+  when:
+    - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout | version_compare('2.15', operator='lt') )
+    - ansible_architecture != "s390x"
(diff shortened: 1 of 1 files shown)


sxa555:
Seems ok ... Approvals welcome :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
```diff
@@ -54,13 +54,23 @@
     # TODO: Package installs should not use latest
     - skip_ansible_lint
 
-- name: Compile and install git from source
+- name: Compile and install git from source on RHEL/CentOS6
+  shell: cd /tmp/git-2.15.0 && ./configure --with-curl=/usr/local/curl-7.61.1 --prefix=/usr/local --without-tcltk && make clean && make -j {{ ansible_processor_vcpus }} && sudo make install
```
jdekonin: Should we be using sudo everywhere or nowhere?  I am fine with either I only ask because I have a systems both ways and I think without sudo I cannot run the playbook directly on the machine, but via AWX its fine.

sxa555: Should be without as it's connecting as root ... That was copied from the equivalent other lina already in the file - can I suggest creating a separate issue for this discussion and we'll apply the outcome globally across the playbooks instead of just addressing it here?

Issue #530, opened 2018-10-12
UB18 x86 ansible failures


jdekonin:
- azul repo errors
-- workaround...delete repo additions as well as zulu-9 package installation
- apt install errors with libmpfr4, libmpfr4-dbg, openjdk-7-jdk
-- workaround...delete from list
- docker role needs rework
-- workaround... run with `--skip-tags docker`

Note that I am using this machine for performing adoptopenjdk testing, and not compilation of openjdk or openj9-openjdk.  So there very well might be other issues when it comes to creating an SDK binary.

karianna:
* Azul Repo errors maybe an http vs https issue.
* What's the failure on `apt install errors with libmpfr4, libmpfr4-dbg, openjdk-7-jdk`?
* What sort of rework does the docker role need?

sxa555:
To be honest, Unless there's a reason to continue using the azul stuff I'd prefer to install ours now that we're close to having all the version coverage.

karianna:
We don't have a solution for OpenJDK 7.  I think adding keys is a good idea

sxa555:
@jdekonin Here's an example of adding a key: https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml#L56 although I'm surprised that what's in there doesn't cover you for docker.

sxa555:
> Note that I am using this machine for performing adoptopenjdk testing, and not compilation of openjdk or openj9-openjdk. So there very well might be other issues when it comes to creating an SDK binary.

We could do with having a way to run a subset of the installs for build / test / JCK purposes. The additional JCK playbook is entirely independent at the moment that those requirements would be best included in the main playbooks. It's on the todo list to merge them ;-)

sxa555:
> We don't have a solution for OpenJDK 7

Unless we bootstrap 8 with 8

karianna:
We could do for later 8 builds as no spec changes are likely to come in.

jdekonin:
Was azul added just for bootjdk for compiling JDK10?  If so, then I agree those entries could just be removed.

openjdk7 not being on the machine is only of importance if someone were to try compiling a JDK8 binary.  The build system compiles on centos6.9.

karianna:
[...]

PR #681, opened 2019-01-16
[WIP] Update Vagrant files to include Ubu16/18 and CentOS6 (default) support

### LINKED ISSUE #673: Allow Vagrantfile to test playbooks on CentOS6

sxa555:

Fixes #673. Note, this change results in there being suitable Vagrant files for Ubuntu 16.04 + 18.04 and CentOS 6. CentOS 6 will be the default by virtue of a "Vagrantfile" symlink that points to the CentOS6 one. If a user wants to select a different OS then they can adjust the symlink.

Alternate options would be to use a Vagrantfile in separate subdirectories, but I feel that this is reasonably clear and that CentOS6 (as the currently default AdoptOpenJDK Linux/x64 build OS) is the best option as a default, while making the others as simple for people to get running with as possible.

WIP mostly because Ubuntu 18.04's playbooks aren't working yet (Ref #530) but that doesn't necessarily stop is putting this in as it'll make it easier to test fixing that issue ;-)


sxa555:
`2` defines effectively a version of the file that the configuration is for. It came from what is in the Ubuntu1604 file (which was the original `Vagrantfile`)  . From the docs:

`Version 1 represents the configuration from Vagrant 1.0.x. "2" represents the configuration for 1.1+ leading up to 2.0.x.`

sxa555:
Documentation on this needs a general overhaul, especially now that we've got the better method for excluding all the adoptopenjdk stuff when you're running. I don't think the current docs are invalidated any further by this change ;-)

karianna:
> `2` defines effectively a version of the file that the configuration is for. It came from what is in the Ubuntu1604 file (which was the original `Vagrantfile`) . From the docs:
> 
> `Version 1 represents the configuration from Vagrant 1.0.x. "2" represents the configuration for 1.1+ leading up to 2.0.x.`

OK lets state that as a comment

karianna:
> Documentation on this needs a general overhaul, especially now that we've got the better method for excluding all the adoptopenjdk stuff when you're running. I don't think the current docs are invalidated any further by this change ;-)

I'm tempted to say last boy scout ;-) - do we have a separate ticket for a doc overhaul?

sxa555:
> > `2` defines effectively a version of the file that the configuration is for. It came from what is in the Ubuntu1604 file (which was the original `Vagrantfile`) . From the docs:
> > `Version 1 represents the configuration from Vagrant 1.0.x. "2" represents the configuration for 1.1+ leading up to 2.0.x.`
> 
> OK let's state that as a comment

Done

sxa555:
> I'm tempted to say last boy scout ;-) - do we have a separate ticket for a doc overhaul?

No but it was in my head ... I'll create one now ...

PR #682, opened 2019-01-16
Ansible: Change the syntax of removing and adding cygwin to PATH

HusainYusufali:
* Add the name of the environment variable
* Make sure that the PATH is available on the entire machine's users instead of just the Administrator user
* This change gets rid of an error when the cygwin path portion of the playbook is rerun against a machine

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -76,14 +76,23 @@
 
 - name: Remove c:\cygwin64\bin from the path if it exists
   win_path:
+    name: PATH
     elements:
       - 'C:\cygwin64\bin'
+    scope: machine
     state: absent
   tags: cygwin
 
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -76,13 +76,20 @@
...
+    name: PATH
     elements:
       - 'C:\cygwin64\bin'
+    scope: machine
     state: absent
   tags: cygwin
 
-- name: Add c:\cygwin64\bin to the start of the %PATH%
-  raw: setx PATH "C:\cygwin64\bin;$env:Path" /m
+- name: Add c:\cygwin64\bin to %PATH%
```
jdekonin: setx is needed to make sure cygwin is always on the front of the path.  If other tools install/write to the front of the path, like openssh, cygwin won't work as desired.

HusainYusufali: Using the `win_shell` module with `cmd` allows the setx command to work without causing errors on playbook reruns.

PR #687, opened 2019-01-25
ansible: Stop installing MinGW installation as OpenJ9 has moved to Clang

sxa555:

Last bit which fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/700


jdekonin:
> Will this impact us trying to build Java 8 versions or if we ever had to respin a 9 or 10?

OpenJ9 has moved to clang for Java8 as well, but you are correct a respin of 9 or 10 would require mingw.  What are the expectations on performing a respin on 9 or 10 though??

Note that these changes won't remove it from existing installations, just impact new installation.

PR #698, opened 2019-02-12
Don't upgrade cmake if it's already at 3.7.2

sxa555:
We are installing 3.11.4 for OpenJFX, but that version is a bit of a pain to install on RHEL6.9, therefore we should tolerate having 3.7.2 for now, but install 3.11.4 for now unless it can be proven
that OpenJFX can't be built with 3.7.2.


sxa555:
Alternate options:
- Special case RHEL6.9 
- Fix 3.11.4 build on RHEL69

PR #702, opened 2019-02-22
Ansible: Add s390x and ppcle64 to RHEL, CENT, and SLES docker tasks

HusainYusufali:
- <s>Checking if docker is installed task was not working correctly</s>
- <s>New command uses `docker version` command to check if it is installed</s>
- Remove `docker_installed` check from `Add Jenkins user to the docker group` task as it's not required

- Add `s390x` and `ppcle64` architectures to RHEL, CENT, and SLES docker tasks

@jdekonin 

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
@@ -93,7 +93,7 @@
     - docker_installed.rc != 0
     - ansible_distribution == "RedHat"
     - ansible_distribution_major_version == "7"
-    - ansible_architecture == "x86_64"
+    - (ansible_architecture == "x86_64" or ansible_architecture == "s390x" or ansible_architecture == "ppc64le")
   tags: docker
 
 - name: Add Docker Repo for RedHat 7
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
```diff
@@ -3,7 +3,7 @@
 # Docker #
 ##########
 - name: Test if docker is already installed
-  shell: docker >/dev/null 2>&1
+  shell: docker version 2>&1 | grep Version | awk '{print $2}'
```
jdekonin: Not sure why the change from "is it installed?" to "obtain a version if installed".  There doesn't appear to be any use of the version number that I see.  Is that check failing as is?

HusainYusufali: The current check is just failing as is. The command uses docker version command and if there is an output that means docker is installed. 
image

jdekonin: The previous method was checking the return code of the command.  `echo $?` after `docker >/dev/null 2>&1` should give a `0` no?  Or maybe this is something that works though a shell and not ansible??

HusainYusufali: I just realized I messed up when testing. I thought that the docker check wasn't working, but it was because of other parameters that weren't there. I will revert it back

Issue #708, opened 2019-02-28
Ability to configure alternative SDK bootsdk installations for AIX


jdekonin:
It would be nice to be able to use a parameter to select either hotspot or openj9 sdks to be installed.  OpenJ9 project would prefer to use OpenJ9 as the bootjdks.  I propose that the default be hotspot based on variable definition that can be modified at runtime.

sxa555:
If this is AIX specific I'd be happy to make OpenJ9 the default on AIX for all builds. Shouldn't negatively impact others. Not sure if there's a specific reason why they want it though over "nice to be able to"

karianna:
Is Hotspot not supported on AIX? I assumed it was.

jdekonin:
Hotspot is supported on AIX.  The playbook is defaulting to OpenJ9.  #713 partially resolves.  @HusainYusufali can you update for JDK10 and 11 as well.  I should've caught that in the initial review sorry.  While you are touching the JDK11 section(s), can you please clean up the duplication?  Thanks.

karianna:
I'm OK swapping between the Eclipse OpenJ9 and Hotspot boot JDK's but as a general philosophy, I think that we should default to Hotspot unless there's a compelling reason to do otherwise.  Cuts down a source of difference.  Happy to be overruled if OpenJ9 should have an OpenJ9 boot JDK

HusainYusufali:
AIX is not supported by hotspot on JDK 10 & 11 as shown by trying to download the jdk from the following 2 links:

https://api.adoptopenjdk.net/v2/binary/releases/openjdk10?openjdk_impl=hotspot&os=aix&arch=ppc64&release=latest&type=jdk
https://api.adoptopenjdk.net/v2/binary/releases/openjdk11?openjdk_impl=hotspot&os=aix&arch=ppc64&release=latest&type=jdk

This is why OpenJ9 is the default for JDK 10 & 11 on AIX. Also, the pull request #718 has been submitted to remove the duplicate JDK 11 task.

sxa555:
@karianna Technically there's no problems using either HotSpot or OpenJ9 as the boot JVM, although for s390x I think there are some significant performance benefits to using the OpenJ9 one and the OpenJ9 team did request that we switch, but I'm not sure what the driver was for this request on AIX.

jdekonin:
[...]

PR #713, opened 2019-03-05
Add hotspot jdk and openj9 parameter to AIX

HusainYusufali:
- When using the AIX playbook, Java 8 & 9 by default is the hotspot jdk
- If OpenJ9 Java 8 & 9 are to be installed instead of hotspot the extra variable `bootjdk = openj9` must be specified at runtime
- Issue #708 

fyi @jdekonin 

### DIFF
--- a/ansible/playbooks/aix.yml
+++ b/ansible/playbooks/aix.yml
@@ -261,7 +261,7 @@
 
         - name: Transfer and Extract Java8
           unarchive:
-            src: https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=openj9&os=aix&arch=ppc64&release=latest&type=jdk
+            src: https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl={{ bootjdk }}&os=aix&arch=ppc64&release=latest&type=jdk
             dest: /usr/java8_64
             remote_src: yes
           when: java8.stat.isdir is not defined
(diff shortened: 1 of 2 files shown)


ansible/playbooks/aix.yml
```diff
@@ -251,20 +251,57 @@
...
-          when: java8.stat.isdir is not defined
+          when:
+            - java8.stat.isdir is not defined
+            - bootjdk == "openj9"
+          ignore_errors: yes
 
         - name: Transfer and Extract Java8
           unarchive:
             src: https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=openj9&os=aix&arch=ppc64&release=latest&type=jdk
             dest: /usr/java8_64
```
jdekonin: Could `openjdk_impl=openj9` be used as a variable to cut down on the amount of change?

HusainYusufali: Yeah that's a way better way to do it.

HusainYusufali:
I added the variable `bootjdk: hotspot` as the default. Unless the variable bootjdk is specifically assigned to openj9 it will get hotspot

ansible/playbooks/aix.yml
```diff
@@ -13,6 +13,7 @@
     PERL_MM_OPT: "INSTALL_BASE=/home/{{ Jenkins_Username }}"
   vars:
     swap_size: 4096
+    bootjdk: hotspot
```
jdekonin: Should load this from adoptopenjdk_variables.yml.

jdekonin:
Looks good, that one minor change of having the default in the variable file and I think its good.

ansible/playbooks/aix.yml
HusainYusufali: `bootjdk: hotspot` removed from `aix.yml` and added to `adoptopenjdk_variables.yml`

jdekonin:
Sorry I didn't see @sxa555 comment in #708.  It appears OpenJ9 is fine as the default for AIX.

HusainYusufali:
Changes were made in the `adoptopenjdk_variables.yml` to make the default jdk `openj9` instead of `hotspot`.

PR #717, opened 2019-03-06
Install Rust in Windows playbook

charliegracie:
Rust is required to compile IcedTea-Web. The Rust role will
ensure that Rust is installed and on the path.

Signed-off-by: Charlie Gracie <charlie.gracie@gmail.com>


charliegracie:
I will remove the Draft once I have verified it works. I should be able to test it shortly.

PR #718, opened 2019-03-07
Ansible: Remove duplicate jdk11 task on aix

HusainYusufali:
- Removed redundant installation of JDK 11

fyi @jdekonin 


sxa555:
Can you change the comments on all of the remaining ones to say `Boot JDK xx` as well please, since the reference to `IBM Java` is generally incorrect as the playbook is downloading them from adoptopenjdk.net

PR #719, opened 2019-03-07
Modify the install location of Rust

charliegracie:
The IcedTea-Web makefiles can not handle the default installer
path for Rust. The path appears to not be able to have spaces in
it. This change installs rust in `c:\rust' instead of the default.

Signed-off-by: Charlie Gracie <charlie.gracie@gmail.com>


charliegracie:
Sorry about missing this in the first PR. I guess my machine already had Rust installed to a location that worked for IcedTea-Web so I missed that the default install location in `c:\Program Files\` did not work in the IcedTea-Web builds.

Issue #714, opened 2019-03-05
Sort out yum repos on pLinux and zLinux boxes


sxa555:
docker-ce repo doesn't work on several p/z-Linux, EPEL failed on zLinux - details in https://github.com/AdoptOpenJDK/openjdk-build/issues/938

Machines have been reset to the "failure" state so probably won't install anything at the moment. Will need appropriate playbook adjustments which this issue can cover as I want to close the immediate build one.

sxa555:
docker-ce seems to have been added on `ppc64le` and `s390x` under https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/702 - @HusainYusufali @jdekonin Can you investigate / resolve please?

sxa555:
Kicked off another run via ansible on `build-marist-rhel74-s390x-2` but it failed as the docker-ce repo is already configured on the system.

Have issued: `rm -f /etc/yum.repos.d/docker-ce.repo`

sxa555:
Also renamed `/etc/yum.repos.d/epel*repo` as the update was failing:

```
[root@build-marist-rhel74-s390x-2 yum.repos.d]# yum update
Loaded plugins: langpacks, product-id, search-disabled-repos, subscription-
              : manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
epel/s390x/metalink                                     |  50 kB     00:00     
Could not parse metalink https://mirrors.fedoraproject.org/metalink?repo=epel-7&arch=s390x error was 
No repomd file
```

sxa555:
Now running through on build-marist-s390x-1 ... Had an issue with disk space, and have symlinked /usr/local/gcc to another location under /data but AWX has been able to complete the playbook execution there now

sxa555:
Playbook run successfully. Now doing the same on build-marist-rhel74-s390x-1 and removing docker-ce from build-osuosl-centos74-ppc64le-2 (but retaining epel on there as it's not causing a problem). No comparable updates were needed on build-osuosl-centos74-ppc64le-1 as it didn't have the `docker-ce.repo` in place (although it currently has a `docker.repo` pointing at http://ftp.unicamp.br/pub/ppc64el/rhel/7/docker-ppc64el which may need investigated/addressed )

sxa555:
On the two zLinux marist boxes I have issued the following:
```
mv /var/cache/yum /data/var-cache-yum
ln -s /data/var-cache-yum /var/cache/yum
```
[...]

PR #724, opened 2019-03-12
Revert "Ansible: Add s390x and ppcle64 to RHEL, CENT, and SLES docker tasks"

jdekonin:
Reverts AdoptOpenJDK/openjdk-infrastructure#702

Addresses #714


jdekonin:
Created #725 for follow up on adding that capability back in with working repos

Issue #715, opened 2019-03-05
Ability to configure alternative SDK bootsdk installations for Windows


jdekonin:
It would be nice to be able to use a parameter to select either hotspot or openj9 sdks to be installed. OpenJ9 project would prefer to use OpenJ9 as the bootjdks. I propose that the default be hotspot based on variable definition that can be modified at runtime.

Same change as #713 should work nicely.

HusainYusufali:
Pull request #742 opened to add this feature

PR #742, opened 2019-03-22
Ansible: Add bootjdk variable to install hotspot or openj9 on Windows

HusainYusufali:
This pull request fixes issue #715 

* Add bootjdk variable to adoptopenjdk_variables.yml

* Load adoptopenjdk_variables.yml into the main.yml file

* Change hard link to API when downloading Windows JDK

* API downloads the latest Windows hotspot or openj9 jdk for Java[8,9,11]

* Remove versions from test cases and zip files as they can change across playbook deployments

* Powershell commands that create persistent environment variable of Java folder that is installed

* Create symlink using environment variable

* NOTE: win_shell module acts differently when creating environmental variable vs creating a symlink

fyi @jdekonin

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
@@ -4,46 +4,52 @@
 ##########
 - name: Test if Java 8 is already installed
   win_stat:
-    path: 'C:\Program Files\Java\jdk8u172-b11'
+    path: 'C:\openjdk\jdk-8\bin'
   register: java8_installed
   tags: Java8
 
 - name: Check if Java 8 is already downloaded
   win_stat:
-    path: 'C:\temp\jdk8u172-b11.zip'
+    path: 'C:\temp\jdk-8.zip'
   register: java8_download
   tags: Java8
 
 - name: Download Java 8
   win_get_url:
-    url: https://github.com/AdoptOpenJDK/openjdk8-releases/releases/download/jdk8u172-b11/OpenJDK8_x64_Win_jdk8u172-b11.zip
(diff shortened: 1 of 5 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
```diff
@@ -4,46 +4,52 @@
 ##########
 - name: Test if Java 8 is already installed
   win_stat:
-    path: 'C:\Program Files\Java\jdk8u172-b11'
+    path: 'C:\openjdk\jdk8u\bin'
```
karianna: I would possibly just go with jdk8 to be consistent (the other bootjdk dirs don't seem to use the `u`)

HusainYusufali: I changed all instances of `jdk8u` to `jdk8` and tested it to make sure it works

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
```diff
@@ -4,26 +4,26 @@
 ##########
 - name: Test if Java 8 is already installed
   win_stat:
-    path: 'C:\Program Files\Java\jdk8u172-b11'
+    path: 'C:\openjdk\jdk8\bin'
```
jdekonin: Whatever the name format, it should be consistent for 8,9,11; the others look more like this jdk-8

HusainYusufali: All instances of `jdk8` have been changed to `jdk-8` and tested to make sure playbook deployment still works

PR #744, opened 2019-03-22
Ansible: Add Windows role Java12

HusainYusufali:
* Add Java12 to main.yml

* Add new role Java12

Pull request #742 must be merged first before this commit will work correctly. The new syntax in this role depend on changes from that pull request

fyi @jdekonin

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -41,7 +41,8 @@
     - Java7                       # JDK8 build bootstrap
     - Java8                       # JDK9 build bootstrap
     - Java9                       # JDK10 build bootstrap
-    - Java11                      # JDK11 buid bootstrap
+    - Java11                      # JDK11 build bootstrap
+    - Java12                      # JDK12 build bootstrap
     - ANT                         # Testing
     - MSVS_2013
     - MSVS_2017                   # OpenJ9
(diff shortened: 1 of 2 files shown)


sxa555:
What's the requirement for this (https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/745 doesn't explain it) - I don't think we're building anything that needs a boot JDK12 at the moment

jdekonin:
OpenJ9 is going to be building JDK13 at some point in the near future, this is just having the framework in place before the requirement.

gdams:
can we not look at using the MSI installers now? https://adoptopenjdk.net/releases.html#x64_win

sxa555:
@gdams We could, but I'm not sure there's any significant advantage to it, and I personally have a preference for being consistent across versions.

gdams:
well I guess it allows us to test our installers? I assume on linux and mac that we'll use our rpm/deb/brew files?

HusainYusufali:
Can you download the installer via the API and are there silent install options?

gdams:
yes the installer is available via the API. And yes it can be silently installed. see https://github.com/AdoptOpenJDK/openjdk-installer/tree/master/wix#deploy-via-active-directory-gpo

jdekonin:
Is the installer going to modify system parameters?  If its testing of the installers that is being looked for it would make more sense to me for this testing to be happening through a different process.  One that can be triggered outside the application of a machine installation.

jdekonin:
I suggest if an alternative method of adding the SDK to the system is desired then a new issue be opened to address.

Issue #747, opened 2019-03-29
Ansible: Change lib install location for Freetype 32-bit on Windows


HusainYusufali:
The location of freetype libraries on windows build scripts is `/cygdrive/c/openjdk/freetype-2.5.3/lib` for 32-bit builds. The real location of the freetype library for 32-bit builds is `/cygdrive/c/openjdk/freetype-2.5.3/lib32`. A symlink is needed to point to this real location

sxa555:
@ali-ince Presumably your recent changes render this irrelevant?

ali-ince:
Yes @sxa555, we can now remove all freetype relevant stuff from the playbook as it's now part of the build.

I can create a PR later today with the relevant changes.

karianna:
Superseded by https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/766

PR #748, opened 2019-03-29
Ansible: Change lib install location for freetype 32 bit on Windows

HusainYusufali:
* create a symlink to point `/cygdrive/c/openjdk/freetype-2.5.3/lib` to `/cygdrive/c/openjdk/freetype-2.5.3/lib32`

* symlink was created because `/cygdrive/c/openjdk/freetype-2.5.3/lib` is the freetype location for 32-bit builds in windows build scripts

Fixes issue #747 
fyi @sxa555 

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freetype/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freetype/tasks/main.yml
@@ -42,7 +42,7 @@
 
 - name: Check if freetype 32bit has been built
   win_stat:
-    path: 'C:\openjdk\freetype-2.5.3\lib32'
+    path: 'C:\openjdk\freetype-2.5.3\lib'
   register: freetype32_built
   tags: Freetype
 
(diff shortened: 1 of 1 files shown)


sxa555:
> Why would we add to the machine configuration if its the build script that is pointed at the incorrect location? I would think that would be a better spot no??
Or standardising on what is already working i.e. adjust the playbooks to install into `lib` instead of `lib32`? I'm definitely not keen on having two "options" available - we need to standardise on just one.

HusainYusufali:
Okay, changes will be made to install into `lib` instead of `lib32`

ali-ince:
Btw, with [#994](https://github.com/AdoptOpenJDK/openjdk-build/pull/994) we're enabling during the build build of freetype for hotspot. I don't know if the same is possbile for openj9 but just wanted to give an update.

jdekonin:
So what this change effectively does is force other systems in place to adjust their build scripts to conform to adopt openjdk.  I won't stand in the way of this change being merged, but I do object as this will cause additional workload for openj9 build and test systems to be updated to this new location.

HusainYusufali:
The [`windows.sh`](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/platform-specific-configurations/windows.sh) build script uses both `lib` and `lib32`. So either way one of these will have to be changed in the build script. 

I believe the easiest way to do this is by not changing the current Freetype role and continue using `lib32`. The one word in [`windows.sh`](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/platform-specific-configurations/windows.sh) at line 39 column 205 will need to be changed from `lib` to `lib32`. Finally, the Windows machines will need to rename the directory `/cygdrive/c/openjdk/freetype-2.5.3/lib` to `/cygdrive/c/openjdk/freetype-2.5.3/lib32`   

I have successfully built JDK8 32-bit using the original Freetype role and changing the `windows.sh` file.

sxa555:
@HusainYusufali The references to the directories have now been removed from `windows.sh` so no updates should be required. Is there a reason why we need this change to go in now?

HusainYusufali:
No this pull request is no longer needed

PR #750, opened 2019-04-01
Ansible: Fix OpenSSL installation on Windows

### LINKED ISSUE #749: Ansible: Fix OpenSSL installation on Windows
OpenSSL is not installing when deploying the Windows playbook even though no errors are reported. The tasks to fix this should be unarchiving openssl, changing the installation method, and add symlinks to reroute the build scripts OpenSSL location to the new OpenSSL.

HusainYusufali:
* Fix OpenSSL checks (downloaded, 32 and 64 bit installed, and symlinks created)

* Change method of unarchiving openssl-1.1.1.tar.gz

* Change method of installing and compiling OpenSSL

* Create OpenSSL symlinks

Fixes #749 

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
@@ -3,12 +3,14 @@
 # OpenSSL #
 ###########
 - name: Check if OpenSSL 32bit installed
-  win_shell: Test-Path -Path C:\openjdk\OpenSSL-1.1.1-x86_32
+  win_stat:
+    path: C:\openjdk\OpenSSL-1.1.1-x86_32
   register: openssl32_installed
   tags: openssl
 
 - name: Check if OpenSSL 64bit installed
-  win_shell: Test-Path -Path C:\openjdk\OpenSSL-1.1.1-x86_64
+  win_stat:
+    path: C:\openjdk\OpenSSL-1.1.1-x86_64
   register: openssl64_installed
   tags: openssl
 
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -2,54 +2,96 @@
...
+  register: openssl64_symlink
+  tags: openssl
+
 - name: Download OpenSSL-1.1.1
   win_get_url:
     url: https://www.openssl.org/source/openssl-1.1.1.tar.gz
     dest: C:\temp\openssl-1.1.1.tar.gz
-  when: (openssl32_installed == False) or (openssl64_installed == False)
+  when: (openssl_downloaded.stat.exists == false)
   tags: openssl
```
jdekonin: This is going to trigger a download every time run, whether the installed check is positive or not.

What problem does this resolve?  Why replace the use of win_shell at the top of the file but continue to use at the bottom?

HusainYusufali: It shouldn't trigger a download every time. It uses a the `openssl_downloaded` check instead of `openssl_installed` checks. I did this because the download would trigger every time openssl wasn't installed even though the package was downloaded. This would cause a playbook crash because the package was already there.

The `win_shell` checks were not giving a false output which skipped the installation of OpenSSL. `win_shell` was used at the bottom because the module `win_unzip` only works for zip files and installing OpenSSL had to be done through `cmd`

sxa555: Also what do we need the symlinks for? Since the target names don't have spaces in them I'm not sure what benefit they provide.

HusainYusufali: The build scripts point to `C:\progra~1\OpenSSL` and `C:\progra~2\OpenSSL` when looking for OpenSSL which doesn't exist

jdekonin: So `C:\temp\openssl-1.1.1.tar.gz` isn't deleted at the end of the role, I was going on the assumption that it would've been.  Leading to the next question...why isn't `C:\temp\openssl-1.1.1.tar.gz` being cleaned up?  Its no longer required after the install.

HusainYusufali: Just forgot to add the final cleanup of all the source files

HusainYusufali: Changes have been made to cleanup the source files

sxa555: I think we need to modify the build scripts. That was the location that OpenSSL was initially installed into for OpenJ9 as it is the default location that the source will install into (See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/526)

I would suggest either leave it as the default, or modify the build scripts to point to the "new" location consistent with the playbooks, as creating symlinks out of `C:\Program Files` looks really messy.

HusainYusufali: I think we should modify the build scripts and remove the symlinks. Would this change to [windows.sh](https://github.com/HusainYusufali/openjdk-build/commit/50a1748183f26ba4e85a77f865dad96d6fc06652) be sufficient?

sxa555: Should be yes :-)
If/when you create a PR from that branch can you assign it to me, mention this PR, and mark it `DO NOT MERGE` since I don't want to integrate it until we've actually made the change on all windows build machines for obvious reasons ;-)

HusainYusufali: This is the [PR]( https://github.com/AdoptOpenJDK/openjdk-build/pull/1028) I created. I am unable to assign it to or add labels to it. I am also removing the symlinks on this PR

jdekonin: At OpenJ9, we found that jobs running through jenkins connecting via cygwin ssh do not like spaces in the path.  The symlink was added to work around this issue.

jdekonin: Seems I got the directory backwards as @HusainYusufali clarified.  So ignore my last post.

sxa555: @jdekonin once this goes in presumably some work will be needed on the openj9 build side to match the "new" directory too?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -3,54 +3,80 @@
...
 
-- name: Cleanup OpenSSL
+- name: Cleanup OpenSSL source files
   win_file:
-    path: C:\temp\openssl-1.1.1
+    path: C:\temp\{{ item }}
     state: absent
+  with_items:
+    - openssl-1.1.1.tar.gz
```
sxa555: Relatively minor nit but can we avoid untarring twice by leaving the .tar file from the first `7z x` between 32-bit and 64-it installs, and cleaning it up at the end?

jdekonin: Good catch.  Additionally, is there a configure or make option to reset or clean such that the unpacking step is only done once.

HusainYusufali: The new changes I made only extract the files once and clean them all at the end. The `openssl-1.1.1` directory is cleaned by adding the `nmake -f makefile clean` to be used for both installations

jdekonin: Looks good, thanks for simplifying.

PR #754, opened 2019-04-05
Updates/cleanups to support Ubuntu 18.04

sxa555:
Fixes (at least partially, maybe not the docker, but should be ok for my aarch64 Ubuntu 18.04 system) https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/530

On Ubuntu 18 this removes openjdk-7 (We can bootstrap with 8 if needed, and adoptopenjdk builds on CentOS/RHEL so not an issue for us) and `libmpfr4` and `libmpfr4-dbg` (although `libmpfr-dev` is still installed, which may pull in any other bits this is needed for anyway)

If any of the missing things end up being needed we can add them back in later, but these changes won't destabilise the current adoptopenjdk infrastructure.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -74,6 +74,8 @@ hosts:
 
       - azure:
           win2012r2-x64-1: {ip: 13.68.134.204, user: adoptopenjdk}
+          win2012r2-x64-2: {ip: 40.121.164.56, user: adoptopenjdk}
+          win2012r2-x64-3: {ip: 13.68.219.237, user: adoptopenjdk}
 
       - osuosl:
           ubuntu1604-ppc64le-1: {ip: 140.211.168.227, user: ubuntu}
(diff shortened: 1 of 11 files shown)


sxa555:
@jdekonin How so? The OpenJFX stuff needs to be different between U16 and U18 so is consistent, and the s390x stuff was moving definitions to make them consistent across the Ubuntu versions instead of having term separate so I'd argue they are part of the same "Support multiple Ubuntu's" modification :-)

jdekonin:
I guess the addition of the packages `libgstreamer*` would fall under the Updates part of your comment.  Is there something to tie this against issue wise.

For s390, I guess its just the `numactl` package deletion that is of concern, why is that being deleted.

sxa555:
libgstreamer is mentioned in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/530

Good catch on numactl - it's removal was an error possibly while I was resolving a merge conflict - now fixed

sxa555:
It was a merge conflict which I guess no longer exists after something else going in. I was planning to just squash it during the merge in the github UI

sxa555:
I don't think I can easily do it from my git client so would definitely prefer to just do it on the UI

PR #757, opened 2019-04-05
Ansible: Add JDK 12 Role to Unix Playbook

HusainYusufali:
* Added JDK 12 install to Unix playbook

* Add bootjdk variable to adoptopenjdk_variables.yml

* Add ablility for main.yml to read default variables

fyi @jdekonin


sxa555:
Do we need the sections for each architecture now? It looks like all the architecture names used in the downloads match the ansible names, so we should be good to squash them together, unless my quick look over them is incorrect (I see aarch64 is currently locked to HotSpot but that's all)

HusainYusufali:
Linux needs the heap size specified which makes it different. That means s390x and ppc64le can be squashed, but would need to specify that linux and aarch64 should be excluded. I kept them separate because it only adds one task and keeps each architecture separate.

PR #759, opened 2019-04-08
Update OpenSSL to v1.1.1b

vsebe:
Add to PATH via profile for remote desktop connections and via ssh client environment for Jenkins agent.
Link libraries and clean up old versions.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
@@ -1,64 +1,252 @@
 ---
 ###################
-# OpenSSL v1.1.0g #
+# OpenSSL v1.1.1 #
 ###################
 # Required by OpenJ9 for out-of-process JIT compilation (aka JITaaS)
 # Currently only used by the alternate openj9 branch at https://github.com/eclipse/openj9/tree/jitaas
 
-- name: Test if OpenSSL v1.1.0g is installed
-  shell: export LD_LIBRARY_PATH=/usr/local/openssl-1.1.0/lib &&  /usr/local/openssl-1.1.0/bin/openssl version | awk '{print $2}' | cut -c 1-5
+# Note: some systems have already OpenSSL 1.1.1 instaled (as system)
+# do not install 1.1.1b on them
+
+- name: Set OpenSSL facts
+  set_fact:
+    openssl_latest: 1.1.1
+    openssl_latest_release: 1.1.1b
+  tags: openssl-1.1.1
+
+# check if it is installed in custom location or as system
+
+- name: Test if OpenSSL {{ openssl_latest }} or {{ openssl_latest_release }} is installed
+  shell: test -f /usr/local/openssl-{{ openssl_latest_release }}/bin/openssl ; /usr/local/openssl-{{ openssl_latest_release }}/bin/openssl version || /usr/bin/openssl version | grep {{ openssl_latest }}
   when:
     - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
     - ansible_architecture == "x86_64"
   register: openssl_version
-  tags: openssl
+  ignore_errors: True
+  tags: openssl-1.1.1
 
-- name: Download OpenSSL v1.1.0g
+- debug: var=openssl_version.stdout
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
+    - ansible_architecture == "x86_64"
+  tags: openssl-1.1.1
+
+# install latest release
+
+- name: Convert openssl_latest_release variable
+  shell: echo {{ openssl_latest_release }} | sed 's/\./_/g'
+  register: converted_openssl_latest_release
+  when:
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -1,64 +1,184 @@
...
+    - openssl_version.stdout == ''
+  ignore_errors: True
+  tags: openssl
+
+- name: Create /etc/ld.so.conf.d/openssl-{{ openssl_latest_release }}.conf
+  copy:
+    dest: /etc/ld.so.conf.d/openssl-{{ openssl_latest_release }}.conf
+    content: /usr/local/openssl-{{ openssl_latest_release }}/lib
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
```
jdekonin: Can you specify which os / release variants you tested this against?

vsebe: centos 6, rhel6, ubuntu14, ubuntu16, sles12

sxa555:
Hi, what is the driver for this change (i.e. is there a related issue) and are there any other related PRs to handle the fact this directory name will now be different from what it was before?

vsebe:
@jdekonin ready for review, thank you!

sxa555:
@vsebe See my question from yesterday :-)

vsebe:
@sxa555 - the directory name change has no impact on OpenJ9 build and tests.
This PR also sets openssl-1.1.1 as default for the Jenkins user (see the tasks that add openssl to the PATH) on Linux x86 platforms.

sxa555:
I'm really not comfortable with changing the default like that - why is that needed instead of just doing it in whatever jobs require it? That has the potential to lead to confusion.

vsebe:
@sxa555 - to add it to the jobs would have update either multiple Jenkins files across multiple repositories or update the slaves configuration.
Setting it as default could be made optional.

sxa555:
@vsebe Talking to @mpirvu in slack he says that the JITaaS code will build against 1.0.2 which is available on most machines anyway, so I'm even less clear about why we need 1.1.1 at all just now. Can you clarify please?

jdekonin:
JITaaS code will build against v1.0.2 or v1.1.1 long term releases, but since v1.0.2 is end of life at the end of 2019 as per [openssl.org](https://www.openssl.org/policies/releasestrat.html), it is likely that OpenJ9 will be using v1.1.1 as the requirement.  @pshipton or @mpirvu, please state otherwise if that isn't the case.

Could we not have this role included but maybe not added to main.yml until such time that it is something that AdoptOpenJDK farm machines would require?  I would suggest that this role get a different tag though, say `openssl-1.1.1`?

sxa555:
> Could we not have this role included but maybe not added to main.yml 

Yeah I would be ok with that :-)

Ref the support statement, that doesn't prevent the distributions from continuing to support it beyond that. I haven't looked to see if RedHat have made a statement on it yet but I doubt RHEL7 will be upgraded beyond 1.0.2 even if official support at openssl does not continue, so it will still be the default supported version on many systems, including our build ones.

mpirvu:
As stated in a different item my preference would be to build JITaaS code against openssl 1.1.x because:
(1) That's what we filed a CQ for
(2) We have done much more testing with 1.1.x than with 1.0.x
(3) At some point 1.0.x will become obsolete.

vsebe:
@jdekonin @sxa555 ready for your review, thank you!

sxa555:
@mpirvu Just for clarification when you refer to "CQ" is that the [Eclipse Contribution Questionnaire](https://wiki.eclipse.org/Development_Resources/Contribution_Questionnaire)?

mpirvu:
> @mpirvu Just for clarification when you refer to "CQ" is that the Eclipse Contribution Questionnaire?

Yes.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -1,64 +1,252 @@
...
+    owner: "{{ Jenkins_Username }}"
+    mode: 0644
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
+    - ansible_architecture == "x86_64"
+    - openssl_version.stdout == ''
+  tags: openssl-1.1.1
+
+- name: Add OpenSSL-1.1.1b to Jenkins user's environement
+  shell: echo PATH=/usr/local/openssl-1.1.1b/bin:$PATH > /home/{{ Jenkins_Username }}/.ssh/environment
```
AdamBrousseau: Should `openssl-1.1.1b` be hardcoded here?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -1,64 +1,252 @@
...
+    - openssl_version.stdout == ''
+  tags:
+    - openssl
+    - skip_ansible_lint
+
+- name: Enable user environment in the OpenSSH daemon configuration file
+  replace:
+    path: /etc/ssh/sshd_config
+    regexp: '^PermitUserEnvironment no'
+    replace: 'PermitUserEnvironment yes'
```
AdamBrousseau: If the file doesn't have this line, will it be added?

PR #760, opened 2019-04-08
Add OpenSSL 1.0.2 prereq on CentOS/RHEL6 for OpenJ9

sxa555:
This can be downloaded live, but I'd rather have it cached. CentOS6 (used on x64 for building) has openssl 1.0.1 which cannot be used by OpenJ9.
Separate PR in to the build repository to start using this https://github.com/AdoptOpenJDK/openjdk-build/pull/1023

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml
@@ -1,0 +1,58 @@
+---
+###################################################
+# OpenSSL v1.0.2r for building OpenJ9 on CentOS 6 #
+###################################################
+
+- name: Test if OpenSSL v1.0.2r is installed
+  shell: export LD_LIBRARY_PATH=/usr/local/openssl-1.0.2/lib &&  /usr/local/openssl-1.0.2/bin/openssl version | awk '{print $2}' | cut -c 1-5
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
+    - ansible_distribution_major_version == "6"
+  register: openssl_version
+  tags: openssl
+
+- name: Download OpenSSL v1.0.2r
+  get_url:
+    url: https://www.openssl.org/source/openssl-1.0.2r.tar.gz
+    dest: /tmp/openssl-1.0.2r.tar.gz
+    force: no
+    mode: 0755
+    validate_certs: no
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
+    - ansible_distribution_major_version == "6"
+    - ((openssl_version.stdout == '') or ((openssl_version.stdout != '') and (openssl_version.stdout | version_compare('1.0.2', operator='lt', strict=True))))
+  tags: openssl
+
+- name: Extract OpenSSL v1.0.2r
+  unarchive:
+    src: /tmp/openssl-1.0.2r.tar.gz
+    dest: /tmp
+    copy: False
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
+    - ansible_distribution_major_version == "6"
+    - ((openssl_version.stdout == '') or ((openssl_version.stdout != '') and (openssl_version.stdout | version_compare('1.0.2', operator='lt', strict=True))))
+  tags: openssl
+
+- name: Build and install OpenSSL v1.0.2r
+  shell: cd /tmp/openssl-1.0.2r && ./config --prefix=/usr/local/openssl-1.0.2 shared && make && make install
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
+    - ansible_distribution_major_version == "6"
+    - ((openssl_version.stdout == '') or ((openssl_version.stdout != '') and (openssl_version.stdout | version_compare('1.0.2', operator='lt', strict=True))))
+  tags: openssl
+
+- name: Remove downloaded packages for OpenSSL v1.0.2r
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /tmp/openssl-1.0.2r.tar.gz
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+###################################################
+
+- name: Test if OpenSSL v1.0.2r is installed
+  shell: export LD_LIBRARY_PATH=/usr/local/openssl-1.0.2/lib &&  /usr/local/openssl-1.0.2/bin/openssl version | awk '{print $2}' | cut -c 1-5
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
+    - ansible_distribution_major_version == "6" 
+  register: openssl_version
+  tags: openssl
+
```
jdekonin: Tags should be modified such that either/or openssl role can be targeted/omitted.

sxa555:
@vsebe's change is for JITaaS which I believe needs 1.1.1. This is to allow RHEL/CentOS6 to build against the same level main level (1.0.2) as the other CentOS7.4 systems which have that version preinstalled. I don't want to mess with the system version.

Can you clarify what your issue is with the tags? Your comment is against something that already has an `openssl` tag on it.

jdekonin:
The current openssl role uses `tags: openssl`, I just think they should be separated and not grouped into one as they are different roles for different purposes.

vsebe:
@sxa555  - https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/759 installs openssl 1.1.1 on all Linux x86 platforms. Openssl 1.0.2 is not required on centos 6 & rhel 6 x86.

sxa555:
@vsebe My preference was to build the openJ9 JDKs against 1.0.2 since that's the earliest release and the one that's consistent with the CentOS7.4 build platforms therefore safer for compatibility. If we're 100% sure that building against 1.1.1 source (I believe that's a JITaaS requirement so not actually needed for any adoptopenjdk builds) will result in something that will execute against 1.0.2 that I'd be ok with not doing this, but I'd need that confirmation.

sxa555:
@jdekonin Is the openssl111 purely for jitaas? In which case a jitaas tag for 1.1.1 might be more appropriate as per my previous comment to Violete

AdamBrousseau:
afaik 1.1.* is required for JITaaS which is currently only supported on xlinux. I assume Adopt would need this (on at least the compile machines) in order to ship binaries with JITaaS support. Even if we aren't testing JITaaS at Adopt.

sxa555:
@AdamBrousseau I thought the JITaaS was still in a separate branch so we wouldn't even be building it at present, or am I out of date? :-)

karianna:
@sxa555 Will need rebasing

AdamBrousseau:
True, I was confusing it with a separate project. JITaaS is still not in the mainline.

sxa555:
@AdamBrousseau In which case we should probably get @vsebe to modify her 1.1.1b upgrade PR (if we decide to land it) to use `jitaas` instead of `openssl` although I'm still not convinced we should even have it in the playbooks if it's not required for anything in the mainline code.

sxa555:
@jdekonin Given the discussions and modifications in 1.1.1 can you update your review please?

jdekonin:
The merge needs review; there are 3 commits in this PR, 2 of which don't belong.  Travis failure is highlighting trailing white space that should also be fixed.

sxa555:
The merge commit shouldn't be an issue as it'll get squashed on merge. Have resolved whitespace

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+- name: Build and install OpenSSL v1.0.2r
+  shell: cd /tmp/openssl-1.0.2r && ./config --prefix=/usr/local/openssl-1.0.2 shared && make && make install
+  when:
+    - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS")
+    - ansible_distribution_major_version == "6"
+    - ((openssl_version.stdout == '') or ((openssl_version.stdout != '') and (openssl_version.stdout | version_compare('1.0.2', operator='lt', strict=True))))
+  tags: openssl
+
+- name: Remove downloaded packages for OpenSSL v1.1.0g
```
karianna: Is that a correct name?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml~
```diff
@@ -0,0 +1,58 @@
+---
```
karianna: `main.yml~`?   Seems odd

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml
sxa555: In general I'd rather use the build-system provided OpenSSL at whatever level that is. At the moment RHEL6 doesn't have any suitable one, so I'm only installing this explicitly on RHEL/CentOS6 to make a level of 1.0.2 available to build against (the exacty level isn't actually important and it'll never be the system SSL).

sxa555: No :-) Well spotted!

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL102/tasks/main.yml~
sxa555: It's a backup file that shouldn't have been in my add - removing ...

PR #764, opened 2019-04-10
Add test-azure-win2012r2-x64-2 and 3 to inventory

sxa555:

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -74,6 +74,8 @@ hosts:
 
       - azure:
           win2012r2-x64-1: {ip: 13.68.134.204, user: adoptopenjdk}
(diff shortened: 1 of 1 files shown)


ansible/inventory.yml
```diff
@@ -74,6 +74,8 @@ hosts:
 
       - azure:
           win2012r2-x64-1: {ip: 13.68.134.204, user: adoptopenjdk}
+          win2012r2-x64-2: {ip: 40.121.164.56, user: adoptopenjdk}
```
karianna: Is that IP correct?

sxa555: It's definitely the one @tellison gave me for that machine and does work - so it's connecting to some windows box somewhere in the cloud ;-)

PR #766, opened 2019-04-10
Ansible: Remove freetype from windows playbook

ali-ince:
Having updated our build scripts (with [#1015](https://github.com/AdoptOpenJDK/openjdk-build/pull/1015) and [#1025](https://github.com/AdoptOpenJDK/openjdk-build/pull/1025)) to download freetype on demand and to build during the configure phase, we no longer need freetype as an explicit dependency on windows.

This PR removes it from the playbook. Supersedes #747.


jdekonin:
By removing this role OpenJ9 will not be able to use the windows playbooks as is for our machines setups.  Nor will any individual be able use the playbook or guide to setup a windows system.  Could the main.yml not have been modified to not call the role?

karianna:
@jdekonin Apologies we weren't aware of how these were used in downstream projects.   Hmm, looks like that might need to a review step going forward.

karianna:
@ali-ince Can you create a reverse PR for this?  We'll then think about how to do this with less negative impact.

PR #768, opened 2019-04-11
Revert "Remove freetype from windows playbook (#766)"

ali-ince:
This reverts commit 0628f935dccbc37a2ffdde135ec5222ce43b1828.

@jdekonin @karianna this PR reverts #766. 

Is it ok if we remove `freetype` from `main.yml` but leave the role definition?


ali-ince:
@jdekonin I've pushed another commit that removes  `Freetype` from `roles` in `main.yml`.

karianna:
Discussed at the TSC - philosophy is that Adopt should prioritize its own build farm *but* should always make its infrastructure as code open to extension.  This seems like a good compromise, thanks, everyone.

PR #776, opened 2019-04-26
Ansible: Install Python 3.7.3 to Unix Playbook

HusainYusufali:
* Install Python 3.7.3 to Unix Playbook
* Installs from source to Redhat, CentOS, Ubuntu, and SLES

* The following issues require python3 to satisfy the JitBuilder Python dependency:
  * eclipse/omr#2397
  * eclipse/omr#2915
  * eclipse/omr#2968

fyi @jdekonin @vsebe 

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Python3_Source/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Python3_Source/tasks/main.yml
@@ -1,0 +1,83 @@
+---
+##############
+# Python3_Source #
+##############
+- name: Test if python3 is installed on path
+  shell: python3 --version >/dev/null
+  ignore_errors: yes
+  register: python3_installed
+  tags:
+    - python3_source
+
+- name: Ensure zlib1g and libffi-dev is installed on Ubuntu before compiling python3
+  apt:
+    name: "{{ packages }}"
+  vars:
+    packages:
+      - zlib1g
+      - zlib1g-dev
+      - libffi-dev
+  when:
+    - (ansible_distribution == "Ubuntu")
+    - (python3_installed.rc != 0 )
+  tags: python3_source
+
+- name: Ensure zlib-devel and libffi-devel-gcc5 is installed on SLES before compiling python3
+  zypper:
+    name: "{{ packages }}"
+  vars:
+    packages:
+      - zlib-devel
+      - libffi-devel-gcc5
+  when:
+    - (ansible_distribution == "SLES")
+    - (python3_installed.rc != 0 )
+  tags: python3_source
+
+- name: Ensure zlib-devel and libffi-devel is installed on CentOS or RedHat before compiling python3
+  yum:
+    name: "{{ packages }}"
+  vars:
+    packages:
+      - zlib-devel
+      - libffi-devel
+  when:
+    - (ansible_distribution == "CentOS" or ansible_distribution == "RedHat")
+    - (python3_installed.rc != 0 )
+  tags: python3_source
+
+- name: Download Python3 source
+  get_url:
+    url: https://www.python.org/ftp/python/3.7.3/Python-3.7.3.tar.xz
+    dest: /tmp/Python-3.7.3.tar.xz
+    mode: 0440
+  when:
+    - (python3_installed.rc != 0 )
+  tags: python3_source
+
+- name: Extract Python3 source
+  unarchive:
+    src: /tmp/Python-3.7.3.tar.xz
+    dest: /tmp/
+    copy: False
+  when:
+    - (python3_installed.rc != 0 )
+  tags: python3_source
+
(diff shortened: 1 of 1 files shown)


karianna:
Why is this being added?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Python3_Source/tasks/main.yml
```diff
@@ -0,0 +1,95 @@
...
+
+- name: Ensure zlib-devel and libffi-devel is installed on CentOS before compiling python3
+  yum: 
+    name: "{{ packages }}"
+  vars:
+    packages:
+    - zlib-devel
+    - libffi-devel
+  when:
+    - (ansible_distribution == "CentOS")
```
vsebe: This task could run on CentOS and RHEL platforms. This way you can dispose of the task on RHEL: Ensure zlib-devel and libffi-devel is installed on RedHat before compiling python3 L49-59

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -52,6 +52,7 @@
         - (ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "Ubuntu" or ansible_distribution == "SLES")
         - ansible_architecture == "x86_64"
     - OpenSSL102                  # OpenJ9
+    - Python3_Source
```
vsebe: You might remove Python3_Source role as this is a requirement for an OpenJ9 future test.

HusainYusufali: The role has been removed

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Python3_Source/tasks/main.yml
HusainYusufali: RHEL and CENT have been squashed into one task

sxa555:
@HusainYusufali As per @karianna's comment, can you provide a link to an issue that describes why this is needed. Generally unless it's an obvious failure we should have related issues describing the requirement for PRs

HusainYusufali:
These are the issues that require Python to be installed on machines.

* https://github.com/eclipse/omr/issues/2397
* https://github.com/eclipse/omr/pull/2915
* https://github.com/eclipse/omr/pull/2968

karianna:
> These are the issues that require Python to be installed on machines.
> 
> JitBuilder Python dependency:
> 
> * [eclipse/omr#2397](https://github.com/eclipse/omr/issues/2397)
> * [eclipse/omr#2915](https://github.com/eclipse/omr/pull/2915)
> * [eclipse/omr#2968](https://github.com/eclipse/omr/pull/2968)

Thanks - can you state that in the comment for the package declaration?

sxa555:
Hmm that does say that some patches were going in to make the tests suites work with Python 2 though. If it all possible I'd prefer to just stick with the default on the machines instead of introducing another version (And I'm not sure we'd want Python3 as a default). @Leonardo2718 I can't tell who you really are from your github profile but were the tests modified to tolerate 2.7 as suggested in https://github.com/eclipse/omr/pull/2915#issuecomment-428201958?

In principle I'm ok with this change regardless (Although explicitly having prereq pacakges included in the roles is a bit unusual!) since it's not being included by default from the top level playbook (I assume that's intentional!)

HusainYusufali:
Python 2.6 has already been retired and Python 2.7 will be EOL on Jan 1st, 2020. This is why Python 3 is being installed. The prereq packages are to make sure Python 3.7.3 can be installed from source.

sxa555:
Python2 is still the default with RHEL at the moment so I suspect it'll still be supported by them on most of our build systems at least. The justification in those issues had nothing to do with EOL though from what I could see :-)

My personal preference is very much to use the default versions on each distro unless we have a hard requirement on an alternative version in order to keep the machines as "clean" as possible to avoid potential false positive test results ...

Leonardo2718:
@sxa555 
> were the tests modified to tolerate 2.7 as suggested in eclipse/omr#2915 (comment)?

Yes, the tests will work with 2.7. A few small pieces get automatically disabled to avoid failures due to missing functionality, hence the Python 3 preference, but most test will just work.

sxa555:
Indentation failures in travis with the new role causing lint failures

jdekonin:
fyi @0xdaryl

karianna:
Will Python 2.7.x still be available on these hosts? I believe Graal will require it.

jdekonin:
This change just adds 3.7 into /usr/local and leaves whatever is installed on the system alone.  This change also isn't included in the main.yml as a role to execute as it wasn't deemed a requirement for anything other than OpenJ9s JitBuilder.

PR #792, opened 2019-05-28
Remove CUDA toolkit as OpenJ9 does not yet support it

sxa555:

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -30,7 +30,6 @@
     - cmake                       # OpenJ9 / OpenJFX
     - gmake
     - Docker                      # Testing
-    - NVidia_Cuda_Toolkit         # OpenJ9
     - Superuser                   # AdoptOpenJDK Infrastructure
     - Swap_File
     - Crontab


keithc-ca:
It isn't that OpenJ9 doesn't _support_ CUDA, but rather, (I understand) that AdoptOpenJDK has not accepted the EULA to _enable_ CUDA-based features in OpenJ9. What are the objections to accepting the terms of that EULA?

karianna:
> It isn't that OpenJ9 doesn't _support_ CUDA, but rather, (I understand) that AdoptOpenJDK has not accepted the EULA to _enable_ CUDA-based features in OpenJ9. What are the objections to accepting the terms of that EULA?

Someone point me to the EULA and I can do an (IANAL) review.

keithc-ca:
The latest EULA is at https://docs.nvidia.com/cuda/eula/index.html.

karianna:
The Limitations Section contains:

```
You may not use the SDK in any manner that would cause it to become subject to an open source software license. As examples, licenses that require as a condition of use, modification, and/or distribution that the SDK be:
Disclosed or distributed in source code form;
Licensed for the purpose of making derivative works; or
Redistributable at no charge.
```

Gives me enough pause that I'd recommend we get formal legal advice.

Is the OpenJ9 looking at this via the Eclipse Foundation or is this an AdoptOpenJDK only build step/concern?

keithc-ca:
I don't think any of those clauses apply to the way OpenJ9 would use CUDA (we're only including a few header files, we don't link against or redistribute any of their libraries), but I agree it would be good to have legal advice.

Anyone can use the existing open-source projects and download the CUDA headers themselves to build a JVM with CUDA support, so I don't think this implies any new issues for the Eclipse Foundation: I'd say it's just an AdoptOpenJDK concern (but then, IANAL either).

mstoodle:
I went through a similar kind of thing with the Eclipse Foundation a while back.

@keithc-ca wll correct me if I'm wrong, but as I understand it, the CUDA native library support built into the OpenJ9 binaries is merely the ability to load the native CUDA library and program against the external interface of that library. Building Eclipse OpenJ9 with CUDA compiles a lot of extra code in OpenJ9 but nothing from the actual library is built into an OpenJ9 binary. The user of those binaries then needs to install the CUDA library (and accept the EULA, since they're the real "EU"s here). After discussion with the Eclipse Foundation's IP team, our project chose to accept the EULA for our CI machines only for use in our pipelines so that we would be able to build CUDA support and run CUDA tests (which obviously require the library to be installed). I believe that's very similar to what other projects do when they have a dependency on this library.

My (non legal) suggestion is that AdoptOpenJDK should consider a similar arrangement: accept the EULA in order to install the CUDA libraries onto build machines (and test machines to enable running CUDA tests of this capability). But (IMO) AdoptOpenJDK binaries should not bundle the CUDA library; rather, the user should be expected to provide that library after accepting the license terms.
Hope that helps!

mstoodle:
Heh, Keith posted just before I did! .... I interpret that we agree :) .

keithc-ca:
The only point I might disagree with is your phrase 'a _lot_ of extra code': I believe it's exactly two source files (`omrcuda.cpp` in OMR and `GPUHelpers.cpp` in OpenJ9) - I would have said '_some_ extra code'. ;-)

keithc-ca:
... and if I had some free time, I would finish what I started and update `GPUHelpers.cpp` to use the OMR API which would leave us with exactly one source file that needs the CUDA headers.

sxa555:
IMHO it should only be an issue if something in the license impacts the inclusion of the contents of the header files within the product

sxa555:
To tie this together, this reverts https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/362

karianna:
Have added a TSC issue for transparency - I suspect we will follow Eclipse's lead.  Their process for checking this sort of thing is robust.

PR #806, opened 2019-06-18
Fixes to permit Redhat6 on x86 to apply cleanly

jdekonin:
* fix conditional such that openjdk7/8 are installed
* correct alternative for x86 as its openjdk and not the ibm
    package that installed
* #807

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
@@ -84,7 +84,7 @@
 - name: Install Java (Not RedHat 6 on ppc64)
   package: "name={{ item }} state=latest"
   with_items: "{{ Java_NOT_RHEL6_PPC64 }}"
-  when: (ansible_distribution_major_version != "6" and ansible_architecture != "ppc64")
+  when: not (ansible_distribution_major_version == "6" and ansible_architecture == "ppc64")
 
 - name: Install Java when RedHat 6 on ppc64
   package: "name={{ item }} state=latest"
@@ -95,7 +95,7 @@
 # Set default Java #
 ####################
 - name: Set default java version - RHEL 6 x86_64
-  shell: update-alternatives --set java  /usr/lib/jvm/jre-1.8.0-ibm.x86_64/bin/java
+  shell: update-alternatives --set java  /usr/lib/jvm/jre-1.8.0-openjdk.x86_64/bin/java
   when:
     - ansible_distribution_major_version == "6"
     - ansible_architecture == "x86_64"


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -84,7 +84,7 @@
 - name: Install Java (Not RedHat 6 on ppc64)
   package: "name={{ item }} state=latest"
   with_items: "{{ Java_NOT_RHEL6_PPC64 }}"
-  when: (ansible_distribution_major_version != "6" and ansible_architecture != "ppc64")
```
sxa555: I'm not sure if that's going to produce the desired effect - as it'll knock out the installation on any RHEL7. Did you mean to switch the "and" to an "or" instead?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -136,6 +136,39 @@
     - ansible_architecture == "s390x"
   tags: default_java
 
+##############################
```
sxa555: Why is this needed? We should already have the expat from the RHEL/CentOS repos installed.
Also if it does need to be done like this we should do it in a separate role.

jdekonin: It seems expat is available if your starting rh6 release is new enough. My original base was 6.3, where this error occurred. With 6.10, expat.x86_64 v2.0.1-13.el6_8 is available. I'm fine with removing the expat changes if there is an agreement that the playbooks will only run against the latest of a specific release.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
jdekonin: I haven't tested this, but do you know if this will work.
`when: not (ansible_distribution_major_version == "6" and ansible_architecture == "ppc64")`

jdekonin: Okay tested now, this syntax does work.

sxa555: Logically that's identical to my suggestion, so I'm ok with that change :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa555: We have some RHEL6.9 machines, but I wouldn't expect it generally to be running against anything as old as 6.3 - thanks for the update

PR #811, opened 2019-06-26
Add cmake role for s390x

jdekonin:


jdekonin:
This is used by OpenJ9 on s390x as well.  Comment is already present in the main.yml detailing OpenJ9 and OpenJFX as users.

PR #813, opened 2019-07-02
Move Linux/s390x build machine to test farm

sxa555:


sxa555:
~cut~ copy and paste error - fixed @karianna

PR #823, opened 2019-07-15
remove cloudcone machines

gdams:
These are no longer online


sxa555:
Missing `godaddy` entry didn't get picked up by the checks when https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/817 was merged. Thanks  for fixing George

PR #828, opened 2019-07-17
Ansible-doc-rework

Willsparker:
Added some write-up on how to run two VMs on different OSs on the same machine, as I had trouble with this before.

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,5 +1,22 @@
 # Ansible playbooks to download and install dependencies for OpenJDK on various platforms
 
+# Quickstart Guide
+
+For macOS:
+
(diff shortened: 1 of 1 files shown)


ansible/README.md
```diff
@@ -1,5 +1,24 @@
 # Ansible playbooks to download and install dependencies for OpenJDK on various platforms
 
+# Quickstart Guide
+
+To test the ansible scripts, you'll need to install the following programs.
+
+For macOS:
+
+1) Install Homebrew 2.1.7 or later
```
gdams: nit: can you change your numbers to be `1.` rather than `1)` please? It's the standard for markdown.

PR #831, opened 2019-07-18
Add IcedTea-Web onto Windows Playbook

Willsparker:
Implemented and tested it :)

Fixes: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/829

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,5 +1,22 @@
 # Ansible playbooks to download and install dependencies for OpenJDK on various platforms
 
+# Quickstart Guide
+
+For macOS:
+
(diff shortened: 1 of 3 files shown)


gdams:
I'm not 100% sure that this is the best way to do this. @sxa555 is there a reason that we don't just install the Java 8 MSI on windows machines and add ITW as a feature that way?

sxa555:
It would be an option but would require changing the way we install all of the other versions too for bootstrapping if we wanted a level of consistence. This is simpler IMO, and the since ITW can be used across versions if required.

PR #833, opened 2019-07-18
use ansible_distribution rather than os

gdams:


sxa555:
LGTM - thanks for the quite fix

PR #856, opened 2019-07-29
Install expat on Linux s390x

vsebe:
Modify conditional to allow expat installation on Linux s390x.
Add task for Linux s390x to compile and install expat from source.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/SLES.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/SLES.yml
@@ -142,10 +142,18 @@
     - java8_installed.stat.exists == false
   tags: java8_SLES11
 
-##############################
-# expat on SLES 11 on x86_64 #
-##############################
-- name: Install expat on SLES 11
+#########
+# expat #
+#########
+- name: Test if expat is installed
+  stat:
+    path: /usr/local/lib/libexpat.so
+  register: expat_installed
+  when:
+    - ((ansible_distribution_major_version == "11" and ansible_architecture == "x86_64") or (ansible_distribution_major_version == "12" and ansible_architecture == "s390x"))
+  tags: expat
+
+- name: Download expat
   get_url:
     url: https://github.com/libexpat/libexpat/releases/download/R_2_2_5/expat-2.2.5.tar.bz2
     dest: /tmp/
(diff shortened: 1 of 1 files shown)


vsebe:
@jdekonin please review, thx

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/SLES.yml
```diff
@@ -142,35 +142,52 @@
...
   become: yes
   when:
-    - (ansible_distribution == "SLES" and ansible_distribution_major_version == "11")
+    - ansible_distribution_major_version == "11"
     - ansible_architecture == "x86_64"
+    - expat_installed.stat.exists == false
+  tags: expat
+
+- name: Running ./configure & make for expat on Linux s390x
+  shell: cd /tmp/expat-2.2.5 && ./configure && make -j {{ ansible_processor_cores }} && sudo make install
```
jdekonin: Is a configure option needed to get this to install into `/usr/local/lib`?  This is needed for SLES11, 12 and 15?  Might be better to state explicitly unless you can test/verify them all.

vsebe: No, is not. `/usr/local/lib` is the default location.

sxa555:
I'm not massively keen on having a build from source operation within the common role. Could this be done using the `local_srcinstall` role that's used elsewhere in the top level playbook instead? Also why is this specific to s390x?

sxa555:
A couple of quesions:
- Why do we need a separate build section for s390x when the download/extract sections are common
- Is there a reason why this needs to be specific to certain architectures?
- Could this be better achieved using a `local_srcinstall` at the top level of the playbook. I appreciate that this was already in there, but this could be a cleaner solution and I'm not too keen on having source compiles in the `Common` roles unless really necessary

vsebe:
>  Is there a reason why this needs to be specific to certain architectures?
Why do we need a separate build section for s390x when the download/extract sections are common
Is there a reason why this needs to be specific to certain architectures?

Yes, it is. The ansible variable storing the number of cpus is differently named on zLinux - `ansible_processor_cores` than on xLinux - `ansible_processor_vcpus`.

I added a new task to be consistent, please see [ccache role](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml). 
  
>Could this be better achieved using a local_srcinstall at the top level of the playbook?

`local_srcinstall` would need to support vary platforms and architectures. Currently it doesn't. 
Also, the tags do not work with `local_srcinstall`.

The clean up of the `Common` role could be done in a different PR as would require more development and testing.

Issue #842, opened 2019-07-23
Automating the testing of the playbooks for Linux using Vagrant


Willsparker:

sxa555:
Reopening this until we have a full solution that works via our CI :-)

One issue we currently have is that we need to be able to get VirtualBox running within one of our cloud VMs which means getting VirtualBox configured from the command line not to use hardware virtualisation instructions (Alternative is to get a dedicated bare metal VM that we can run VMs on but that would be more expensive)

Willsparker:
@sxa and I looked into it and it seems not possible to run Virtualbox from another VM as hardware virtualisation can't be disabled. So, a bare metal VM will be needed.

sxa555:
Now running on a Thinkpad at https://ci.adoptopenjdk.net/computer/infra-vagrant-1/ but continuing to investigate sourcing a suitable cloud machine for this - will deal with that separately.

See also separate task #890 for adding Debian to this set of tests.

PR #860, opened 2019-07-30
Added/Modified Scripts to verify OpenJDK builds

Willsparker:
Added into the script that builds the vagrant machines to test the playbooks, the scripts to run and build openJDk on the machines, to ensure the ansible playbook correctly installed everything.

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -eu
+export TARGET_OS=linux
+export ARCHITECTURE=x64
+export JAVA_TO_BUILD=jdk8u
+export VARIANT=openj9
+export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
+export JDK_BOOT_DIR=/usr/lib/jvm/java-8-openjdk-amd64
+cd ~/openjdk-build
+build-farm/make-adopt-build-farm.sh
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -8,3 +8,4 @@ export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
 export JDK_BOOT_DIR=/usr/lib/jvm/java-8-openjdk-amd64
 cd ~/openjdk-build
 build-farm/make-adopt-build-farm.sh
+
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
 set -eu
-
-# Takes all arguments from the script
-processArgs()
+processArgs() #takes all arguments from the script
 {
 	if [ $# -lt 2 ]; then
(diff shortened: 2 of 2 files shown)


Willsparker:
Relates to #842

ansible/pbTestScripts/buildJDK.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
karianna: This seems very hardcoded to one particular variation.  If that's on purpose then that's OK but perhaps rename the file to reflect reality.  Else you can parameterize this script and pass arguments into it...

ansible/pbTestScripts/testScript.sh
```diff
@@ -2,7 +2,7 @@
 set -eu
 
 # Takes all arguments from the script
-processArgs()
+processArgs()
```
karianna: whitespace

ansible/pbTestScripts/testScript.sh
```diff
@@ -31,19 +31,22 @@ setupGit()
...
 	cd ~/adoptopenjdkPBTests/$2/ansible
-
-	# Alias the correct vagrant file
-	ln -sf Vagrantfile.$1 Vagrantfile
-	
-	vagrant up
-
-	# Remotely moves to the correct directory in the VM and builds the playbook. Then logs the VM's output to a file, in a separate directory
+	ln -sf Vagrantfile.$1 Vagrantfile	#Alias the correct vagrant file
+	vagrant  up
```
karianna: whitespace

ansible/pbTestScripts/testScript.sh
```diff
@@ -31,19 +31,22 @@ setupGit()
...
 	vagrant ssh -c "cd /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook && sudo ansible-playbook --skip-tags "adoptopenjdk,jenkins" main.yml" 2>&1 | tee ~/adoptopenjdkPBTests/logFiles/$2.$1.log
 
+		# ^ Remotely moves to the correct directory in the VM and builds the playbook. Then logs the VM's output to a file, in a separate directory
```
karianna: Prefer comments above the line of code that it's executing, also - whitespace.

ansible/pbTestScripts/testScript.sh
```diff
@@ -75,16 +78,16 @@ setupFiles
 setupGit $1 $folderName
 
 # For all tested OSs / Playbooks
-for OS in Ubuntu1804 Ubuntu1604 CentOS6
+for OS in Ubuntu1804 Ubuntu1604 CentOS6
```
karianna: whitespace.

ansible/pbTestScripts/buildJDK.sh
sxa555: Agree, although I've got so many other enhancements I would like to get on that I'd quite like to leave this as-is for now then adjust later if you're OK with that Martijn. I may even go for a random selection later. This is basically a simp!eat form of what's needed to run our builds using our environment scripts from openjdk-build. We're hoping to do something similar for running a basic operation fr. The test suites too

karianna: OK :-)

PR #863, opened 2019-07-30
Set limits for Jenkins_User

cwesMills:
This change sets specified limits on the Jenkins User, such as max files
open (nofile), max user processes (nproc), and max core file size (core).

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
@@ -51,3 +51,19 @@
     # TODO: write a condition when NOT to run this step
     - skip_ansible_lint
   when: (ansible_distribution != "MacOSX")
+
+- name: Ensure proper limits are set in /etc/security/limits.conf
+  lineinfile:
+    path: /etc/security/limits.conf
+    line: "{{ Jenkins_Username }} {{ item.limit_type }} {{ item.limit_name }} {{ item.limit_value }}"
+    state: present
+  with_items:
+    - {limit_type: 'hard', limit_name: 'nofile', limit_value: '1048576'}
+    - {limit_type: 'hard', limit_name: 'nproc', limit_value: 'unlimited'}
+    - {limit_type: 'hard', limit_name: 'core', limit_value: 'unlimited'}
+  when:
+    - ansible_architecture == "x86_64"
+    - ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
+    - ansible_distribution_major_version == "6"
+  tags:
+    - jenkins_user


cwesMills:
This change is required for RHEL6 x86-64 machines to pass the tests at
https://ci.adoptopenjdk.net/view/Test_system/job/Test_openjdk8_j9_extended.system_x86-64_linux/

cwesMills:
The travis job was failing because the linter doesn't like the spaces in between the variables of each item. I've added a skip_ansible_linter tag to the task because ansible documentation uses the same format:
```
with_items:
    - { name: 'testuser1', groups: 'wheel' }
    - { name: 'testuser2', groups: 'root' }
```
https://docs.ansible.com/ansible/2.4/playbooks_loops.html

sxa555:
@cwesMills Can you add in the description of the problem (e.g. error message) into this issue, or link it to an issue that has that information please?

sxa555:
I'd like to understand this a little more before approving for two reasons:
1) The default limits on CentOS7 are the same (1024 files, 4096 procs), so why is this change only being requested for 6?
2) I'm a little  nervous about changing system defaults like this unless strictly necessary. Could we just modify the hard limit on the global system and have the scripts that require it set the soft limit as required (ideally not to unlimited) so that we understand when this is required? Otherwise we could end up masking issues we would otherwise see on people's default systems.

cwesMills:
Without ulimit change:
```
11:09:47  SCC stderr JVMSHRC806I Compressed references persistent shared cache "SCC_SysTest_Workload_Cache" has been destroyed. Use option -Xnocompressedrefs if you want to destroy a non-compressed references cache.
11:09:47  STF 08:09:46.722 - Monitoring Report Summary:
11:09:47  STF 08:09:46.722 -   o Process SCC has crashed unexpectedly
11:09:47  STF 08:09:46.722 - Killing processes: SCC
11:09:47  STF 08:09:46.722 -   o Process SCC is not running
11:09:47  **FAILED** at step 8 (Destroy cache). Expected return value=0 Actual=1 at /home/jenkins/workspace/Test_openjdk8_j9_extended.system_x86-64_linux_Nightly/openjdk-tests/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_15646694736693/SharedClasses.SCM01.MultiThread_0/20190801-080937-SharedClasses/tearDown.pl line 67.
```
```
11:38:23  FAILED test targets:
11:38:23  	SharedClasses.SCM01.MultiThread_0
11:38:23  	SharedClasses.SCM01.MultiThreadMultiCL_0
11:38:23  	SharedClasses.SCM23.MultiThread_0
11:38:23  	SharedClasses.SCM23.MultiThreadMultiCL_0
```
With ulimit change:
```
06:36:56  STF 03:36:52.047 - Monitoring processes: SCC
06:36:56  SCC stderr JVMSHRC023E Cache does not exist
06:36:56  STF 03:36:52.230 - Monitoring Report Summary:
06:36:56  STF 03:36:52.230 -   o Process SCC ended with the expected exit code (1)
06:36:56  STF 03:36:52.231 - TEARDOWN stage completed
```
The change is only being requested on RHEL6 x86 because it is the only architecture/OS combination that we have seen this error occur on.

sxa555:
@cwesMills Ref my second point, would modifying only the hard limit in the infrastructure scripts be feasible?

sxa555:
Also, are you sure it doesn't also occur on CentOS6 which which should be the same as RHEL6

gdams:
> Also, are you sure it doesn't also occur on CentOS6 which should be the same as RHEL6

I'd be very surprised if this is exclusive to rhel6 and not centos6.

cwesMills:
It turns out the error does also occur on cent6 so I've added that to the conditional.

As for the limits, the hard limits specify the absolute maximum limits for a user. Whereas the soft limits will default to a much smaller value which causes the test to fail. Therefore the soft limits must be set in the `/etc/security/limits.conf` file, or they must be set from within the test job.

sxa555:
@cwesMills My point exactly - if the tests need a specific value, perhaps the tests should set them and fail if they can't increase it. That approach would lead to less confusion from any third parties running our test suite without having used our playbooks.

jdekonin:
That will only work if the hard limit is already at a level the test can (re)set up too.  But I agree I think there should be documented requirements for what the requirements are wrt system ulimits/kern.sysv* settings.  I just don't believe that doc exists.

sxa555:
Agree Joe - I've got no problem at all with the hard limit going up :-)

sxa555:
Am ok with this and will approve/merge once @Willsparker runs a test on this change ;-)

Willsparker:
All good on the CentOS6 playbook

AdamBrousseau:
What is left outstanding on this?

karianna:
@sxa555 can this be merged now?

PR #867, opened 2019-07-31
add p7zip to test tools

gdams:
required for macOS codesigning

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/MacOSX.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/MacOSX.yml
@@ -19,6 +19,7 @@ Test_Tool_Packages:
   - mercurial
   - pulseaudio
   - cpanminus
+  - p7zip
 
 Test_Tool_Casks:
   - xquartz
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/MacOSX.yml
```diff
@@ -19,6 +19,7 @@ Test_Tool_Packages:
   - mercurial
   - pulseaudio
   - cpanminus
+  - p7zip
```
karianna: Reason?  Add a comment please

gdams:
@karianna can this be merged now?

sxa555:
@gdams You haven't added the requested comment as far as I can tell

gdams:
done

sxa555:
Out of interest, why did we change to use `7z` instead of regular `zip`/`unzip`?
Ref: https://github.com/AdoptOpenJDK/openjdk-build/commit/ec5401f71e290f6b8d788daeaf2506cfda7de95c

gdams:
@sxa555 the built in macOS unzip cannot handle things zipped with PKZIP. See https://unix.stackexchange.com/questions/115825/extra-bytes-error-when-unzipping-a-file

sxa555:
Interesting - are the files initially being created with `pkzip` then? Haven't seen that used in years. Maybe add a comment `Used by signing because unzip can't handle pkzip file-created files`

PR #871, opened 2019-08-01
Add a bootstrap task to ensure python is installed

gdams:
- Adds a bootstrap task to install aptitude/python if they don't exist on the machine.
- Fixes an issue where ant wasn't getting added to machines.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/bootstrap/tasks/partials/ubuntu1604.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/bootstrap/tasks/partials/ubuntu1604.yml
@@ -1,0 +1,19 @@
+---
+
+#
+#  ubuntu 16.04
+#
+
+- name: check for python
+  raw: stat /usr/bin/python
+  failed_when: has_python.rc > 1
+  register: has_python
+
+- name: check for aptitude
+  raw: stat /usr/bin/aptitude
+  failed_when: has_aptitude.rc > 1
+  register: has_aptitude
+
+- name: install python and aptitude
+  when: has_python.rc == 1 or has_aptitude.rc == 1
+  raw: apt-get update && apt install -y python-minimal aptitude
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -3,11 +3,23 @@
 # AdoptOpenJDK - Ansible Playbook #
 ###################################
 - hosts: "{{ groups['Vendor_groups'] | default('localhost:build:test:!*zos*:!*win*:!*aix*') }}"
+  pre_tasks:
+    - name: Check current directoty
+      local_action: stat path="{{ lookup('env', 'PWD') }}/ansible.cfg"
+      register: dir_check
+      tags: debug
+
+    - name: Exit if running from wrong directorys
+      fail: msg="It looks like you're running ansible from the wrong directory, make sure you're in the ansible directory!"
+      when: dir_check.stat.exists == false
+      tags: debug
+
   tasks:
     - block:
       # Set standard variables
         - name: Load AdoptOpenJDKs variable file
           include_vars: group_vars/all/adoptopenjdk_variables.yml
+
   environment:
     PATH: "/usr/local/bin:{{ ansible_env.PATH }}"
 
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -3,7 +3,6 @@
 # AdoptOpenJDK - Ansible Playbook #
 ###################################
 - hosts: "{{ groups['Vendor_groups'] | default('localhost:build:test:!*zos*:!*win*:!*aix*') }}"
-  gather_facts: yes
   tasks:
     - block:
       # Set standard variables
@@ -16,6 +15,7 @@
   # Roles #
   #########
   roles:
+    - bootstrap
     - Debug
     - Version
     - Common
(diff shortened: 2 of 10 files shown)


sxa555:
Can we split this out into separate PRs please since there seems to be a lot of changes bundled into one here and so the commit messages are unclear.

gdams:
@sxa moved the inventory changes to https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/872

sxa555:
I'll need to think about this a bit more before approving as I'm not convinced all this is necessary.

The apache fix and aarch64 conditions are separate from the commit description so should be separated out)
What's the justification for removing `gather facts`? This needs to be documented at least in the PR

gdams:
so gather facts requires python, by moving it to the debug role (setup does the same thing)

```
- name: Fetch Debug Information
  setup:
  tags: debug
```

I can then run a bootstrap role first to check if the ansible prerequisites are installed.

sxa555:
@Willsparker Can you check tomorrow to ensure none of the changes in this PR break any of your test platforms please ;-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/bootstrap/tasks/partials/ubuntu1604.yml
```diff
@@ -0,0 +1,19 @@
...
+  register: has_python
+
+- name: check for aptitude
+  raw: stat /usr/bin/aptitude
+  failed_when: has_aptitude.rc > 1
+  register: has_aptitude
+
+- name: install python and aptitude
+  when: has_python.rc == 1 or has_aptitude.rc == 1
+  raw: apt-get update && apt install -y python-minimal aptitude
```
sxa555: Unless something has changes ansible has generally been able to fall back to `apt` instead of `aptitude` if it's not available which makes this unnecesary

gdams: `aptitude` is considered more efficient for bulk package upgrades

Willsparker:
Screenshot 2019-08-02 at 09 16 11

Doesn't seem to like "stripversion". Running the playbook on an Ubuntu1804 VM

gdams:
@Willsparker `stripversion` is a custom filter written in https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/ed4a605c9522b4a81f9d08695e8af7db842d697a/ansible/plugins/filter/filters.py#L62. I assume that you're running this on internal IBM machines? If you don't wish to use the python inventory library that Adopt does then I'd just suggest skipping the bootstrap task when running internally.

Willsparker:
@gdams Ah! Okay, ran it without and everything seemed to work fine in Ubuntu1804 :)

gdams:
@sxa555 can this be merged now?

jdekonin:
Why are there docker/aarch64 and ant changes mixed in with this pull request?  It would seem those are unrelated issues to this pull request.

sxa555:
@jdekonin I've asked George to split everything not related to the topic into separate PRs

sxa555:
> If you don't wish to use the python inventory library that Adopt does
We are always running with a full clone of the openjdk-infrastructure repo. Have you added something new in here that won't "just work" for anyone cloning and running from the repo @gdams?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/bootstrap/tasks/partials/ubuntu1604.yml
sxa555: Maybe, but I don't think we should necessarily mandate it. Do we know how much of a difference it has? I'm generally not in favour of putting stuff on the machines that isn't either necessary or provides a proven benefit to us.

gdams: the playbooks ran noticeably faster with aptitude installed when I did some tests

gdams:
> Have you added something new in here that won't "just work" for anyone cloning and running from the repo 

@sxa555 the way the playbooks are run in IBM is different because you don't use our custom inventory you don't get a load of variables and custom helpers included. We should not be blocking PR's at Adopt due to issues when running internally on IBM machines. The solution here is to skip running the role.

sxa555:
Not blocking just trying to understand. I don't think Will's issue has anything to do with running on non-Adopt machines which has been working ok until now, and I didn't think anything relating to the inventory hjad changed. Will is doing a straight clone of the repo and running the playbooks with the things I mentioned excluded. If we are introducing a dependency that means it can't run on other people's machines (We make out stuff open to allow people to do that) then we need to understand it.

gdams:
> Something in this PR appears to have stopped it from working. 

As pointed out in an earlier message, our inventory scripts allow us to create custom filters which are referenced in the file below. These filters are only available if you source the filters.py script.

https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/ed4a605c9522b4a81f9d08695e8af7db842d697a/ansible/plugins/filter/filters.py#L62

gdams:
@sxa555 I've removed all other changes now.

sxa555:
Do we need to have all the selinux stuff in there? If someone has it enabled on their systems I'd be somewhat reluctant to just disable it. Under what circumstances would we need the selinux specific python in there?

sxa555:
Based on https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/253/ this doesn't quite appear to work out of the box:

```
TASK [bootstrap : run os-specific bootstrap] ***********************************
fatal: [172.28.128.60]: FAILED! => {"msg": "Unexpected templating type error occurred on ({{ role_path }}/tasks/partials/{{ os|stripversion }}.yml): expected string or buffer"}
```

gdams:
closing for now as I don't have the time to complete this one. If it pops up as an issue again in the future then anybody feel free to take my code :)

PR #873, opened 2019-08-02
improve username logic for running non-sudo tasks on mac

gdams:
this avoids us hard-coding the username on macOS

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
@@ -19,49 +19,49 @@
   register: brew
 
 - name: Install Homebrew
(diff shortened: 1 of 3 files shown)


gdams:
No it’s just that some of the Mac’s have an admin account and others have administrator, using this var allows us to auto detect it

sxa555:
Possibly a silly question but where is the auto-detection. I see
```
        - "ansible_user: {{ ansible_user | default('***Undefined***') }}" 
```
Is it really autodetected somewhere or something we will set as a parameter for each machine?

gdams:
`ansible_user` is a built-in ansible variable :)

sxa555:
So the answer is "It's hard coded in the inventory file and taken from there" right?

gdams:
No it’s got nothing to do with the inventory file, it’s essentially the equivalent of ansible running whoami. On the username that’s used to ssh to the machine before elevating to root

PR #875, opened 2019-08-05
 add aarch64and armv7 docker support

gdams:


sxa555:
I'm wondering if we could just change this conditional to "Not arm32" now (assuming it won't work on arm32 ... If it does we should just remove the architecture conditional altogether!)

gdams:
@sxa555 added the armv7 docker repo too.

sxa555:
(Haven't got round to this yet but still near the top of my list!)

gdams:
@sxa555 I've already confirmed that arm32 works. There is no need for you to run your test

sxa555:
Looks good

PR #876, opened 2019-08-05
install bash v4 or later on macOS

gdams:
needed for openj9


gdams:
openj9 has always needed bash v4 or above. Without it you get this error:

```bash
Bash version 4.0 or later is required!
```

karianna:
interestingly would allow us to upgrade our build scripts to Bash4+ as we held Bash back for Mac OS X reasons.

Issue #666, opened 2019-01-10
UNIX playbook fails if /usr/local/bin does not exist


sxa555:
Objects if /usr/local isn't there, then during the ant role it objects if /usr/local/bin is not in place for the target.

PR #885, opened 2019-08-08
Added directory check for the Ant role

Willsparker:
Fixes the playbook failure that occurs when there is no usr/local/bin directory when trying to create a symlink for ant, by checking and subsequently creating the directory if it's not there.

Fixes: #666


sxa555:
Can you modify this to use the file module please? See https://docs.ansible.com/ansible/latest/modules/file_module.html for examples

Issue #878, opened 2019-08-05
Add Vagrantfile for Debian Linux


sxa555:
We're testing most of the others now, but it would be beneficial to have a Debian variant in the test setup in order to find any Ubuntu/Debian differences given that we have a few Debian systems around now

sxa555:
Update: @Willsparker has hit issues trying to add the ansible 2.4 PPA onto the Debian VM

PR #890, opened 2019-08-12
Added Debian 8 Vagrantfile

Willsparker:
Added the vagrantfile to ansible. Works and installs Ansible 2.8 to the VM.

Fixes: #878

### DIFF
--- a/ansible/Vagrantfile.Debian8
+++ b/ansible/Vagrantfile.Debian8
@@ -5,7 +5,7 @@ $script = <<SCRIPT
 sudo apt-get install tree -y
 sudo apt-get install software-properties-common -y
 sudo apt-add-repository ppa:ansible/ansible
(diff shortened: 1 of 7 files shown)


Willsparker:
Awaiting playbook updates for Debian, @gdams

Willsparker:
Current Issues when running the playbook on a Debian8, by copying and renaming `../Common/tasks/Ubuntu.yml` to `../Common/tasks/Debian8.yml` : 
- Adding both the `openjdk-r` and `ubuntu-toolchain-r` repository to the VM requires the individual source files to be changed to `Trusty` rather than `Jessie`.
- `sudo apt install openjdk-8-jdk` doesn't work due to broken packages - thought this isn't really needed as we could just install `adoptopenjdk8` instead.

Issue #807, opened 2019-06-18
Redhat6 x86 playbooks not 100% functional


jdekonin:
I ran into a few issues creating a machine from scratch yesterday.
- cmake requires c++11 support, which #801 has covered
- fix conditional such that openjdk7/8 are installed
- correct java alternative for x86 as its openjdk and not the ibm package that installed
- compile expat from source, such that cpanm install of XML parser works
- ccache fails to uncompress...seems to be working fine now when doing it manually though

jdekonin:
I think PR #806 covers the issue I raised, minus ccache which is working now so a non-issue.

sxa555:
Can you give the errors that are showing on your install? I don't remember those from the last time I ran the playbooks and it's not entirely obvious what's going wrong from the problem descriptions.. The exception is the IBM java package issue which has also been reported in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/799

jdekonin:
It seems expat is available if your starting rh6 release is new enough.  My original base was 6.3, where this error occurred.  With 6.10, expat.x86_64 v2.0.1-13.el6_8 is available.  I'm fine with removing the expat changes if there is an agreement that the playbooks will only run against the latest of a specific release.

ccache is still not working for me even with 6.10 machine.
```
TASK [../AdoptOpenJDK_Unix_Playbook/roles/ccache : Download ccache.tar.gz] *****
fatal: [aurar1.fyre.ibm.com]: FAILED! => {"changed": false, "dest": "/tmp/ccache.tar.gz", "msg": "Request failed: <urlopen error [Errno 8] _ssl.c:492: EOF occurred in violation of protocol>", "state": "absent", "url": "https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz"}
...
"msg": "Request failed: <urlopen error [Errno 8] _ssl.c:492: EOF occurred in violation of protocol>"
```

sxa555:
@Willsparker I'll assign this to you to see if you've hit the `ccache` issue in the playbook during your CentOS6 testing :-)

sxa555:
Related: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/696

Willsparker:
@sxa555 Issue still seems to persist using CentOS6. 'wget' ing the URL does work though, the error message says:

```
[...]

PR #892, opened 2019-08-13
Changed the Unix PB to complete ccache

Willsparker:
I had to do this by using the `shell` module in ansible, which isn't really suggested, hence `warn=False`. Works for now, though could be looked at later.

helps towards: #807

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
@@ -9,11 +9,8 @@
   tags: ccache
 
 - name: Download ccache.tar.gz
-  get_url:
-    url: https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz
-    dest: /tmp/ccache.tar.gz
-    mode: 0440
-    validate_certs: False
+  command: wget -O /tmp/ccache.tar.gz https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz
+    warn=False
   when: ccache_status.stat.isdir is not defined
   tags: ccache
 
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
```diff
@@ -9,17 +9,14 @@
   tags: ccache
 
 - name: Download ccache.tar.gz
-  get_url:
-    url: https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz
-    dest: /tmp/ccache.tar.gz
-    mode: 0440
-    validate_certs: False
+  shell: cd /tmp && wget https://www.samba.org/ftp/ccache/ccache-3.4.2.tar.gz
```
sxa555: Can you change this to `wget -O /tmp/ccache-3.4.2.tar.gz https://...` instead of the `cd /tmp && ...`

sxa555: Or even wget -O /tmp/ccache.tar.gz to avoid the need for the version to be listed multiple times in the file

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
```diff
@@ -9,17 +9,14 @@
...
+    warn=False
   when: ccache_status.stat.isdir is not defined
   tags: ccache
 
 - name: Extract ccache
   unarchive:
-    src: /tmp/ccache.tar.gz
+    src: /tmp/ccach-3.4.2.tar.gz
```
sxa555: Missing `e` (Irrelevant if we switch to not having the version number in there)

sxa555:
Can you squash the commits and give it a more meaningful commit message please?

PR #895, opened 2019-08-13
Add heap size variable for Windows playbook

### LINKED ISSUE #882: Windows Playbook Java installation API broken
The Windows playbook roles that install the OpenJ9 Java doesn't work because it produces multiple zip files because of the new large heap binary. Adding `heap_size` variable to the URL should fix this. 

```
"OpenJDK8U-jdk_x64_windows_openj9_8u222b10_openj9-0.15.1.zip", "heap_size": "normal"
"OpenJDK8U-jdk_x64_windows_openj9_windowsXL_8u222b10_openj9-0.15.1.zip", "heap_size": "large"
```

The affected roles are `Java8`, `Java11`, and `Java12`.

Not sure what the default should be, but it can be added to `AdoptOpenJDK_Windows_Playbook/group_vars/all/adoptopenjdk_variables.yml`.

HusainYusufali:
- Heap size variable added to playbook. It defaults to normal
- To change heap size the extra variable must be added when running the playbook
- Playbook tested and still install hotspot with normal heap_size as the default
- Fixes #882 

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/group_vars/all/adoptopenjdk_variables.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/group_vars/all/adoptopenjdk_variables.yml
@@ -11,3 +11,4 @@ Nagios_Plugins: Disabled
 
 ### Default JDK ###
 bootjdk: hotspot
+heapsize: normal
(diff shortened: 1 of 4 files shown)


sxa555:
I don't have a particular issue with this, but is there an issue that this is a solution to? When would you need to choose the large heap variant?

HusainYusufali:
This is so the when installing OpenJ9 variants it doesn't provide multiple variants and fail the playbook. We don't need to have `heapsize` the variable. I can just make it normal and remove the `heapsize` variable

sxa555:
No I'm happy with that. I'll hold off on merging it until @Willsparker has a chance to test it though as it'll be useful to have a windows PR to test something he's working on for me ;-)

Willsparker:
Tested on a Windows Server 2012 Vagrant VM and it's all good 👍

PR #896, opened 2019-08-14
Added checks in PB test script to allow testing of branches

Willsparker:
Added checks that determine if the URL contains the word "tree". If it does, it makes a note of the branch name, and `git clone`s that single branch to test on. 
Also some small other additions:

- Changed `~` to `$HOME`

- Added CentOS7 to the list of tested OSs.

- Changed `set -eu` to `set -u`, as `-eu` was causing the test script to cancel when failing on a particular OS, without testing the proceeding OSs.

### DIFF
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
-set -eu
+set -u
+
+branchName='NULL'
+folderName=' '
+gitURL=' '
 
 # Takes all arguments from the script
 processArgs()
@@ -13,21 +17,34 @@ processArgs()
 
 setupFiles()
 {
-	cd ~
+	cd $HOME
 	mkdir -p adoptopenjdkPBTests || true
 	cd adoptopenjdkPBTests
 	mkdir -p logFiles || true
 }
 
-# Takes in git URL as arg 1, foldername as arg 2
+# Takes in git URL as arg 1, foldername as arg 2, branchName as arg 3
 setupGit()
 {
-	cd ~/adoptopenjdkPBTests
-	if [ ! -d "$2" ]; then		#if folder doesn't exist
-   		git clone $1
-	else				#if it does, ensure up to date
-		cd "$2"
-    		git pull $1
+	cd $HOME/adoptopenjdkPBTests
+	if [ "$3" = "NULL" ]; then
+		echo "Not a branch"
+		if [ ! -d "$2-master" ]; then
+   			git clone $1
+			mv $2 $2-master
+		else
+			cd "$2-master"
+    			git pull $1
+		fi
+	else
+		echo "Branch detected"
+		if [ ! -d "$2-$3" ]; then
+   			git clone -b $3 --single-branch $1
+			mv $2 "$2-$3"
+		else
+			cd "$2-$3"
+			git pull origin $3
+		fi
 	fi
 }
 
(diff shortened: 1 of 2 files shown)


ansible/pbTestScripts/testScript.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
-set -eu
+set -u
```
karianna: I think you want to leave the `-e` in there...

ansible/pbTestScripts/testScript.sh
```diff
@@ -1,5 +1,9 @@
...
+
+branchName='NULL'
+folderName=' '
```
karianna: Best not to assign anything to initialize rather than a single whitespace string.

ansible/pbTestScripts/testScript.sh
```diff
@@ -13,21 +17,34 @@ processArgs()
...
 setupGit()
 {
-	cd ~/adoptopenjdkPBTests
-	if [ ! -d "$2" ]; then		#if folder doesn't exist
-   		git clone $1
-	else				#if it does, ensure up to date
-		cd "$2"
-    		git pull $1
+	cd $HOME/adoptopenjdkPBTests
+	if [ "$3" = "NULL" ]; then
```
karianna: As a general comment, refactor the script so that $1, $2 and $3 are immediately assigned to meaningfully name variables.

Also, I'm not sure a single `=` sign is the safe way to check equality in shell

ansible/pbTestScripts/testScript.sh
```diff
@@ -13,21 +17,34 @@ processArgs()
...
+		echo "Not a branch"
```
karianna: General comment -> Give more context when echoing to the user

ansible/pbTestScripts/testScript.sh
```diff
@@ -56,36 +77,56 @@ destroyVM()
...
-	elif grep -q '\[ERROR\]' *$1.log
+	elif grep -q '\[ERROR\]' *$2.$1.log
 	then
 		printf "\n$1 playbook was stopped\n"
 	else
 		printf "\n$1 playbook succeeded\n"
 	fi
 }
 
+# Takes in the URL passed to the script
```
karianna: And does what with it? :-)

ansible/pbTestScripts/testScript.sh
```diff
@@ -56,36 +77,56 @@ destroyVM()
...
+splitURL()
+{
+	# breaks down url to array, and extracts info
+	IFS='/' read -r -a array <<< "$1"
```
karianna: Is IFS a global var or should it be a local one?

ansible/pbTestScripts/testScript.sh
```diff
@@ -56,36 +77,56 @@ destroyVM()
...
+	if [ ${array[@]: -2:1} == 'tree' ]
+	then
+		branchName=${array[@]: -1:1}
+		folderName=${array[@]: -3:1}
+		unset 'array[${#array[@]}-1]'
+		unset 'array[${#array[@]}-1]'
+		for I in "${array[@]}"
```
karianna: lowercase `i` if you're looping over a local var

ansible/pbTestScripts/testScript.sh
```diff
@@ -56,36 +77,56 @@ destroyVM()
...
+		do
+			gitURL="$gitURL$I/"
```
karianna: I think you're overriding this var each time, is that the intent?

ansible/pbTestScripts/testScript.sh
karianna: Hmm, saw your comment above...  OK, I'll let this one slide ;-)

karianna:
Useful change!

ansible/pbTestScripts/testScript.sh
Willsparker: Yep, it's to build back up the required git URL, without "tree" or the branch name in it.
i.e.
`https://github.com/AdoptOpenJDK/openjdk-infrastructure/tree/ccache` --> `https://github.com/AdoptOpenJDK/openjdk-infrastructure/`

Issue #864, opened 2019-07-30
Investigate feasibility of testing Windows playbooks under Vagrant


sxa555:
How easily can Windows systems be created and run via Vagrant? Can we automate the playbook testing on "clean" OS installs as we can now do on Linux (https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/842)?

Willsparker:
Written a vagrantfile that can boot a Windows 10 VM, however when using `vagrant ssh` it boots into a linux shell script. `cmd` can then be run from there.

Willsparker:
Created a vagrantfile that can boot a windows VM, and have a list of reliable steps towards running the playbook remotely. A `readme.md` file will need to be created, or an existing one altered to give steps on how to do this.
Overview of the steps taken:
1. `vagrant up` on Vagrantfile.WindowsS12
2. On windows vagrant machine:
  - `wget https://raw.githubusercontent.com/ansible/ansible/devel/examples/scripts/ConfigureRemotingForAnsible.ps1 -OutFile .\ConfigureRemotingForAnsible.ps1`
  - ` .\ConfigureRemotingForAnsible.ps1 -EnableCredSSP ` 
  - `ipConfig` to find the ip-address for the VM
3. (if on Mac host) `export OBJC_DISABLE_INITIALISE_FORK_SAFETY=yes`
4. On Host machine:
  - change `hosts.txt` in `../ansible/playbooks/AdoptOpenJDK_Windows_Playbook/` to include the Windows VM ip address.
  - add in `../AdoptOpenJDK_Windows_Playbook/group_vars/all/adoptopenjdk_variables.yml`, the line `ansible_winrm_transport=CredSSP`
  - From `../ansible` run `sudo ansible-playbook -i playbooks/AdoptOpenJDK_Windows_Playbook/hosts.win -u vagrant playbooks/AdoptOpenJDK_Windows_Playbook/main.yml`

.. and the playbook should run.

Issue #791, opened 2019-05-23
Enable Local Execution of Windows Playbook


adamfarley:
When running the Ansible playbook locally on Windows, there are various blockers which prevent successful execution. The ones we know about are listed below, as well as the work-arounds we used to resolve them.

Note: After all these changes, the playbook still don't run to completion locally. This is just the point where we gave up, and set up a remote ansible server.

The objective of this issue is to enable someone to install ansible and run the Windows playbook to completion locally without failure.

Known issues:

- The fact that there are PowerShell commands in the Windows playbook isn't advertised anywhere except inside the playbook.

- Said commands can't be run unless you've enabled script execution via, e.g:  Set-ExecutionPolicy Unrestricted

- Ansible can't just be installed. Users need to follow a guide like the one here:

 www.oznetnerd.com/installing-ansible-windows

- When configuring cygwin packages during installation, users need to choose the modules specified in the guide, except replacing all the "python" modules with python37 modules (python 2 modules can be added as-is).

- The file "/etc/ansible/hosts" needs to be created from scratch, with the content being: 
```
[local] 
1270.0.1
```

- Users need to edit main.yml file for windows, and replace 
```
"- hosts: etc etc"
```
with
```
- hosts: 127.0.0.1
  connection: local
```

- People need to run the pip install as a python module:
```
python37 -m pip install ansible
```

- Users need to install winrm using the command below, to prevent an error when ansible can't find it.
```
Run: python37 -m pip install "pywinrm>=0.2.2"
```

- The main.yml file should also be edited to remove the "Windows Updates" and "Jenkins" roles, because the former didn't work on a windows 10 machine (something about not being able to find the reboot_needed variable), and the latter fails for some other reason. Neither are needed if the user's machine is up to date, and they're running locally.

- the "/playbook_git_repo_location/group_vars/all/adoptopenjdk_variables.txt" file needs to be edited to replace
```
#ansible_password: CHANGE_ME
```
with
```
ansible_user: <your username>
ansible_password: <your password>
```

- When running the cygwin role, it restarts the computer after editing the path variable. This is a problem while running locally, as it stops the process, and you have to restart it again.

- When finally running the playbook command, ansible on windows doesn't recognise the -s option.

sxa555:
[...]

PR #900, opened 2019-08-16
Adding a Windows Server 2012 Vagrantfile and Set-Up instructions

Willsparker:
Having a Vagrant file for Windows Server 2012 allows the testing of the Windows playbook on a local VM. The README was updated to contain instructions on how to set this up, and subsequently run the playbook on the VM. The Vagrantfile takes care of configuring the VM to allow winRM connections, so only changes to the host machine is needed.

Fixes: #864


ansible/README.md
```diff
@@ -97,6 +97,19 @@ Yes, in order to access the package repositories (we will perform either `yum in
 
 4) The Ansible playbook will download and install any dependencies needed to build OpenJDK
 
+## How do I run the playbook on a Windows Machine?
+
+As Ansible can't run on Windows, it has to run remotely. To test the playbook, a Vagrant VM can be used.
```
karianna: Can run on windows so you have to run remotely?  Doesn't quite make sense to me....

ansible/README.md
```diff
@@ -97,6 +97,19 @@ Yes, in order to access the package repositories (we will perform either `yum in
...
+You can do this by following these steps:
+
+1) `ln -sf Vagrantfile.WindowsS12 Vagrantfile && vagrant up` in the `../ansible` directory
+2) Note the IP address that is output from running `vagrant up` and place it into a text file. e.g. `hosts.win`. Place this text file in `../anisble/playbooks/AdoptopenJDK_Windows_Playbook`
```
karianna: type in the directory path for ansible

ansible/Vagrantfile.WindowsS12
```diff
@@ -0,0 +1,20 @@
...
+# vi: set ft=ruby :
+
+$script = <<SCRIPT
+Start-Process powershell -Verb runAs
+wget https://raw.githubusercontent.com/ansible/ansible/devel/examples/scripts/ConfigureRemotingForAnsible.ps1 -OutFile .\\ConfigureRemotingForAnsible.ps1
+.\\ConfigureRemotingForAnsible.ps1 -EnableCredSSP
+ipConfig
+SCRIPT
+
+Vagrant.configure("2") do |config|
```
karianna: 2 is a magic number

ansible/Vagrantfile.WindowsS12
```diff
@@ -0,0 +1,20 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+$script = <<SCRIPT
```
karianna: Comment at the top of this block explaining roughly what it does

ansible/README.md
```diff
@@ -97,6 +97,20 @@ Yes, in order to access the package repositories (we will perform either `yum in
 
 4) The Ansible playbook will download and install any dependencies needed to build OpenJDK
 
+## How do I run the playbook on a Windows Machine?
+
+As Ansible can't run on Windows, it has to be run on a seperate system and then pointed at a Windows machine (such as a VM). To test the playbook, a vagrant VM can be used.
```
karianna: Thanks for the update, but I'll still bite here :-) - https://www.ansible.com/integrations/infrastructure/windows why can't we run on windows directly?

Willsparker: I was going off what the FAQs say in the Ansible documentation:
https://docs.ansible.com/ansible/latest/user_guide/windows_faq.html

Also doing it this way seemed easier than dealing with the problems found in issue #791

sxa555: @karianna That page is talking about ansible's Windows modules that you can use when remotely connecting in, not executing the playbooks natively on the host. I believe it should be doable but it's a bit of a pig to set up. We have https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/791 to cover anything we can do to make this easier in the future, but at the moment we don't easily support it or test it.

ansible/Vagrantfile.Win2012
```diff
@@ -0,0 +1,23 @@
...
+Start-Process powershell -Verb runAs
+wget https://raw.githubusercontent.com/ansible/ansible/devel/examples/scripts/ConfigureRemotingForAnsible.ps1 -OutFile .\\ConfigureRemotingForAnsible.ps1
+.\\ConfigureRemotingForAnsible.ps1 -EnableCredSSP
+ipConfig
+SCRIPT
+
+# 2 = version of configuration file for Vagrant 1.1+ leading up to 2.0.x
+Vagrant.configure("2") do |config|
+
+  config.vm.define :adoptopenjdkW10 do |adoptopenjdkW10|
```
sxa555: Should probably be `W2012` since this isn't Windows 10

ansible/README.md
karianna: Got it thanks both.

PR #918, opened 2019-08-28
Update to testScript.sh to allow for Parallelisation

Willsparker:
Requested by @sxa555 .
Changes to the script include:
- Option to build a single, user specified OS, VM, or all of them as before.
- Option to build a JDK on the VM, or not, thereby allowing JUST playbook testing if needed.
- Tidied up passing arguments to the script.
- Added a usage message.

### DIFF
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -35,12 +35,29 @@ processArgs()
 usage()
 {
 	echo
-	echo "Usage: ./testScript.sh 	--vagrantfile | -v <OS_Version>		Specifies which OS the VM is
+	echo "Usage: ./testScript.sh			--vagrantfile | -v <OS_Version>		Specifies which OS the VM is
 					--all | -a 				Builds and tests playbook through every OS
 					--retainVM | -r				Option to retain the VM once building them
 					--build | -b				Option to enable testing a native build on the VM
 					--URL | -u <GitURL>			The URL of the git repository
 					--help | -h				Displays this help message"
+}
+
+defaultVars()
+{
+	#Sets WORKSPACE to home if WORKSPACE is empty or undefined. 
+	if [ ! -n "${WORKSPACE:-}" ]; then
+		echo "WORKSPACE not found, setting it as environment variable 'HOME'"
+		WORKSPACE=$HOME
+	fi
+	if [ "$gitURL" == "" ]; then
+		echo "No GitURL specified; Defaulting to adoptopenjdk/openjdk-infrastructure"
+		gitURL=https://github.com/adoptopenjdk/openjdk-infrastructure
+	fi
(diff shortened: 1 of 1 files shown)


ansible/pbTestScripts/testScript.sh
```diff
@@ -110,24 +167,33 @@ splitURL()
...
-	if [[ $2 = "n" ]]
-	then
-		destroyVM
-	fi
-done
-for OS in Ubuntu1804 Ubuntu1604 CentOS6 CentOS7
-do
-	searchLogFiles $OS $branchName
-done
+# Testing all of the OSs
```
sxa555: To avoid duplication I'd suggest setting `vagrantOS` to `"Ubuntu1604 Ubuntu1804 CentOS6 CentOS7"` in the parameter parsing section as a default, and then usiong `for OS in $vagrantOS` - that will render this if section irrelevant.

ansible/pbTestScripts/testScript.sh
```diff
@@ -4,15 +4,70 @@ set -u
...
+		local opt="$1";
+		shift;
+		case "$opt" in
+			"--Vagrantfile" | "-v" )
+				vagrantOS="$1"; shift;;
+			"--all" | "-a" )
+				vagrantOS="all";;
+			"--build" | "-b" )
+				testNativeBuild=true;;
+			"--VM" | "-vm" )
```
sxa555: Not sure this option is clear. Perhaps `--retainVM` and `-r`?

ansible/pbTestScripts/testScript.sh
```diff
@@ -4,29 +4,104 @@ set -u
...
+					--retainVM | -r				Option to retain the VM once building them
+					--build | -b				Option to enable testing a native build on the VM
+					--URL | -u <GitURL>			The URL of the git repository
+					--help | -h				Displays this help message"
+}
+
+defaultVars()
+{
+	#Sets WORKSPACE to home if WORKSPACE is empty or undefined. 
+	if [ ! -n "${WORKSPACE:-}" ]; then
```
sxa555: What's this construct doing that wouldn't be covered by "if [ -z "$WORKSPACE" ]?

ansible/pbTestScripts/testScript.sh
```diff
@@ -4,29 +4,104 @@ set -u
 branchName='NULL'
```
sxa555: Why `NULL` as the default here instead of `''` which is used for everything else?

ansible/pbTestScripts/testScript.sh
Willsparker: `if [ -z "$WORKSPACE" ]` will error if $WORKSPACE is undefined

sxa555: It's the `set -u` that's causing that

Issue #851, opened 2019-07-29
Ensure cuda SDK is on all build machines


sxa555:
From @keithc-ca:
`Only the x64 Linux build is CUDA-enabled. I suspect the build machines for Windows and ppc64le Linux don’t have the CUDA toolkit installed in the expected place.`

EDIT: Requirement from the OpenJ9 call today is to have this ready for the next quarterly release

sxa555:
Autodetection added in https://github.com/AdoptOpenJDK/openjdk-build/pull/1187/files

sxa555:
Should be on all build machines now - will keep an eye out and see if any miss it. Separate PR to enable it on CentOS/ppc64le but I've run it on there already so it's ready to go and should build with it tonight. This issue should close once the PR is approved/mereged

sxa555:
@keithc-ca That path does exist on the windows machine that the referenced job was on, although it just contains 3 `.txt` files - I'm guessing from what you've said that's not right?

sxa555:
OK the playbook for that isn't installing it fully then. I'll put it on the machines today and assign to @willsparker to fix the playbook.

sxa555:
Removing `CUDAToolkit_9.0` from https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/8558f94fecce59121d07f59164c686184d7cb351/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml#L19 appears to make it install everything, including the files we want but I suspect there's a smaller set we could do with selecting in the silent installer.

keithc-ca:
According to [1], I think the command you want is:
```
C:\temp\cuda_9.0.176_windows_network-exe.exe -s compiler_9.0
```

[1] https://docs.nvidia.com/cuda/archive/9.0/cuda-installation-guide-microsoft-windows/index.html

keithc-ca:
Verified that CUDA is enabled in `OpenJDK8U-jdk_x64_windows_openj9_2019-08-31-14-43.zip` and `OpenJDK11U-jdk_x64_windows_openj9_2019-09-03-05-14.zip`. Thanks, everyone.

PR #922, opened 2019-08-30
Changed NVidia_Cuda_Toolkit role to fully install on Windows

Willsparker:
Changed as per the conversation on #851 .
Tested on a Windows 2012R2 Vagrant VM. Once playbook was ran with the changes, `include` is in the `C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v9.0` directory, and contains `cuda.h`, as that's what the build checks for.


sxa555:
Below interesting it works be good to verify on a Windows 10/Server 2016 system too since that uses a different path through the code. Adam's system will likely do that if he's getting the playbook working on 2016 so possibly encourage him to pick up this change

Willsparker:
Okay, I'll help him get that working today and test it.

Willsparker:
Okay, I was able to create a Win2016 Server Vagrant VM, ran the playbook and it put the files in the correct place as Win2012 did

PR #926, opened 2019-09-03
Adding correct tags to the Windows Playbook

Willsparker:
As per @sxa555 's request.
The addition of these tags are to remain consistent with the Unix playbook's Jenkins tag, and allows testing of the playbook that don't require parts of the `Vendor_Files` folder.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freemarker/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Freemarker/tasks/main.yml
@@ -7,11 +7,15 @@
   win_stat:
     path: 'C:\openjdk\freemarker.jar'
   register: freemarker_installed
-  tags: Freemarker
+  tags:
+    - Freemarker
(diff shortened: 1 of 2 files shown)


Willsparker:
I have put Freemarker tasks under the `adoptopenjdk` tag, as this is the case in the Unix playbook also. Freemarker appears to be the only role that appears in the Unix Playbooks and has the `adoptopenjdk` tag, that is also in the Windows playbook.

Willsparker:
Comparing the FreeMarker role between the Unix and Windows playbooks, it seems Windows' Freemarker doesn't rely on the Jenkins role being run, whereas Unix's does. However, it also seems that the Windows' FreeMarker just downloads the jar and then does nothing with it.

Willsparker:
@sxa555 Taken the WIP off the PR. The FreeMarker tag doesn't need the `adoptopenjdk` tag, but if you wanted it consistent with the Unix playbook, (i.e, it only executes if the Jenkins role does), then it can stay on. Doesn't seem any other role needs the `adoptopenjdk` tag.

Issue #938, opened 2019-09-13
Add build testing into the automated playbook testing


sxa555:
Follow on to #842 and #924 - we should verify that the builds work on a machine set up with the playbooks

PR #942, opened 2019-09-19
Modified testScript.sh to allow for Win2012 to build JDK

Willsparker:
Fixes: #938 

With this, all OS' currently supported in `testScript.sh` should be able to test a build.


ansible/pbTestScripts/buildJDKWin.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -eu
+export TARGET_OS=linux
```
karianna: Is linux right?

ansible/pbTestScripts/buildJDKWin.sh
```diff
@@ -0,0 +1,11 @@
...
+export ARCHITECTURE=x64
+export JAVA_TO_BUILD=jdk8u
+export VARIANT=openj9
```
karianna: Probably should be hotspot by default

ansible/pbTestScripts/buildJDKWin.sh
```diff
@@ -0,0 +1,11 @@
...
+export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
+export JDK_BOOT_DIR=/usr/lib/jvm/java-8-openjdk-amd64
```
karianna: should that be JDK8_BOOT_DIR?

ansible/pbTestScripts/buildJDKWin.sh
Willsparker: I tried it with `windows` but it just gave error messages, seems to work with `linux` though

sxa555: That doesn't sound right at all - what error messages?

Willsparker: Actually! My mistake- Just retried it and it seems to work fine.

Willsparker:
Adding [WIP] tag to this PR, as I'm consistently having an error when trying to build a JDK on Windows, in that the build hangs up when attempting to move the workspace- though it also can't find a viable version of Java in a directory that definitely has one, so I'm going to investigate if that is due to an exporting an environment variable issue.

Willsparker:
I've hit an issue where, having built gradle, and getting `./security/cacerts/`, the build scripts attempt to access a recursive directory : 
```
/cygdrive/c/vagrant/pbTestScripts/workspace
rsync: send_files failed to open "/cygdrive/c/vagrant/pbTestScripts/workspace/build/installedalsa/include/alsa/alsa/alsa/alsa ... etc"
```
I've got this from running the following:
```
#!/bin/bash
set -eu
export TARGET_OS=windows
export ARCHITECTURE=x64
export JAVA_TO_BUILD=jdk8u
export VARIANT=openj9
export JDK7_BOOT_DIR=c:/openjdk/jdk7
export JDK8_BOOT_DIR=/usr/lib/jvm/java-8-openjdk-amd64
export JDK_BOOT_DIR=c:/openjdk/jdk7
export JDK11_BOOT_DIR=c:/openjdk/jdk-11

export PATH=/usr/bin:$PATH 
c:/openjdk-build/build-farm/make-adopt-build-farm.sh
```

Willsparker:
@sxa555 @karianna I'm not convinced it's possible to build a JDK on Windows, complete automatically- due to the problems listed under https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/553, there is a requirement for some user to go onto the VM and install `VS2010 SP1` before a build can be completed.

karianna:
@Willsparker merge conflicts...

Willsparker:
I've got to the point where the playbook installs the dependencies correctly, and a JDK can now be built on the Windows machine without additional user input- however, when attempting to do this through the script, one of these error messages occur:
```
fatal: mmap failed: Resource temporarily unavailable
error: https://github.com/adoptopenjdk/openjdk-jdk8u.git did not send all necessary objects
```
```

      0 [main] git 579 fhandler_disk_file::fixup_mmap_after_fork: requested 0x6FFF8FA0000 != 0x0 mem alloc base 0x0, state 0x10000, size 132998073876480, Win32 error 1455
   3442 [main] git 579 C:\cygwin64\usr\libexec\git-core\git.exe: *** fatal error in forked process - recreate_mmaps_after_fork_failed
   7928 [main] git 579 cygwin_exception::open_stackdumpfile: Dumping stack trace to git.exe.stackdump
      1 [main] git 578 dofork: child -1 - forked process 2112 died unexpectedly, retry 0, exit code 0x100, errno 11
error: cannot fork() for index-pack: Resource temporarily unavailable
fatal: fetch-pack: unable to fork off index-pack
```
Both occur after saving `./security/cacerts`. Google specifies it's a cygwin error, it seems.

Willsparker:
Errors have been hit. When attempting to run the `BuildJDKWin.sh` script via `vagrant powershell -c`, when the version of git that is used is the one installed with the playbook (known as WinGit from now), an error will occur where it successfully clones a repository to `/tmp/openjdk-build-dir/`, however when attempting to `cd` into that repository, it doesn't exist- this is because WinGit has caused the repo to be at `C:/tmp/openjdk-build-dir/`
Source : https://github.com/git-for-windows/git/issues/568
Keeping to Cygwin's inbuilt Git, causes an error in which cloning a repository with result in a well known `fork(): Resource Temporarily Unavailable` error or `fatal: mmap failed: Resource temporarily unavailable` error. This can supposedly be solved by rebasing the Cygwin packages, however I'm yet to make this work, let alone implement it into the script so automation of building the JDK can be achieved.

Willsparker:
The new direction I've taken this is to add a role into the playbook that will run the build script on the machine. The previous problems are therefore side-stepped as a different piece of software is communicating with the VM. This contains the tag `build`, and so can be skipped when a JDK doesn't need to be built on the VM.

ansible/pbTestScripts/buildJDKWin.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -eu
+export TARGET_OS=windows
+export ARCHITECTURE=x64
+export JAVA_TO_BUILD=jdk8
+export VARIANT=hotspot
+export JDK7_BOOT_DIR=/cygdrive/c/openjdk/jdk7
+export JDK11_BOOT_DIR=/cygdrive/c/openjdk/jdk-11
```
karianna: why do we exprt java 11 for an 8 build?

ansible/pbTestScripts/testScript.sh
```diff
@@ -144,6 +144,14 @@ testBuild()
 	vagrant ssh -c "cd /vagrant/pbTestScripts && ./buildJDK.sh"
 }
 
+testBuildWin()
+{	
+	# Ensures the git config won't change line endings
+        vagrant powershell -c "Start-Process powershell -Verb runAs; C:/cygwin64/bin/sed -i -e 's/autocrlf.*/autocrlf = false/g' C:\\ProgramData/Git/config"
```
karianna: indent inconsistency.

ansible/pbTestScripts/testScript.sh
```diff
@@ -197,7 +205,11 @@ startVMPlaybookWin()
 		vagrant powershell -c "Start-Process powershell -Verb runAs; \$size = (Get-PartitionSupportedSize -DriveLetter c); Resize-Partition -DriveLetter c -Size \$size.SizeMax"
 	fi
 	# run the ansible playbook on the VM & logs the output.
-	ansible-playbook -i playbooks/AdoptOpenJDK_Windows_Playbook/hosts.win -u vagrant --skip-tags jenkins,adoptopenjdk playbooks/AdoptOpenJDK_Windows_Playbook/main.yml 2>&1 | tee $WORKSPACE/adoptopenjdkPBTests/logFiles/$folderName.$branchName.$OS.log
+	ansible-playbook -i playbooks/AdoptOpenJDK_Windows_Playbook/hosts.win -u vagrant --skip-tags jenkins,adoptopenjdk,build playbooks/AdoptOpenJDK_Windows_Playbook/main.yml 2>&1 | tee $WORKSPACE/adoptopenjdkPBTests/logFiles/$folderName.$branchName.$OS.log
+	if [[ "$testNativeBuild" = true ]]; then
+                testBuildWin
```
karianna: indent whitespace?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
```diff
@@ -62,3 +62,4 @@
     - nasm                        # OpenJ9
     - Rust                        # IcedTea-Web
     - IcedTea-Web                 # For Jenkins webstart
+    - Build                       # Role to build JDK8
```
karianna: Only JDK 8?

Willsparker: For now- I can look into generalising that role (a lot like I did in #955) so we can build multiple, and then give the option to choose which JDK to build in the script. Currently this is just meant to test if the playbook installs all the dependencies required for 8

ansible/pbTestScripts/buildJDKWin.sh
Willsparker: Initially, it was to run Gradle- however at some point through this PR, the script figured out how to detect jdk-8, so it's been removed (though I did double check).

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
karianna: OK, we can update that in a later PR

Willsparker:
As per @sxa555 's request, I've removed the role from the playbooks. Instead, the testScript has instead been altered to used the command `ansible` to run the `buildScript` on the Windows VM, however just running the ad-hoc command, therefore it doesn't require addition to the playbook as a role.

ansible/pbTestScripts/testScript.sh
```diff
@@ -144,6 +144,15 @@ testBuild()
 	vagrant ssh -c "cd /vagrant/pbTestScripts && ./buildJDK.sh"
 }
 
+testBuildWin()
+{
+	# Ensures the git config won't change line endings
+	vagrant powershell -c "Start-Process powershell -Verb runAs; C:/cygwin64/bin/sed -i -e 's/autocrlf.*/autocrlf = false/g' C:\\ProgramData/Git/config"
+	vagrant powershell -c "cd C:/; if (-not (Test-Path C:/openjdk-build -PathType Container) ) { echo 'Cloning openJDK-build repo' ; git clone https://github.com/adoptopenjdk/openjdk-build ; sleep 3 }"
+	# Runs the build script via ansible
```
sxa555: Can you add a comment into this script saying why we're not using `vagrant powershell`to make it clear why we're doing something that superficially seems a bit odd ... Possibly include a link to https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/942#issuecomment-539946564

Issue #939, opened 2019-09-13
Add basic test verification to the playbook testing


sxa555:
Follow on to #938 - once we've verified that the build prereqs are installed correctly by ansible we should verify that we have things needed to run the test harnesses - the `MachineInfo` target mentioned by @smlambert in https://adoptopenjdk.slack.com/archives/C53GHCXL4/p1564580457030100 would seem a reasonable choice and should verify the various perl and ant prereqs that we have.

Willsparker:
@smlambert I've tried running that test on an Ubuntu1604 VM, however it was unable to find `MachineInfo` as a test target. This was after declaring the following environment variables:
```
BUILD_LIST=systemtest,openjdk
JDK_CUSTOM_TARGET=MachineInfo
```
This was after following the instructions here:
https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally

smlambert:
BUILD_LIST=system,openjdk
TARGET=MachineInfo

smlambert:
In fact, for the purpose of this test, you do not need openjdk dir in your build list.  The build list tells us which test directory you want to compile, and which test targets (defined in playlists) that you want to generate. MachineInfo is a test target defined in playlists in system directory, so you'd only need to indicate that dir to enable that test.

If you do not need to run any openjdk tests, you do not need to build that directory.

PR #943, opened 2019-09-20
Added functionality for testScript.sh to run a test on a built JDK

Willsparker:
Ref: #939 

Currently only implemented for the unix playbooks. If the option is specified, the script will run the `MachineInfo` test against a version of JDK8 that is built on the VM, having been setup by the playbook. It won't allow a test to be ran unless a JDK is built.

### DIFF
--- a/ansible/pbTestScripts/testJDK.sh
+++ b/ansible/pbTestScripts/testJDK.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+mv /vagrant/pbTestScripts/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8* ~
+export TEST_JDK_HOME=$(find ~ -maxdepth 1 -type d -name "*jdk8u*"|grep -v ".*jre.*")
+cd $HOME && mkdir -p testLocation
+cd testLocation && git clone https://github.com/adoptopenjdk/openjdk-tests
+cd openjdk-tests
+./get.sh -t $HOME/testLocation/openjdk-tests
+cd TestConfig
+make -f run_configure.mk
+export BUILD_LIST=MachineInfo
+make compile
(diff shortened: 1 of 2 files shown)


ansible/pbTestScripts/testJDK.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+mv /vagrant/pbTestScripts/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8* ~
+export TEST_JDK_HOME=$(find ~ -maxdepth 1 -type d -name "*jdk8u*"|grep -v ".*jre.*")
```
sxa555: Any reason not to just set `TEST_JDK_HOME` to `/vagrant/pbTestScripts/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8*`?

Willsparker: There's 2 directories in there- the JDK to test, and the JRE, which has the same name but `-jre` at the end

ansible/pbTestScripts/testJDK.sh
```diff
@@ -0,0 +1,13 @@
...
+cd $HOME && mkdir -p testLocation
+cd testLocation && git clone https://github.com/adoptopenjdk/openjdk-tests
+cd openjdk-tests
+./get.sh -t $HOME/testLocation/openjdk-tests
+cd TestConfig
+make -f run_configure.mk
+export BUILD_LIST=MachineInfo
+make compile
+make _MachineInfo
```
sxa555: Looks like this doesn't have a newline character at the end - can you add one please then I'll approve

Willsparker: Done! 👍

PR #946, opened 2019-10-01
Added install component to Windows PB CUDA role

Willsparker:
Whilst going through the PR to test builds on Windows, I've found an issue that the Windows PB doesn't install the NVidia Management Library as part of the Cuda toolkit, (an error stated that it couldn't find `nvml.h`). With the addition of this component put into the CUDA installation, the correct file is there.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
@@ -16,7 +16,7 @@
   tags: NVidia_Cuda_Toolkit
 
 - name: Install NVidia CUDA toolkit - Windows 7,8 and Server 2008,2012
-  win_shell: C:\temp\cuda_9.0.176_windows_network-exe.exe -s compiler_9.0
+  win_shell: C:\temp\cuda_9.0.176_windows_network-exe.exe -s compiler_9.0 nvml_dev_9.0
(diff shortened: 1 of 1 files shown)


sxa555:
Interesting - I don't think we had that on all the build boes - do you have the fragment of the log that showed this (i.e. the full context of the error message so we can include it here for anyone who searches on the error later)

Willsparker:
@sxa555 I think this came around with the change in #922 and #851 , as we're now not installing all the components of NVidia Cuda Toolkit like we were before. The error message is as below: 

```
C:\cygwin64\tmp\openjdk-build-dir\workspace\build\src\build\windows_x86_64-normal-server-release\vm\compiler\runtime\GPHelpers.cpp(36) : fatal error C1083: Cannot open include file: 'nvml.h' : No such file or directory
make[4]: *** [C:/cygwin64/tmp/openjdk-build-dir/workspace/build/src/build/windows-x86_64-normal-server-release/vm/compiler/build/rules/win-msvc/common.mk:83: C:/cygwin64/tmp/openjdk-build-dir/workspace/build/src/build/windows-x86_64-normal-server-release/vm/objs/compiler/runtime/GPUHelpers.obj] Error 2
```

sxa555:
@keithc-ca FYA `compiler_9.0` was introduced as  part of https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/851#issuecomment-526179321 - we're going to modify it as per this PR.

If you have any objections to this change feel free to voice them, otherwise i'll go ahead and merge.

keithc-ca:
This seems right to me.

Issue #383, opened 2018-06-17
Add VS2010 SP1 to playbook


gdams:
we need VS2010 SP1 to be added to the ansible windows playbooks.

gdams:
error for reference:

```bash
jvm.obj : warning LNK4197: export 'JVM_GetThreadStateNames' specified multiple times; using first specification
   Creating library jvm.lib and object jvm.exp
LINK : fatal error LNK1123: failure during conversion to COFF: file invalid or corrupt
NMAKE : fatal error U1077: 'C:\progra~2\micros~2.0\vc\bin\amd64\link.exe' : return code '0x463'
Stop.
NMAKE : fatal error U1077: 'cd' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\BIN\amd64\nmake.EXE"' : return code '0x2'
Stop.
make[3]: *** [generic_build2] Error 2
make[2]: *** [product] Error 2
make[1]: *** [/cygdrive/c/Users/jenkins/workspace/openjdk8_build_x86-64_windows/openjdk/build/windows-x86_64-normal-server-release/hotspot/_hotspot.timestamp] Error 2
make: *** [hotspot-only] Error 2
```

Willsparker:
Related #553 and this is involved in the work I'm doing for #942

PR #947, opened 2019-10-02
Added VS2010_SP1 role to Windows playbook

Willsparker:
Fixes: #383 
Ref: #553 , #942 

Found a way that should download and install VS2010_SP1 from the Windows Playbook. Must have MSVS2010 installed first. I've tested this on a Windows VM that did have MSVS2010 installed prior, but having ran MSVS_2010 and the new VS2010_SP1 tags on the playbook, it showed `Microsoft Visual Studio 2010 Version 10.0.40219.1 SP1Rel` under `help --> about microsoft visual studio`, which I believe means SP1 installed correctly.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -35,6 +35,7 @@
     - Windows_Updates
     - Jenkins                     # AdoptOpenJDK infrastructure
     - MSVS_2010
+    - VS2010_SP1
     - cygwin
     - Firefox
     - Strawberry_Perl             # Testing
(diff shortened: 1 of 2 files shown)


sxa555:
(I can't merge until you fix the travis errors though)

Willsparker:
@sxa555 I was googling to try to find a way of downloading VS2010SP1, and I found this atricle : https://old.windowsvalley.com/download-visual-studio-2010-service-pack-1/ which has a link at the bottom ( The one in the playbook role ), that seems to come from Microsoft, but doesn't require signing in to download it. As for the `/q` problem from before, I'm not sure why it seems to be working, honestly- it may be due to installing a different version VS2010SP1 to the one in #553

Willsparker:
Just FYI, the Travis check were initially failing because there was no "when" in the tasks in the role- I've fixed this by `stat` checking the setup file has been downloaded, which is probably not the best way of doing it, but I couldn't find any extra files that VS2010SP1 adds to the Visual Studio 10.0 folder in Program Files.

PR #949, opened 2019-10-03
Altering testScript to run a test on a built JDK on a Windows VM

Willsparker:
Fixes: #939 

This is the last OS that is currently supported that needs to have its built JDK tested. Currently WIP until #942 is finished and working.


Willsparker:
Much like #942 I've found the best way of communicating with the local Windows VM is to just use Ansible again. Prototyping has managed to get the same output as what happens when the Unix test is executed, so I'll go down this route.

Willsparker:
The changes to `testScript.sh` are yet to be committed as I'm waiting for #942 to be merged, to avoid merge conflicts.
Otherwise, this should allow the JDK to be tested. As mentioned before, an attempt of using `vagrant powershell -c " ... "` was there, however Ansible proved much easier to get working.

PR #950, opened 2019-10-04
Added disksize parameter to Win2012 Vagrantfile

Willsparker:
Required to allow for the Ansible Playbook to fully run through - otherwise it ends up running out of diskspace around the `Clang` roles. Also added info into `README.md` as it requires the host machine install a vagrant plugin.

### DIFF
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#/bin/bash
 set -eu
 
 branchName=''
@@ -66,6 +66,10 @@ checkVars()
 		echo "No Vagrant OS specified; Defaulting to testing all of them"
 		vagrantOS="all"
 	fi
+        if [[ ! $(vagrant plugin list | grep 'disksize') ]]; then
+                echo "Can't find vagrant-disksize plugin, installing . . ."
+                vagrant plugin install vagrant-disksize
+        fi
 }
 
 checkVagrantOS()
@@ -166,6 +170,8 @@ startVMPlaybook()
 startVMPlaybookWin()
 {
 	local OS=$1
+	# The number of bytes the disk should be (this is 95GB in bytes)
+	local diskSizeBoundary=102005473280;
 	if [ "$branchName" == "" ]; then
 		cd $WORKSPACE/adoptopenjdkPBTests/$folderName-master/ansible
 		branchName="master"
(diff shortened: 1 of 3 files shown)


Willsparker:
This didn't resize the volume of the C: drive on vagrant up and so wouldn't work for automation. Trying something new.

Willsparker:
A method using Powershell has come to light that can make use of this plugin, and then extend the volumes. I've found another method to give the extra disk space to the VM, without the use of a plugin, but this is a much more convoluted and less stable way of doing it.

Willsparker:
With the latest commit the Windows Vm will boot with a 100GB C drive, which is more than enough to run the playbook to completion and build a JDK. This solves the issue of giving the VM the disk-space but it not utilising it.

ansible/pbTestScripts/testScript.sh
```diff
@@ -184,6 +188,12 @@ startVMPlaybookWin()
 		# Add the "ansible_winrm_transport" to adoptopenjdk_variables.yml
 		echo -e "\nansible_winrm_transport: credssp" >> playbooks/AdoptOpenJDK_Windows_Playbook/group_vars/all/adoptopenjdk_variables.yml
 	fi
+        # If the disk size is less than 970MB, resize the C: drive to be as big as possible
+	local sizeOfDisk=$(du -k $WORKSPACE/.vagrant.d/boxes/mwrock*/*/virtualbox/*.vmdk | cut -f 1)
+        if [ $sizeOfDisk -lt 9700000 ]; then
```
sxa555: Is there an extra zero in here?

karianna:
> Assuming that 970000 can be adjusted (or otherwise justified!) this looks good

Sometimes putting magic numbers into CONST vars helps with this :-)

ansible/pbTestScripts/testScript.sh
Willsparker: Indeed there is, however, I've noticed that that file size doesn't change even once the volume is extended... I'm not really sure why, so I'm going to find a different way to do that `if` statement

Willsparker:
> > Assuming that 970000 can be adjusted (or otherwise justified!) this looks good
> 
> Sometimes putting magic numbers into CONST vars helps with this :-)

The script will automatically increase it to it's max size, which I believe is about 99.7GB- I thought 970MB would be a fair boundary to test from, to see if the volume has been raised before or not- however as I've commented, i'm going to rethink the `if` statement as it doesn't seem to work :-)

Willsparker:
I've taken out the `if` statement, and it will always try to resize the disk. There's no big issue if this occurs when the disk is already at max size apart from it error-ing, hence the `|| true`, but this is unlikely as the Jenkins builds are made to create a completely new VM anyway.

sxa555:
How about running `Get-Partition | select size` to get the current size of the disk?

ansible/pbTestScripts/testScript.sh
```diff
@@ -188,8 +190,13 @@ startVMPlaybookWin()
 		# Add the "ansible_winrm_transport" to adoptopenjdk_variables.yml
 		echo -e "\nansible_winrm_transport: credssp" >> playbooks/AdoptOpenJDK_Windows_Playbook/group_vars/all/adoptopenjdk_variables.yml
 	fi
-        # Resize the c drive, or return true if it's already been resized
-        vagrant powershell -c "Start-Process powershell -Verb runAs; \$size = (Get-PartitionSupportedSize -DriveLetter c); Resize-Partition -DriveLetter c -Size \$size.SizeMax" || true
+	# getting the current c drive information and cutting it down to the number of GB it is.
+	export currentDiskSize=$(vagrant powershell -c "Start-Process powershell -Verb runAs; Get-Partition -Driveletter c" | grep "GB")
+	currentDiskSize=$(echo $currentDiskSize | cut -d ' ' -f 5 | cut -d '.' -f 1)
+	if [[ $currentDiskSize -lt $diskSizeBoundary ]]; then
```
sxa555: My only concern with doing this is that if the units are in TB then this could get rather confused - can you run `vargrant powershell` in a way that allows you to run the `select` statement afterwards to show the raw number without units?

Willsparker: Fixed :) It's now done in bytes

PR #953, opened 2019-10-07
Added Java13 to Windows and Unix playbooks

Willsparker:
fixes: #951 
Done by copying the respective java 12 files, and `sed`ing 12 to 13 (and then checked). 
Tested on Ubuntu1604 and Windows 2012, and they appear in the correct folders :-)


Willsparker:
This has been implemented into PR #955

Issue #954, opened 2019-10-07
Make ansible roles for adoptopenjdk common across versions (9+)


sxa555:
At the moment we have separate directories and roles in the ansible playbooks for installing java 8-13. This is resulting in a lot of duplication of directories that only differ in version number. Instead of just adding to the duplication by adding a java 13 version, I would suggest creating a generic "adoptopenjdk" role which can take a parameter with the verison number, and then calling it multiple times.

Bear in mind that we should leave `adoptopenjdk8` alone as it uses a separate directory naming scheme, whereas versions 9 and above all use `jdk-<version>` directory.

We can use the [local_srcinstall](https://github.com/sxa555/local_srcinstall) role as an example of how to do this with a parameter

PR #955, opened 2019-10-07
Replaced Individual Java roles with single one in PBs

Willsparker:
fixes: #954 
fixes: #951 

A more efficient way to install all the versions of OpenJDK we want on the machines, and it makes it easier to add to the list over time. I've also removed the unneeded JDKs (9 and 11), as we're not building 10 or 12 anymore.


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -39,10 +39,16 @@
...
-    - adoptopenjdk11              # JDK12 build bootstrap
-    - adoptopenjdk12              # JDK13 build bootstrap
+    - role: adoptopenjdk_install
+      jdk_version: 9
+    - role: adoptopenjdk_install
+      jdk_version: 10
+    - role: adoptopenjdk_install
+      jdk_version: 11
+    - role: adoptopenjdk_install
+      jdk_version: 12
```
sxa555: Can you add the comment back into these saying "JDK13 build bootstrap" etc.

Willsparker:
Just FYI, I did try doing something along the lines of :
```
role: adoptopenjdk_install
jdk_version: {{ item }}
with_items:
  - 9
  - 10
  ...
```
But I was getting an error saying something along the lines of `item` is undefined in the adoptopenjdk_install role, so opted for this instead.

PR #960, opened 2019-10-16
pbTest: Removes folder once completed

Willsparker:
Removes the folder once the test has been completed. This will stop builds on the Jenkins machine from failing due to changes to the git repository in the Windows VM (changes that are required to connect to the Windows VM).


Willsparker:
I've added into this PR the additional options for `grep` so it works correctly when checking a Window's VM disk size to determine if resizing is necessary.

PR #987, opened 2019-11-07
WinVF: Fixed finding IP

Willsparker:
Currently the IP address is found through taking the 5th line from `Get-NetIPAddress | % {$_.IPAddress }`. This is usually fine, however occasionally, the IP Address on the machine is on the 4th line - resulting in Ansible attemping to connect to the wrong IP address 
e.g.: https://ci.adoptopenjdk.net/job/SXA-PlaybookCheckParallel/OS=Win2012,label=infra-vagrant-1/30/console

I've found that the IP Address will always start with `172` so I use that as a basis to find the correct IP from the output.


Willsparker:
The latest commit will find it, assuming that the first IPAddress listed that falls under `Ethernet*` is the one that's required to connect to the machine.

PR #988, opened 2019-11-08
UnixPB: Changed condition for yum autoremove in Clean_Up

Willsparker:
An update to a fix yesterday - I felt this was better as it just blacklists CentOS6, as opposed to whitelisting CentOS7, meaning it's more future proof if other versions of CentOS are added.


Willsparker:
@sxa555 Commited

PR #883, opened 2019-08-06
Changed Protobuf configure options

cwesMills:
Changed configure options for Protobuf to make Protobuf more optimized.

Signed-off-by: Colton Mills <millscolt3@gmail.com>

PR #989, opened 2019-11-08
Changed Protobuf configure options

jdekonin:
* changed configure options for Protobuf to make Protobuf more optimized
* replacing #883

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Protobuf/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Protobuf/tasks/main.yml
@@ -7,13 +7,67 @@
 #
 # Note: this role is optional
 # to apply the role, add "Protobuf" to the playbook roles (see ../../../main.yml)
+# this role also requires gcc7 role already applied to the system
+
+- name: Checking for /usr/bin/gcc-7
+  stat: path=/usr/bin/gcc-7
+  register: gcc7
+  tags: protobuf
+- name: Setting CC to /usr/bin/gcc-7 if found
+  set_fact:
+    CC: "/usr/bin/gcc-7"
+    CXX: "/usr/bin/g++-7"
+  when: gcc7.stat.exists == True
+  tags: protobuf
+
+- name: Checking for devtools-7 installation
+  stat: path=/opt/rh/devtoolset-7/root/usr/bin/gcc
+  register: gcc7
+  tags: protobuf
+- name: Setting CC to devtools-7 if found
+  set_fact:
+    CC: "/opt/rh/devtoolset-7/root/usr/bin/gcc"
+    CXX: "/opt/rh/devtoolset-7/root/usr/bin/g++"
+  when: gcc7.stat.exists == True
+  tags: protobuf
+
+- name: Checking for gcc-7.3 binary installation
+  stat: path=/usr/local/gcc/bin/gcc-7.3
+  register: gcc7
+  tags: protobuf
+- name: Setting CC to gcc-7.3 binary found
+  set_fact:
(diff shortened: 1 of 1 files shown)


jdekonin:
The is an inconsistency with functionality in what is generated when using gcc 5.5 or older; we used 7.3 because it is already on the system via playbook.

sxa555:
Do you have the details of the inconsistency (is there an issue?) and will it mean we have to redeploy protobuf on all machines? Also can you add a comment to that effect into the playbook as it might look confusing to someone in the future.

jdekonin:
I have no direct reference no.  But I believe the answer is that yes you would need to redeploy protobuf.   @dchopra001, can you please comment on why gcc7?

dchopra001:
I no longer have a specific example of the issue.

In my internal testing I ran into a lot of strange linking issues. One of the steps I took to resolve them at the time was to build protobuf with gcc-7 as well. It's _possible_ that some of those were related to the particular set of machines I was using and not the gcc level itself. Also, the logic to do this in those scenarios was a lot simpler.

I suppose we can remove the gcc-7 requirement here and try to build protobuf without gcc-7 and see if everything is working (we statically link the library as well, so I guess it _should_ be okay). If we experience issues then we'll know to move up to 7 again.

sxa555:
Thanks @dchopra001 that sounds like a good option for now to keep the playbooks simple. We haven't had any reported issues to my knowledge with the building and testing on the adoptopenjdk systems which will have been done with the default gcc on those systems. Tagging @smlambert in case she's seen anything

smlambert:
Thanks for the tag, I am just back from travelling, I will take a closer look at recent test runs to see if anything new has appeared.

dchopra001:
Actually @sxa555 I think it might be better if we decided to compile protobuf with gcc-7. Compiling with gcc-7 actually makes sense because protobufs are used to support JITServer in OpenJ9 and JITServer must be compiled with gcc-7 as well. I think it would be good if we kept this consistency as it can avoid problems now and into the future. Moreover, this is how we have been building on our own machines and inside Dockerfiles as well. It would be good if the build machines are setup to behave the same way. In the event that there are functional or performance issues in the future, this can be important detail.

Additionally, I recall us running into issues when protobuf was compiled with gcc 5.5 on some ubuntu 16 machines. Given the amount of issues we encountered and varying compiler levels among different machines, I think it'll be wiser if we keep the setup standard across all platforms by using gcc-7 to build on all of them.

sxa555:
@dchopra001 I'm still nervous as it will mean inconsistencies in installation depending on whether gcc7 has been installed on the machine or not. Where are the Dockerfiles that you're referring to for the JITaaS setup? I would be interested in seeing the details of any problems, since I believe the OpenJ9 machines are set up the same as the adoptopenjdk ones and so will likely be using a version compiled with GCC 4.8 up til now.

Also @jdekonin if we're going to do this we'll probably need to add a case for `/usr/local/gcc/bin/gcc-7.4` for Linux/s390x too :-)

dchopra001:
@sxa555 Here's the JITServer build dockerfile for x86:
https://github.com/eclipse/openj9/blob/b204b03c1861227451b05e187e3e8736ac7fa456/buildenv/docker/jdk8/x86_64/ubuntu16/jitserver/build/Dockerfile#L61-L65

jdekonin:
Although for PR and build testing jenkins uses these docker files.
https://github.com/eclipse/openj9/tree/b204b03c1861227451b05e187e3e8736ac7fa456/buildenv/jenkins/docker-slaves/

Additional check for 7.4 added.

dchopra001:
@mpirvu FYI

dchopra001:
@sxa555 Are you okay with approving this change now and letting the protobuf installation happen with gcc-7?

sxa555:
Yep

PR #992, opened 2019-11-13
pbTests: Addition of time messages for debug

Willsparker:
Outputs the time the playbook finishes. They're helpful for debugging the Jenkins Job (`SXA-PlaybookCheckParallel`) as theres been issues of the playbooks failing (due to timeout or some other reason) and knocking out Vagrant connections. They're also be helpful when we add the `--build` option to the job. 

Otherwise, it doesn't hurt to have them in when running `testScript.sh` locally.


ansible/pbTestScripts/testScript.sh
```diff
@@ -158,6 +158,7 @@ startVMPlaybook()
 	vagrant up
 	# Remotely moves to the correct directory in the VM and builds the playbook. Then logs the VM's output to a file, in a separate directory
 	vagrant ssh -c "cd /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook && sudo ansible-playbook --skip-tags adoptopenjdk,jenkins main.yml" 2>&1 | tee $WORKSPACE/adoptopenjdkPBTests/logFiles/$folderName.$branchName.$OS.log
+	echo "The playbook finished at : " && date | awk '{ print $5 }'
```
sxa555: Use `date +%T` instead of filtering with awk

sxa555: Actually the best way of writing this is:
```
echo The playbook finished at : `date +%T`
```

Issue #865, opened 2019-07-30
Set up machine and appropriate CI jobs to run the new playbook testing scripts


sxa555:
This will be putting any work from #842 and #864 into production kicked off via our CI tooling.
Will also require the installation of VM software (likely VirtualBox) onto one of our systems

Willsparker:
To keep a log of the work being done :
- There's a consistent error when running the playbooks in parallel, in that (it seems) when one playbook fails (due to an external reason), the other ones with stop instantly and say `Connection to 127.0.0.1 closed by remote host`. Hypothesis that this is connected either to `vagrant halt` or `vagrant destroy`.
- The Windows section of the `testScript` will find the wrong IP address occasionally, due to the the way I put finding it into the script. As of #990 I believe this should work, however there are now issues of  previous runs leaving files around.
- Certain roles in (mainly) the unix playbook timing out : `ant-contrib`,`cmake`,`ccache`,`Nvidia-Cuda`

Willsparker:
The `Connection to 127.0.0.1 closed by remote host` issue wasn't tied to either `vagrant halt` or `vagrant destroy`. It seems it is, however, tied to the multiple VMs running at once in some way - which is yet to be determined. A workaround for this has been to reduce the build executors on the Jenkins slave to 1, although this is not ideal as the job now takes ~8 hours to complete.
Another potential fix is to start the`testScript` in the Jenkins job to start at different times- This appeared to work in https://ci.adoptopenjdk.net/job/SXA-PlaybookCheckParallel/52/ wherein the failed OSs were due to timeout or unresponsive network issues.

The timeouts didn't occur when running a single playbook at once, either- though this may just be luck.

Willsparker:
<img width="382" alt="Screenshot 2019-11-15 at 13 22 19" src="https://user-images.githubusercontent.com/52783864/68946509-1157f300-07ab-11ea-99cd-1b093e815712.png">

ayy

karianna:
@Willsparker so this can be closed?

sxa555:
@karianna I think he's added that image onto the wrong issue as that one is for Dockerfile regular testing - https://ci.adoptopenjdk.net/view/Tooling/job/SXA-DockerfileCheck/

[...]

PR #999, opened 2019-11-18
pbTests: Updated VMDestroy

Willsparker:
Updates to make the script a bit more useful - this will destroy all of the VMs that have the OS that is passed to the script. This is more work towards #865 as a way of halting and destroying all VMs before all the `testScript.sh` s are ran, but not interfering with the playbook being executed.

This will still be useful if using `testScript.sh` locally, however would need to be used with more caution if there's VMs that are required to be retained.

### DIFF
--- a/ansible/pbTestScripts/vmDestroy.sh
+++ b/ansible/pbTestScripts/vmDestroy.sh
@@ -1,50 +1,58 @@
 #!/bin/bash
 set -eu
 
+osToDestroy=''
+
 # Takes in all arguments
 processArgs()
 {
-	if [ ! -n "${WORKSPACE:-}" ]; then
-		echo "WORKSPACE not found, setting it as environment variable 'HOME'"
-		WORKSPACE=$HOME
-	fi
 	if [ $# -lt 1 ]; then
 		echo "Script takes 1 input argument: "
-		echo "The project whose VMs are to be destroyed"
+		echo "The OS of the VMs you want to destroy."
 		exit 1
 	fi
-	
 }
 
-# Takes project name as arg 1, and OS as arg 2
-destroyVM()
-{
-	cd $WORKSPACE/adoptopenjdkPBTests/$1/ansible
-	ln -sf Vagrantfile.$2 Vagrantfile	# Correct Vagrantfile alias
-	vagrant destroy -f			# Force destroy without question
+checkOS() {
+	local OS=$1
+        case "$OS" in
+                "Ubuntu1604" | "U16" | "u16" )
+			osToDestroy="U16";;
+                "Ubuntu1804" | "U18" | "u18" )
+                        osToDestroy="U18";;
+                "CentOS6" | "centos6" | "C6" | "c6" )
+                        osToDestroy="C6" ;;
+                "CentOS7" | "centos7" | "C7" | "c7" )
+                        osToDestroy="C7" ;;
+                "Windows2012" | "Win2012" | "W12" | "w12" )
+                        osToDestroy="W2012";;
+                "all" )
+                        osToDestroy="U16 U18 C6 C7 W2012" ;;
+                *) echo "Not a currently supported OS" ; listOS; exit 1;
+        esac
+}
+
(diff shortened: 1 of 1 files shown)


Willsparker:
Thought I'd pre-emptively put in a confirmation check and `force` option.

ansible/pbTestScripts/vmDestroy.sh
```diff
@@ -1,50 +1,82 @@
...
-		destroyVM $1 $OS
-	done
-fi
+checkOS
+if [[ "$force" == False ]]; then
+	userInput=""
+	echo "Are you sure you want to destroy ALL Vms with the following OS(s)? (Y/n)"
+	echo "$osToDestroy"
+	read userInput
+	if [[ "$userInput" != "Y" ]]; then
```
sxa555: Should at the very least accept `y` too

ansible/pbTestScripts/vmDestroy.sh
```diff
@@ -1,50 +1,82 @@
...
-	if [ -d "$1" ]; then
-		echo "$1 found!"
-		return 0
-	else
-		echo "$1 not found"
-		exit 1
-	fi
+destroyVMs() {
+	local OS=$1
+	vagrant global-status | grep "adoptopenjdk$OS" | awk '{ print $1 }' | xargs -I {} vagrant destroy -f {}
```
sxa555: Why is `-I` and `{}` needed here?

ansible/pbTestScripts/vmDestroy.sh
```diff
@@ -1,50 +1,82 @@
...
 
-# Takes project name as arg 1, and OS as arg 2
-destroyVM()
-{
-	cd $WORKSPACE/adoptopenjdkPBTests/$1/ansible
-	ln -sf Vagrantfile.$2 Vagrantfile	# Correct Vagrantfile alias
-	vagrant destroy -f			# Force destroy without question
+usage() {
+
+		echo "Usage: ./vmDestroy <option>
```
sxa555: `vmDestroy.sh`

ansible/pbTestScripts/vmDestroy.sh
```diff
@@ -1,50 +1,82 @@
...
-	if [ -d "$1" ]; then
-		echo "$1 found!"
-		return 0
-	else
-		echo "$1 not found"
-		exit 1
-	fi
+destroyVMs() {
+	local OS=$1
+	vagrant global-status | grep "adoptopenjdk$OS" | awk '{ print $1 }' | xargs vagrant destroy -f
```
sxa555: Just to optimise it a bit use `vagrant global-status | awk "/adoptopenjdk$OS/ {print\$1 }"

ansible/pbTestScripts/vmDestroy.sh
```diff
@@ -1,50 +1,82 @@
...
+                        osToDestroy="U18";;
+                "CentOS6" | "centos6" | "C6" | "c6" )
+                        osToDestroy="C6" ;;
+                "CentOS7" | "centos7" | "C7" | "c7" )
+                        osToDestroy="C7" ;;
+                "Windows2012" | "Win2012" | "W12" | "w12" )
+                        osToDestroy="W2012";;
+                "all" )
+                        osToDestroy="U16 U18 C6 C7 W2012" ;;
+                *) echo "$OS is not a currently supported OS" ; listOS; exit 1;
```
sxa555: The error message when you get if you don't supply a parameter is not helpful as `$OS` is blank so this is confusing ... And also doesn't give any indication  that you need `-o` in front of any name from the list.

ansible/pbTestScripts/vmDestroy.sh
sxa555: Add `-o <os_list>` to this message, and indicate that the list is space separated underneath

sxa555: Also it's `./vmDestroy.sh`

Issue #781, opened 2019-05-03
Linux build requirements for ITW


charliegracie:
To complete the builds for ITW on Linux the following packages have to be installed on a CentOS6.9 machine (or RHEL6.9 I guess) that has basic build tools (autoconf, gcc, make, etc.).

There are a few extra requirements but the biggest bottlenecks are Rust and a newer version of pkg-config. With these installed the other requirements can be downloaded to the workspace and they are small so the downloads are very quick.

To install Rust:
```
curl -o rust.tar.gz "https://static.rust-lang.org/dist/rust-1.34.1-x86_64-unknown-linux-gnu.tar.gz"
tar -zxf rust.tar.gz
cd rust-1.34.1-x86_64-unknown-linux-gnu
./install.sh
```

To install a newer version of pkg-config:
```
curl -o pkg-config.tar.gz "https://pkg-config.freedesktop.org/releases/pkg-config-0.29.2.tar.gz"
tar -zxf pkg-config.tar.gz
cd pkg-config-0.29.2
./configure --with-internal-glib
make install
```

I will update with the other requirements in the future when I get the ansible playbook changes ready.

sxa555:
For now, I've installed them on [build-joyent-centos69-x64-1](https://ci.adoptopenjdk.net/computer/build-joyent-centos69-x64-1) which is now tagged with `icedteabuild` to be able to select it. You'll need to put `/usr/local/bin` explicitly at the start of the path to pick up the newer pkgconfig.

I'll leave this open until I've sorted the playbooks as well after we confirm that this is ok.

charliegracie:
Hmm it has the same pkg-config issue.  I think it is because the `pkg.m4` it is finding is the one from the original version of pkg-config and not the new version installed into `/usr/local/...`

charliegracie:
I am not sure how to force it to find the new definitions...

sxa555:
That sounds rather broken if it's not using it's `local` one first :-(

Do you have a pkg-config invocation that I can use to reproduce it?

jdekonin:
You need to setup a PKG_CONFIG_PATH environment variable.  Same problem exists on AIX.

https://askubuntu.com/questions/210210/pkg-config-path-environment-variable

charliegracie:
@jdekonin I thought that might be it too but I do not believe it is.  I tried setting it and it did not resolve the issue.

[...]

PR #1008, opened 2019-11-22
Playbook to setup machines to build Icedtea-web

Haroon-Khel:
Includes a new directory, within the ansible/playbooks directory, which stores the ansible roles and playbook required to configure machines to build Icedtea-web.

Includes the Common and GIT_Source roles from the AdoptOpenJDK_Unix_Playbook directory. These, in conjunction with the new pkg-config, rust and automake roles, are required to configure machines for an Icedtea-web build.

See issue #781


Haroon-Khel:
@sxa555 please review when you get the chance. Thanks

Haroon-Khel:
@sxa555 
Have only tested on Centos 7.6, so I will remove the other files for the other OSs in the Common role. I am looking to reduce the number of items in the Common role

Haroon-Khel:
@sxa555 The roles for the OSs, on which this playbook hasn't been tested, have been removed. And the Common role has been condensed to include fewer items, though i'm still in the process of reducing the number of items.

Haroon-Khel:
@sxa555 I have removed what I think to be all of the unnecessary items from the Common role. Let me know what you think.

Issue #994, opened 2019-11-14
Issue to log timeouts occurring when using adoptopenjdk playbooks


sxa555:
As part of the playbook testing we're doing under vagrant we're seeing various intermittent timeouts with external services e.g. cmake, sourceforget (for ant-contrib) and at least in once case today our own adoptopenjdk api. Due to the vagrant testing we're running the playbooks more frequently so it's more likely to show up intermittent issues This issue will log when these are occurring.

We can log them in comments saying somethign like this to include date and time to see if that has any bearing on the outcome:

```
Ubuntu1604    | ant-contrib     | 14/11 | 0805
```

Willsparker:
CentOS6:
  -  Ant Contrib: 
  ```
  TASK [Ant-Contrib : Download ant-contrib] **************************************
fatal: [localhost]: FAILED! => {"changed": false, "dest": "/tmp/", "gid": 0, "group": "root", "mode": "01777", "msg": "Request failed: <urlopen error _ssl.c:477: The handshake operation timed out>", "owner": "root", "secontext": "system_u:object_r:tmp_t:s0", "size": 4096, "state": "directory", "uid": 0, "url": "https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.tar.gz"}
	to retry, use: --limit @/vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/main.retry
  ```

Willsparker:
To Do: 
- [ ] Add Retries to the following:
  - [ ] Ant-Contrib
  - [ ] Git-Source
  - [ ] adoptopenjdk_install

sxa555:
I think we have mitigations in for all known issues now, so resolving this one...

PR #1011, opened 2019-11-25
UnixPB: Added retries to roles that often fail

Willsparker:
After #1000 , Decided it'd be best to add retries to other links that tend to fail, like the (still growing) list on #994.


Haroon-Khel:
@sxa555 Could you merge this? I do not have the authority

Willsparker:
@Haroon-Khel You can make them fail by changing them to an invalid URL - it should still retry - it will just fail 3 times (hopefully).

Haroon-Khel:
@Willsparker Can confirm, providing an invalid URL will make ansible retry 3 times before failing.

PR #1016, opened 2019-11-27
pbTests: alter testScript to run ansible locally and point at VM

Willsparker:
The change will make `testScript.sh` run the playbook on each Linux VM as it does with the Windows - runs ansible locally and points it at the VM.
This have been put in because currently, when running the jenkins job `SXA-PlaybookCheckParallel`, the use of `vagrant ssh -c` appears to be the cause of vagrant VMs being halted unintentionally. 
The `--build` and `--test` options have also been changed to not use `vagrant ssh -c`.

Has been tested on macOS and on the Jenkins job.

### DIFF
--- a/ansible/Vagrantfile.CentOS7
+++ b/ansible/Vagrantfile.CentOS7
@@ -23,6 +23,10 @@ echo "--------------------------------------------------------------------------
 echo "*** '$VAGRANT_MOUNT/playbooks/hosts' file for ansible-playbook contains ***"
 cat $VAGRANT_MOUNT/playbooks/hosts
 echo "---------------------------------------------------------------------------------------"
+# Get IPs of the VM, and output to shared folder
+ifconfig | grep "inet" | awk '{print $2}' | sed 's/addr://'  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
+# Put the host machine's IP into the authorised_keys file on the VM
+[ -r /vagrant/id_rsa.pub ] && mkdir -p $HOME/.ssh && cat /vagrant/id_rsa.pub >> $HOME/.ssh/authorized_keys
 SCRIPT
 
 # 2 = version of configuration file for Vagrant 1.1+ leading up to 2.0.x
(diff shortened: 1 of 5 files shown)


ansible/pbTestScripts/testScript.sh
```diff
@@ -78,6 +78,10 @@ checkVars()
                 echo "Can't find vagrant-disksize plugin, installing . . ."
                 vagrant plugin install vagrant-disksize
         fi
+	if [[ ! -f ~/.ssh/id_rsa ]]; then
```
sxa555: We probably shouldn't be generating a new key for the user who runs this. I suggest we create a new "one-time" key for each run of the script and store it in a suitable location (probably the shared directory) and use that instead.

ansible/Vagrantfile.CentOS7
```diff
@@ -23,6 +23,10 @@ echo "--------------------------------------------------------------------------
 echo "*** '$VAGRANT_MOUNT/playbooks/hosts' file for ansible-playbook contains ***"
 cat $VAGRANT_MOUNT/playbooks/hosts
 echo "---------------------------------------------------------------------------------------"
+# Get IPs of the VM, and output to shared folder
+ifconfig | grep "inet" | awk '{print $2}' | sed 's/addr://'  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
```
sxa555: Does this definitely work ok on all VMs? I would expect this would return at least two values - an external one and `127.0.0.1`

ansible/pbTestScripts/testScript.sh
```diff
@@ -162,17 +161,26 @@ startVMPlaybook()
 	else
 		cd $WORKSPACE/adoptopenjdkPBTests/$folderName-$branchName/ansible
 	fi
+
 	ln -sf Vagrantfile.$OS Vagrantfile
+	# Copy the machine's ssh key for the VMs to use, after removing prior files
+	[[ -f id_rsa.pub ]] && rm id_rsa.pub
+	cp ~/.ssh/id_rsa.pub $WORKSPACE/adoptopenjdkPBTests/$folderName-$branchName/ansible
```
sxa555: This is where we should generate the aforementioned ssh key

ansible/Vagrantfile.CentOS7
Willsparker: Indeed - I didn't think there'd be any harm in generating `hosts.tmp` even if the user wasn't starting the VM through `testScript.sh` too. And the way it picks the highest was lifted straight from the Windows section - you wrote it :-)

PR #1020, opened 2019-11-29
pbTests: Stops script if playbook fails; Cleans up after Windows VM

Willsparker:
This will stop the script if the playbooks don't succeed - as theres no point trying to build if the playbooks fail.
Due to this, the Windows section will clear up the files from the old run, just before doing a new one.


ansible/pbTestScripts/testScript.sh
```diff
@@ -194,6 +195,9 @@ startVMPlaybookWin()
 		cd $WORKSPACE/adoptopenjdkPBTests/$folderName-$branchName/ansible
 	fi
 	ln -sf Vagrantfile.$OS Vagrantfile
+	# Remove the Hosts files if they're found
+	[[ -f playbooks/AdoptOpenJDK_Windows_Playbook/hosts.tmp ]] && rm -f playbooks/AdoptOpenJDK_Windows_Playbook/hosts.tmp
```
sxa555: Get rid of the check on these two lines as it's not needed when using `rm -f`

Willsparker: Resolved with latest commit

Issue #1015, opened 2019-11-26
Debian 8 Playbook Issues


Willsparker:
Extension from #890 . Using this issue as a way of documenting the issues I have running the playbook on Debian 8, assuming `..AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml` is a copy of `..AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml`

Willsparker:
Adding both repositories `openjdk-r` and `ubuntu-toolchair-r/test` don't work as Ubuntu version "Jessie" doesn't exist.

Fix: add `update_cache = false` to both `apt-add-repository` tasks, and add another task under them that executes `sec -i 's/jessie/trusty/g' /etc/apt/sources.list.d/ *` ( except without the wildcard because Ansible doesn't like those )

Willsparker:
Installing `openjdk-8-jdk` is met with an error message stating:
```
The following packages have unmet dependencies:
 openjdk-8-jdk-headless : Depends: openjdk-8-jre-headless (= 8u222-b10-1~14.04) but it is not going to be installed
 openjdk-8-jre : Depends: openjdk-8-jre-headless (= 8u222-b10-1~14.04) but it is not going to be installed
                 Depends: libjpeg8 (>= 8c) but it is not installable
                 Depends: libpulse0 (>= 1:0.99.1) but 5.0-13 is to be installed
                 Recommends: fonts-dejavu-extra but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

This will need to be removed from `AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Debian.yml` and an alternative found.

Willsparker:
Nvidia_Cuda_Toolkit isn't installed on Debian 8 (nor Ubuntu18) due to the conditions. Simple addition of `or (ansible_distribution == "Debian" and ansible_distribution_major_version == "8")` to the end of each condition checks.

Willsparker:
The `local_srcinstall` role for `Protobuf` doesn't execute on Debian. Add `or ansible_distribution == "Debian"` to the conditionals.

sxa555:
https://developer.download.nvidia.com/compute/cuda/repos/ doesn't have any download for Debian so the addition of the extra conditionals does not help you locate something suitable for downloading on that platform later on in the role.

Willsparker:
[...]

PR #1021, opened 2019-11-29
UnixPB: Changes to allow for Debian8 playbook run through

Willsparker:
fixes: #1015 
@sxa555 @gdams 
These are all the changes I've had to implement to allow for a complete run through of the playbook. I was also able to build JDK8-openJ9, JDK11-openj9 and JDK13-HS when it ran through.
I've based the `Common/tasks/Debian.yml` and `Common/vars/Debian.yml` files on the `Ubuntu.yml` files of the respective repositories and changed them to suit:
  - Made sure `update_cache=no` is put in for the first 2 repositories, as the source files are required to be changed from `Jessie` to `Trusty`
  - Added the above task to change `Jessie` to `Trusty`
  - Removed all the assigning Java tasks
  - Removed openjdk-8-jdk from the "Build_Tool_Packages" *

In addition to this I have:
  - Changed the conditionals to include Debian in the OpenSSL102 roles, as system OpenSSL from the Vagrant Box is 1.0.1 which is slightly too old for OpenJ9.
  - Added Debian to the ProtoBuf localsrc_install conditionals.

* This is due to openjdk8 not being in the Jessie package repository ( https://packages.debian.org/jessie/java/ ).

The only thing in addition to this PR to make Debian work is the addition of someway of install JDK8, as it currently can't (see #1015 ). This could be remedied by adding an OpenJDK8 role to the Unix PB, but that's up for discussion.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
@@ -1,0 +1,151 @@
+---
+##########
+# Debian #
+##########
+
+#########################################
+# Configure Repos and Update the system #
+#########################################
+- name: Add the openjdk repository to apt
+  apt_repository: repo='ppa:openjdk-r/ppa' update_cache=no
+  when:
+    - ansible_architecture != "armv7l"
+  tags: patch_update
+
+- name: Add the ubuntu toolchain repository to apt
+  apt_repository: repo='ppa:ubuntu-toolchain-r/test' update_cache=no
+  when:
+    - ansible_architecture != "armv7l"
+  tags: patch_update
+
+- name: Change Jessie to Trusty in the sources files
+  command: sed -i "s/jessie/trusty/g" /etc/apt/sources.list.d/ppa_openjdk_r_ppa_jessie.list  /etc/apt/sources.list.d/ppa_ubuntu_toolchain_r_test_jessie.list
+  when:
+    - ansible_architecture != "armv7l"
+  tags: patch_update
+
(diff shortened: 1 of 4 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
```diff
@@ -0,0 +1,151 @@
+---
+##########
+# Ubuntu #
```
sxa555: No it isn't ;-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Debian.yml
```diff
@@ -0,0 +1,109 @@
+---
+##########
+# Ubuntu #
```
sxa555: Still not Ubuntu ;-)

sxa555:
Has this been tested on Raspbian which mostly identifies itself as Debian?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
Willsparker: Noted

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Debian.yml
Willsparker: Duly noted xD

Willsparker:
@sxa555 Yet to be tested on Raspbian- I'll do that tomorrow

PR #1022, opened 2019-11-29
pbTest: Remove $gitURL and use rm -f

Willsparker:


ansible/pbTestScripts/testScript.sh
```diff
@@ -160,7 +160,7 @@ startVMPlaybook()
 
 	ln -sf Vagrantfile.$OS Vagrantfile
 	# Copy the machine's ssh key for the VMs to use, after removing prior files
-	[[ -f id_rsa.pub ]] && rm id_rsa.pub id_rsa
+	[[ -f id_rsa.pub ]] && rm -f id_rsa.pub id_rsa
```
sxa555: Remvoe the check - not needed with `rm -f`

PR #1026, opened 2019-12-03
pbTest: Ensure in correct dir to start build; Add to grep to log file search; Refactor

Willsparker:
PR to change `testScript` to do several things:
 - Ensure that the inventory file can be picked up by the `ansible all` command when running a build on the Unix playbook, by being in the correct directory.
- Change the `searchLogFiles` function to `exit 1` when `unreachable=1` is detected.
- Refactor so `branchName` is set to `master` earlier on, so multiple `if` statements can be stripped out.

### DIFF
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -125,7 +125,7 @@ setupFiles()
 setupGit()
 {
 	cd $WORKSPACE/adoptopenjdkPBTests
-	if [ "$branchName" == "" ]; then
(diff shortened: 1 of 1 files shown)


Willsparker:
@sxa555 the `searchLogFiles` function `cd`s out of the current directory. The `if` statement, which I'll admit is not written very well, will set `branchName` to be master, if it's empty.

Willsparker:
Note: Since my last comment, I've decided to use this PR to clear up some other things in `testScript.sh`. See the added description.

PR #1030, opened 2019-12-06
pbTests: Refactor and clean-workspace option

Willsparker:
General refactor of the script to make it more readable and put the order of functions more like the order of execution. 

Also changed the `checkVagrantOS` function to detect which vagrant files are available, to then provision the VM and run the playbook against it. It future proofs it for when additional Vagrantfiles are added.

A `clean-workspace` option has been added as well. This will delete the old workspace if detected before git cloning a brand new one. Added this as it will stop any "Please stash your git changes before attempting to pull" messages that occur in the `SXA-PlaybookCheckParallel` job.

### DIFF
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -177,10 +177,10 @@ startVMPlaybook()
 	local pb_failed=$(searchLogFiles $OS)
 	if [[ "$testNativeBuild" = true && "$pb_failed" == 0 ]]; then
 		cd $WORKSPACE/adoptopenjdkPBTests/$folderName-$branchName/ansible
-		ansible all -i playbooks/AdoptOpenJDK_Unix_Playbook/hosts.unx -u vagrant -b -m raw -a "cd /vagrant/pbTestScripts && ./buildJDK.sh"
+		ansible all -i playbooks/AdoptOpenJDK_Unix_Playbook/hosts.unx -u vagrant -m raw -a "cd /vagrant/pbTestScripts && ./buildJDK.sh"
 		echo The build finished at : `date +%T`
 		if [[ "$runTest" = true ]]; then
-	        	ansible all -i playbooks/AdoptOpenJDK_Unix_Playbook/hosts.unx -u vagrant -b -m raw -a "cd /vagrant/pbTestScripts && ./testJDK.sh"
+	        	ansible all -i playbooks/AdoptOpenJDK_Unix_Playbook/hosts.unx -u vagrant -m raw -a "cd /vagrant/pbTestScripts && ./testJDK.sh"
 			echo The test finished at : `date +%T`
 		fi
 	fi
--- a/ansible/pbTestScripts/testScript.sh
+++ b/ansible/pbTestScripts/testScript.sh
@@ -9,6 +9,7 @@ retainVM=false
 testNativeBuild=false
 runTest=false
 vmHalt=true
+cleanWorkspace=false
 
 # Takes all arguments from the script, and determines options
 processArgs()
@@ -31,6 +32,8 @@ processArgs()
 				runTest=true;;
 			"--no-halt" | "-nh" )
 				vmHalt=false;;
+			"--clean-workspace" | "-c" )
+				cleanWorkspace=true;;
 			"--help" | "-h" )
 				usage; exit 0;;
 			*) echo >&2 "Invalid option: ${opt}"; echo "This option was unrecognised."; usage; exit 1;;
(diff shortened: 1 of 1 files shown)


ansible/pbTestScripts/testScript.sh
```diff
@@ -82,70 +85,71 @@ checkVars()
...
+	cd ${WORKSPACE}/adoptopenjdkPBTests/${folderName}-${branchName}/ansible
+	local vagrantOSlist=$(find Vagrantfile.* | cut -d. -f 2)
+	if [[ -f "Vagrantfile.${vagrantOS}" ]]; then
+		echo "Vagrantfile detected"
+	elif [[ "$vagrantOS" == "all" ]]; then
+		vagrantOS=$vagrantOSlist
+	else
+	        echo "Vagrantfile not found. These are the currently found Vagrantfiles:"
+	        echo $vagrantOSlist
+        	exit 1;
```
sxa555: Superfluous `;`

ansible/pbTestScripts/testScript.sh
```diff
@@ -209,88 +210,54 @@ startVMPlaybookWin()
...
+        local OS=$1
+        cd $WORKSPACE/adoptopenjdkPBTests/logFiles
+        echo
+        if grep -q 'failed=[1-9]\|unreachable=[1-9]' *$folderName.$branchName.$OS.log
+        then
+                return 1;
+        elif grep -q '\[ERROR\]' *$folderName.$branchName.$OS.log
+        then
+                return 1;
+        elif grep -q 'failed=0' *$folderName.$branchName.$OS.log
```
sxa555: If this is the only cast that can return zero (since the else clause returns `1`, why have all the extra checks?

ansible/pbTestScripts/testScript.sh
```diff
@@ -82,70 +85,71 @@ checkVars()
...
-		"CentOS7" | "centos7" | "C7" | "c7" )
-			vagrantOS="CentOS7" ;;
-		"Windows2012" | "Win2012" | "W12" | "w12" )
-			vagrantOS="Win2012";;
-		"all" )
-			vagrantOS="Ubuntu1604 Ubuntu1804 CentOS6 CentOS7 Windows2012" ;;
-		*) echo "Not a currently supported OS" ; vagrantOSList; exit 1;
-	esac
+	cd ${WORKSPACE}/adoptopenjdkPBTests/${folderName}-${branchName}/ansible
+	local vagrantOSlist=$(ls -l Vagrantfile.* | cut -d. -f 2)
```
sxa555: That needs to be `-1` not `-l` otherwise the list will be gibberish

Willsparker: In my testing, both output the same as `cut`s delimiter is `.` However, I do agree that `-1` would be better :-)

Willsparker:
Latest Commit will remove the `-b` option on the `ansible all` commands that build and run tests against the JDK- this is because we don't need to be root to build a JDK, and the test needs to be done in the `vagrant` user's workspace.

ansible/pbTestScripts/testScript.sh
sxa555: Not sure where you've done your testing - on a CentOS6 VM there's a `.` character after the permissions field at the start:
```
[vagrant@adoptopenjdkC6A vagrant]$ ls -l Vagrantfile.*
-rw-r--r--. 1 vagrant vagrant 1872 Dec 16 03:51 Vagrantfile.CentOS6
-rw-r--r--. 1 vagrant vagrant 1864 Dec  2 10:21 Vagrantfile.CentOS6~
-rw-r--r--. 1 vagrant vagrant 1864 Dec  2 10:21 Vagrantfile.CentOS7
-rw-r--r--. 1 vagrant vagrant 1963 Dec  2 10:21 Vagrantfile.Ubuntu1604
-rw-r--r--. 1 vagrant vagrant 1963 Dec  2 10:21 Vagrantfile.Ubuntu1804
-rw-r--r--. 1 vagrant vagrant 1925 Dec  2 10:21 Vagrantfile.Win2012
[vagrant@adoptopenjdkC6A vagrant]$ 
```

Willsparker: > Not sure where you've done your testing
Not on a CentOS6 VM ;-) This will only be executed on the Jenkins machine, which I believe will only ever be Ubuntu - and that doesn't have a dot. Output from the `infra-vagrant-1` machine : 
```
lrwxrwxrwx  1 jenkins jenkins   22 Dec  9 09:05 Vagrantfile -> Vagrantfile.Ubuntu1604
-rw-rw-r--  1 jenkins jenkins 2208 Dec  9 09:05 Vagrantfile.CentOS6
-rw-rw-r--  1 jenkins jenkins 2208 Dec  9 09:05 Vagrantfile.CentOS7
-rw-rw-r--  1 jenkins jenkins 2052 Dec  9 09:05 Vagrantfile.Debian8
-rw-rw-r--  1 jenkins jenkins 2307 Dec  9 09:05 Vagrantfile.Ubuntu1604
-rw-rw-r--  1 jenkins jenkins 2307 Dec  9 09:05 Vagrantfile.Ubuntu1804
-rw-rw-r--  1 jenkins jenkins 1965 Dec  9 09:05 Vagrantfile.Win2012
```
Regardless, it's been changed 👍

PR #1034, opened 2019-12-11
pbTests: Find correct JDK versions for different OSs

Willsparker:
As the build in `SXA-PlaybookCheckParallel` are JDK8-J9 builds, they need a higher GCC version to be set, so this PR ensures that.

Different OSs have different names for the OpenJDK-8 that's installed via the playbook. The change made should allow both Ubuntu and CentOS OSs in the job to find the correct JDK.
i.e, CentOS6 has `java-1.8.0-openjdk-1.8.0.232.b09-1.el6_10.x86_64`
and Ubuntu1604 has `java-1.8.0-openjdk-amd64`

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -6,13 +6,18 @@ if [ -z "${WORKSPACE:-}" ]; then
 		WORKSPACE=$HOME
 fi
 [[ ! -d "$WORKSPACE/openjdk-build" ]] && git clone https://github.com/adoptopenjdk/openjdk-build $WORKSPACE/openjdk-build
+
+# If detected GCC version is <7, set it to 7 
+export gccVer=$(gcc --version | grep "gcc" | awk '{ print $4 }' | cut -d. -f1)
+[[ ${gccVer} -lt 7 ]] && ln -sf /usr/bin/gcc-7 /usr/bin/gcc && ln -sf /usr/bin/g++-7 /usr/bin/g++
+
 export TARGET_OS=linux
 export ARCHITECTURE=x64
 export JAVA_TO_BUILD=jdk8u
 export VARIANT=openj9
 export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
-export JDK_BOOT_DIR=/usr/lib/jvm/java-1.8.0-openjdk-amd64
-export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
+# Differences in openJDK8 name between Ubuntu and CentOS
+export JAVA_HOME=$(find /usr/lib/jvm/ -name java-1.8.0-openjdk-\*)
 cd $WORKSPACE/openjdk-build
 build-farm/make-adopt-build-farm.sh
 
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -7,15 +7,16 @@ if [ -z "${WORKSPACE:-}" ]; then
 fi
 [[ ! -d "$WORKSPACE/openjdk-build" ]] && git clone https://github.com/adoptopenjdk/openjdk-build $WORKSPACE/openjdk-build
 
-# If detected GCC version is <7, set it to 7 
-export gccVer=$(gcc --version | grep "gcc" | awk '{ print $4 }' | cut -d. -f1)
-[[ ${gccVer} -lt 7 ]] && ln -sf /usr/bin/gcc-7 /usr/bin/gcc && ln -sf /usr/bin/g++-7 /usr/bin/g++
-
 export TARGET_OS=linux
 export ARCHITECTURE=x64
 export JAVA_TO_BUILD=jdk8u
 export VARIANT=openj9
-export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
+
+# Differences in openJDK7 name between OSs. Search for CentOS one
+export JDK7_BOOT_DIR=$(find /usr/lib/jvm/ -name java-1.7.0-openjdk.x86_64)
+# If the CentOS JDK7 can't be found, search for the Ubuntu one
+[[ -z "$JDK7_BOOT_DIR" ]] && export JDK7_BOOT_DIR=$(find /usr/lib/jvm/ -name java-1.7.0-openjdk-\*)
+
 # Differences in openJDK8 name between Ubuntu and CentOS
 export JAVA_HOME=$(find /usr/lib/jvm/ -name java-1.8.0-openjdk-\*)
 cd $WORKSPACE/openjdk-build


ansible/pbTestScripts/buildJDK.sh
```diff
@@ -6,13 +6,18 @@ if [ -z "${WORKSPACE:-}" ]; then
 		WORKSPACE=$HOME
 fi
 [[ ! -d "$WORKSPACE/openjdk-build" ]] && git clone https://github.com/adoptopenjdk/openjdk-build $WORKSPACE/openjdk-build
+
+# If detected GCC version is <7, set it to 7
```
sxa555: If this is required, something has gone wrong. The logic in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/platform-specific-configurations/linux.sh sets `CC` and `CXX` to the correct value - hacking symlinks outside the playbook shouldn't be required.

ansible/pbTestScripts/buildJDK.sh
```diff
@@ -6,13 +6,18 @@ if [ -z "${WORKSPACE:-}" ]; then
...
+
+# If detected GCC version is <7, set it to 7 
+export gccVer=$(gcc --version | grep "gcc" | awk '{ print $4 }' | cut -d. -f1)
+[[ ${gccVer} -lt 7 ]] && ln -sf /usr/bin/gcc-7 /usr/bin/gcc && ln -sf /usr/bin/g++-7 /usr/bin/g++
+
 export TARGET_OS=linux
 export ARCHITECTURE=x64
 export JAVA_TO_BUILD=jdk8u
 export VARIANT=openj9
 export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
```
sxa555: Are the packages with the OS really different in their naming formats between java 7 and 8?

ansible/pbTestScripts/buildJDK.sh
Willsparker: Looking at the logic of that - it only sets the correct value of `CC` and `CXX` when it's building JDK11+-J9, or when it's `s390x` and below JDK11-J9, it will set gcc to 4.8.

ansible/pbTestScripts/buildJDK.sh
Willsparker: Looking at C6, U16 and U18, it seems they also have different names for Java 7:
C6: `java-1.7.0 `**or** `java-1.7.0-openjdk-1.7.0.241.x86_64`
U16: `java-1.7.0-openjdk-amd64` **or** `java-7-openjdk-amd64`
U18: No Java 7 installed. Probably due to `openjdk-7-jdk` only being under the Ubuntu1604 list of extra packages to install: https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/29ce5a65c94510fd91dcd29c42dd07ffcb3ff006/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Ubuntu.yml#L72

ansible/pbTestScripts/buildJDK.sh
sxa555: Incorrect - the `s390x` at the start is overriden by this section:
`if [ "$JAVA_FEATURE_VERSION" -gt 10 ] || [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]; then`
which will take effect on all OpenJ9 versions.

ansible/pbTestScripts/buildJDK.sh
```diff
@@ -15,7 +15,7 @@ export TARGET_OS=linux
 export ARCHITECTURE=x64
 export JAVA_TO_BUILD=jdk8u
 export VARIANT=openj9
-export JDK7_BOOT_DIR=/usr/lib/jvm/java-1.7.0
+export JDK7_BOOT_DIR=$(find /usr/lib/jvm/ -name java-1.7.0-\*)
```
sxa555: On my C6 system that would pick up at least two entries which would make a mess of the `JDK7_BOOT_DIR` variable (since there is one symlink to something else,. both matching that pattern)

```
[root@159 jvm]# ls -ld java-1.7.0-*
drwxr-xr-x 3 root root 4096 May  3  2018 java-1.7.0-openjdk-1.7.0.171.x86_64
drwxr-xr-x 3 root root 4096 Aug 10  2018 java-1.7.0-openjdk-1.7.0.181.x86_64
drwxr-xr-x 2 root root 4096 Nov 21  2018 java-1.7.0-openjdk-1.7.0.191.x86_64
drwxr-xr-x 2 root root 4096 Apr 23  2019 java-1.7.0-openjdk-1.7.0.211.x86_64
drwxr-xr-x 7 root root 4096 Oct 24 01:00 java-1.7.0-openjdk-1.7.0.241.x86_64
lrwxrwxrwx 1 root root   35 Oct 24 01:00 java-1.7.0-openjdk.x86_64 -> java-1.7.0-openjdk-1.7.0.241.x86_64
```

ansible/pbTestScripts/buildJDK.sh
sxa555: If they all have a "well-defined" symlink name, then perhaps test for each possibility (i.e. two of them - `java-1.7.0` or `java-1.7.0-openjdk-amd64`) and set it accordingly?

ansible/pbTestScripts/buildJDK.sh
Willsparker: Ah my mistake.
I've just had a look at an U16 machine that had the playbook run on it, and gcc7 isn't installed to `/usr/local/gcc`. CentOS6 does however have gcc-7.3 installed to `/opt/rh/devtoolset-7/root/usr/bin` as the script you linked is looking for. 
On Ubuntu machines, it installs it to `/usr/bin/x86_64-linux-gnu-gcc-7`

sxa555: U16 should also by symlinked to `/usr/bin/gcc-7` (Which is used in the guard to stop https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_7/tasks/main.yml from installing it into /usr/local but as you say not in the environment file above. Have created https://github.com/AdoptOpenJDK/openjdk-build/pull/1446/files to resolve this - feel free to review that for me ;-)

ansible/pbTestScripts/buildJDK.sh
Willsparker: New commit in that will look for the CentOS symlink first, and then look for the Ubuntu symlink after. If it can't find it, then the variable remains blank and it will default to building JDK8 with JDK8 - for instance in the case of Ubuntu1804

ansible/pbTestScripts/buildJDK.sh
Willsparker: Reviewed and merged the above PR, therefore removed this block of code.

PR #1036, opened 2019-12-11
UnixPB: Changes so PB runs on Raspbian Buster

Willsparker:
Extension of #1021 

Changes to allow the playbook to run through on Raspbian Buster. I noticed there was an override to cause Raspbian to run the Ubuntu version of the playbook. I deleted this as there is now a Debian 
specific playbook. 

The only additions were adding a couple of repositories to allow for the playbook to install gcc-4.8 on the RPI - This is not available to Raspbian Buster by default. I also had to add the Raspbian specific repository and gpg key for Docker, as the Debian repository relies on the `auf-dkms` package that is not available on Raspbian (ref: https://github.com/raspberrypi/linux/issues/3021).

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
@@ -38,6 +38,16 @@
     - ansible_architecture == "x86_64"
   tags: patch_update
 
+# Both repositories needed for gcc/g++-4.8
+- name: Add additional repositories for Raspbian Buster
+  apt-repository: repo={{ item }}
+  with_items:
+    - deb-src http://raspbian.raspberrypi.org/raspbian/ buster main contrib non-free rpi
+    - deb http://mirrordirector.raspbian.org/raspbian/ jessie main contrib non-free rpi
+  when:
+    - (ansible_distribution_major_version == "10" and ansible_architecture == "armv7l")
+  tags: patch_update
+
 - name: Add additional repositories for BeagleBone
   apt_repository: repo={{ item }}
   with_items:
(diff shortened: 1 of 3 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
```diff
@@ -232,6 +241,16 @@
     - ansible_architecture == "x86_64"
   tags: docker
 
+- name: Add Docker repo for Raspbian arm7vl
+  apt_repository:
+    repo: "deb https://download.docker.com/linux/raspbian {{ ansible_distribution_release }} stable"
+    state: present
+  when:
+    - docker_installed.rc != 0
+    - ansible_distribution == "Debain"
```
sxa555: Typo!

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
```diff
@@ -38,6 +38,15 @@
     - ansible_architecture == "x86_64"
   tags: patch_update
 
+- name: Add additional repositories for Raspian Buster
+  apt-repository: repo={{ item }}
+  with_items:
+    - deb-src http://raspbian.raspberrypi.org/raspbian/ buster main contrib non-free rpi
```
sxa555: What's the `deb-src` line needed for in this case?

Willsparker: That will allow installation of `gcc-4.8`. On a fresh install of Raspbian Buster, it's already in `/etc/apt/sources.list` but commented out. See https://raspberrypi.stackexchange.com/a/101115

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
Willsparker: Fixed!

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
sxa555: I'm slightly tempted to just leave it out and just build on the default on that system, but can you add a comment saying that they're needed for gcc/g++ 4.8 (Also `s/Raspian/Raspbian`) :-)

Willsparker: Commit in :-)

PR #1039, opened 2019-12-12
winPB: Change Git version to latest; use win_shell module

Willsparker:
There was an error with downloading GIt on the windows playbook : 
```
TASK [GIT : Install GIT] *******************************************************
fatal: [172.28.128.4]: FAILED! => {"changed": false, "cmd": "C:\\temp\\git.exe /SILENT /COMPONENTS=\"icons,ext\\reg\\shellhere,assoc,assoc_sh\"", "msg": "Exception calling \"CreateProcess\" with \"5\" argument(s): \"CreateProcessW() failed (The file or directory is corrupted and unreadable, Win32ErrorCode 1392)\"", "rc": 2}
```
The downloading task before it seemed to incorrectly downloading the git executable, as the size was much smaller than what it should've been. Narrowed it down to the `win_get_url` module being the issue. 
The addition of the line `[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12` is to set the tls protocol to version 1.2, as thats what the webpage requires when using `wget`.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/GIT/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/GIT/tasks/main.yml
@@ -16,9 +16,7 @@
   tags: git
 
 - name: Download GIT installer
-  win_get_url:
-    url: 'https://github.com/git-for-windows/git/releases/download/v2.14.3.windows.1/Git-2.14.3-64-bit.exe'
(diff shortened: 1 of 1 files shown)


Willsparker:
@sxa555 I'm inclined to agree- it's causing issues with other roles later on as well. However, it works perfectly fine in the `Strawberry_Perl` role, which occurs before the `GIT` task. Not sure about if the corrupt size was consistent, but it was significantly smaller than it should've been.

PR #1040, opened 2019-12-12
pbTests: Update testJDK.sh

Willsparker:
Updated `testJDK.sh` to work now that builds are semi reliable (once #1034 is merged).

Tested on built JDKs in Ubuntu1604. Ubuntu1804 and CentOS6.

### DIFF
--- a/ansible/pbTestScripts/testJDK.sh
+++ b/ansible/pbTestScripts/testJDK.sh
@@ -1,14 +1,12 @@
 #!/bin/bash
 
-mv /vagrant/pbTestScripts/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8* ~
-export TEST_JDK_HOME=$(find ~ -maxdepth 1 -type d -name "*jdk8u*"|grep -v ".*jre.*")
-cd $HOME && mkdir -p testLocation
-cd testLocation && git clone https://github.com/adoptopenjdk/openjdk-tests
-cd openjdk-tests
-./get.sh -t $HOME/testLocation/openjdk-tests
-cd TestConfig
+mv $HOME/openjdk-build/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8* $HOME
+export TEST_JDK_HOME=$(find $HOME -maxdepth 1 -type d -name "*jdk8u*"|grep -v ".*jre.*")
+mkdir -p $HOME/testLocation
+[ ! -d $HOME/testLocation/openjdk-tests ] && git clone https://github.com/adoptopenjdk/openjdk-tests $HOME/testLocation/openjdk-tests
+$HOME/testLocation/openjdk-tests/get.sh -t $HOME/testLocation/openjdk-tests
+cd $HOME/testLocation/openjdk-tests/TKG || exit 1
 make -f run_configure.mk
 export BUILD_LIST=MachineInfo
 make compile
 make _MachineInfo
-


ansible/pbTestScripts/testJDK.sh
```diff
@@ -1,14 +1,11 @@
...
-cd $HOME && mkdir -p testLocation
-cd testLocation && git clone https://github.com/adoptopenjdk/openjdk-tests
-cd openjdk-tests
-./get.sh -t $HOME/testLocation/openjdk-tests
-cd TestConfig
-make -f run_configure.mk
+mkdir -p $HOME/testLocation
+[ ! -d $HOME/testLocation/openjdk-tests ] && git clone https://github.com/adoptopenjdk/openjdk-tests $HOME/testLocation/openjdk-tests
+$HOME/testLocation/openjdk-tests/get.sh -t $HOME/testLocation/openjdk-tests
+cd $HOME/testLocation/openjdk-tests/TKG && make -f run_configure.mk
```
sxa555: We probably want to explicitly fail if the `cd` fails, not just skip this one command so
```
cd $HOME/testLocation/openjdk-tests/TKG || exit 1
make -f run_configure.mk
```
Without this it'll still try to run `make compile` and `make _MachineInfo` from the wrong directory if the clone fails for example

Willsparker: Good point - Changes have been made :-)

ansible/pbTestScripts/testJDK.sh
```diff
@@ -1,14 +1,12 @@
 #!/bin/bash
 
-mv /vagrant/pbTestScripts/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8* ~
+mv $HOME/openjdk-build/workspace/build/src/build/linux-x86_64-normal-server-release/images/jdk8* $HOME
 export TEST_JDK_HOME=$(find ~ -maxdepth 1 -type d -name "*jdk8u*"|grep -v ".*jre.*")
```
sxa555: You seem to have changed the previous line from `~` to `$HOME` but not this one - any particular reason?
Also is there a reason not to just set `TEST_JDK_HOME` to the original location instead of moving it?

ansible/pbTestScripts/testJDK.sh
```diff
@@ -1,14 +1,12 @@
...
-cd $HOME && mkdir -p testLocation
-cd testLocation && git clone https://github.com/adoptopenjdk/openjdk-tests
-cd openjdk-tests
-./get.sh -t $HOME/testLocation/openjdk-tests
-cd TestConfig
+mkdir -p $HOME/testLocation
```
sxa555: Any reason not to use `$WORKSPACE` if available? Having jenkins jobs messing around in the user's home directory is not ideal

Willsparker: This will be executed on the VMs that are built on the machine. So `$WORKSPACE` will never be set, and messing around in the user's home directory won't affect the machine running the job :-)

ansible/pbTestScripts/testJDK.sh
Willsparker: Ah, Sorry, missed that.
 As for setting TEST_JDK_HOME, I initially moved it as it was easier to move both the JDK and the JRE, then set just the JDK - see : https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/943#discussion_r326642706

ansible/pbTestScripts/testJDK.sh
sxa555: Fair point

PR #1042, opened 2019-12-13
Add Linux perf machines

gdams:


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
```diff
@@ -64,6 +64,6 @@
   when:
     - ansible_architecture == "x86_64"
     - ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
-    - ansible_distribution_major_version == "6"
+    - ansible_distribution_major_version == "6" or ansible_distribution_major_version == "7"
```
sxa555: Is this definitely needed on 7? If so, why would it have been locked to 6 in the first place?

gdams: Looking at the original PR it was only tested on 6. This is also only needed on perf machines so may have not been spotted on previous test machines.

sxa555: Where is the documented requirement for that on RHEL7?

sxa555: @smlambert Can you comment on this? core hard limit is unlimited on RHEL7 by default I believe, open files `4096` and process 7260. I can see that some of the performance tests might just about need more than 4096 files open but I'd have thought the other limits wouldn't be a problem with the defaults for performance machines.

karianna:
@gdams - needs a rebase

gdams:
@sxa can we merge this now?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
sxa: @smlambert Any view on my previous comment?

smlambert: We did hit an issue with file handle limits for some tests.  I do not recall hitting any other resource issues though.

sxa: > @sxa can we merge this now?

Do we have the access control on this machine set up, or is there a reason not to? I don't want to merge until the setup of the machines is finalised since they will otherwise be difficult to support.

sxa:
> @sxa can we merge this now?

As per my earlier comment is the access control sorted? As far as I can see I can't log into it.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
gdams: @sxa I've added your key to both machines. The KeyBox key is also on the machines

gdams:
@sxa your key has been added, please can this be merged now

ansible/inventory.yml
```diff
@@ -79,6 +79,10 @@ hosts:
           macos1015-x64-1: {ip: 208.83.1.221, user: Administrator}
           macos1015-x64-2: {ip: 208.83.1.117, user: Administrator}
 
+      - softlayer:
```
sxa, line 82:
```suggestion
      - ibmcloud:
```

sxa (changes requested):
Thanks - just catching up on apparently a month worth of github notifications I hadn't got to.

I've recently changed `softlayer` to `ibmcloud` under https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1632 - I think we should probably change this one too (since it's now out of date with https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/60c0bdc15a6cfbe7d0fffc00e8b45a563e7b1ccd/ansible/plugins/inventory/adoptopenjdk_yaml.py#L48

The only reason I say "probably" is because these machines are charged to us differently so that might be a reason to keep them separate. But for basic simplicity/clarity I'd change these to `ibmcloud` too.

gdams:
@sxa updated PTAL

sxa (approved):
LGTM :-D

PR #1045, opened 2019-12-16
Resolve the dependency error for RedHat s390x

JennyChenyj:
- If the 32 bit C lib is installed on machines,
 it will be skipped.

Signed-off-by: Jenny Chen <Jenny.Chen@ibm.com>

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
@@ -73,7 +73,7 @@
   tags: build_tools
 
 - name: Install additional build tools for RHEL on s390x
-  package: "name={{ item }} state=latest"
+  package: "name={{ item }} state=present"
   with_items: "{{ Additional_Build_Tools_RHEL_s390x }}"
   when:
     - ansible_architecture == "s390x"


sxa555:
I'm not clear how this provides a solution to the problem. Surely if the issue is related to multiarch then a switch from latest to present won't make a difference - it looks to be objecting to the initial installation of the package unless I'm mistaken.

The machine that showed the error does not have the 31-bit packages installed, so a fix for when it's already installed doesn't seem like the right answer ...

```
[linux1@localhost ~]$ rpm -qa | grep glibc
glibc-common-2.17-292.el7.s390x
glibc-2.17-292.el7.s390x
glibc-headers-2.17-292.el7.s390x
glibc-devel-2.17-292.el7.s390x
```

@Haroon-Khel Can you verify this patch on a new clean machine internally please?

JennyChenyj:
Thanks for explaining the problem! I think the problem might cause by multiple versions of nss-softokn-freebl. Could we check using this command? 'yum list nss-softokn-freebl' @sxa555 @Haroon-Khel

JennyChenyj:
I also tried to run the playbook on a new machine that has s390x installed. The multilib versions problem did not occur. 
Screen Shot 2019-12-17 at 11 28 15 AM
Thanks for checking it again!

Haroon-Khel:
@JennyChenyj I ran the playbook on a new RHEL 7 s390x machine, which started out with the s390x versions of the packages installed. The installation of the s390 versions of the packages caused no error. The machine also has both s390 and s390x versions of `nss-softokn-freebl` 

`yum list nss-softokn-freebl`:
Screenshot 2019-12-17 at 17 06 45

JennyChenyj:
> @JennyChenyj I ran the playbook on a new RHEL 7 s390x machine, which started out with the s390x versions of the packages installed. The installation of the s390 versions of the packages caused no error. The machine also has both s390 and s390x versions of `nss-softokn-freebl`
> 
> `yum list nss-softokn-freebl`:
> Screenshot 2019-12-17 at 17 06 45

Thanks for checking it!

JennyChenyj:
I also tried the original playbook with the state ```latest``` on a new machine. It also caused no error. Do we want this pull request? @Haroon-Khel @sxa555

sxa555:
@Haroon-Khel Does that mean that if we sequence the playbook to ensure that the s390x versions are installed first then it's ok? (Presumably at the moment it's the other way round?)

Haroon-Khel:
@sxa555 The s390x versions are already installed first, then the s390 versions. Im beginning to think that this error might have been specific to our particular machine.

JennyChenyj:
If so, I will close this pull request for now. Thanks!

PR #1050, opened 2019-12-18
add additional aarch64 test machines

gdams:
closes: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1037

FYI @smlambert

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -102,11 +102,13 @@ hosts:
           centos7-x64-3: {ip: 160.153.246.194, user: adoptopenjdk}
           centos7-x64-4: {ip: 160.153.244.42, user: adoptopenjdk}
 
-      - openlab:
-          ubuntu1604-armv8-1: {ip: 199.204.45.100, user: ubuntu}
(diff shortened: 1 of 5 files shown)


smlambert:
To confirm, yes, xvfb is currently how we enable certain openjdk jdk_awt and jdk_swing tests, that being said, the Jenkins Xvfb plugin does not seem to work for OS X, so we have disabled those tests in some cases (see https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/playlist.xml#L155 and https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/playlist.xml#L415

If we had to, we could do same on aarch64 (not ideal though).

smlambert:
I don't have a strong opinion for the other parts of the PR.  I would not hold up a merge of this but rather come in with follow-up PR, if need be. 

How are these playbook changes tested?  I can run some tests on a machine to confirm if other testing looks good.

gdams:
@sxa555 I've added the centos deps to the playbook for now as a workaround. Are you happy to merge this and then we can work on a more permanent solution?

Issue #998, opened 2019-11-16
Add checksum option to all ansible get*url operations


sxa555:
We should have everything checksummed to protect us against corrupt/malicious downloads etc.

PR #1051, opened 2019-12-18
UnixPB: Add checksums to all `get_url` modules

Willsparker:
Fixes: #998 
This will be broken up into a few groups:
 - The 1st group of checksums commit entail all the roles with `get_url` tasks that have provided a checksum with the download. I.e. `cmake`s one is from https://cmake.org/files/v3.11/cmake-3.11.4-SHA-256.txt

- The 2nd group will entail all roles that don't provide a checksum. These will be found by executing `wget` on each URL and installing them onto a VM. If the installation works correctly, I'll assume the downloaded item was correct and use that checksum.


Willsparker:
Taking out of [WIP] as I tested the changes made on my branch here: https://ci.adoptopenjdk.net/job/SXA-PlaybookCheckParallel/213/ 
All relevant changes haven't broken the playbooks.

Issue #1033, opened 2019-12-10
Failing SLES 12 tasks in Unix playbook


Haroon-Khel:
Currently the unix playbook fails while running on an x86_64 SLES 12 machine on our intranet. The failing tasks are:

- The `zypper_repository` task in `roles/Common/tasks/SLES.yml`. The repo at `https://download.opensuse.org/repositories/devel:/tools:/scm/SLE_12_SP3/` is no longer available.

- A temporary solution is to change the repo to `https://download.opensuse.org/repositories/devel:/tools:/scm/SLE_12_SP5/`, but this results in failures when trying to install `systemtap-sdt-devel` and `git`, in `roles/Common/vars/SLES.yml`

- The task `Install XML::Parser` in `roles/CPAN`

- The tasks `Enable NVidia CUDA toolkit Repo for SLES12 on x86_64` and `Install NVidia CUDA toolkit for SLES12 on x86_64` in `roles/NVidia_Cuda_Toolkit`

- `Mount swap file` in `roles/Swap_File`

Willsparker:
Looking into it :)

Willsparker:
The above link to the SP5 repo doesn't contain `systemtap-std-dev` and to install git, `git-core` need to be used. However, https://download.opensuse.org/repositories/devel:/tools/SLE_12_SP4/ has both :-)

Willsparker:
Regarding the `XML::Parser` issue, the machine requires `expat-devel` or `libexpat1-dev` installed, of which I can't find a way of downloading, either through `wget` or adding a repository and using `zypper`. The alternative is installing `perl-XML-parser` in the `Common` role, according to https://github.com/chorny/XML-Parser/README.md, though I'm not sure how to verify this.

Willsparker:
The `NVidia_Cuda_Toolkit` task was failing as it was trying to `rpm -i ` all words after the actual package. Not entirely sure why, but separating the `rpm` and `sed` command fixes the issue.

Willsparker:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1062#issuecomment-570226824

PR #1062, opened 2019-12-23
UnixPB: Update playbook to allow SLES12 to work

Willsparker:
fixes: #1033 

Fixes all of the issues listed in 1033. 

The only concern I have is the changes implemented to not install CPAN's XML::Parser, and instead install `perl-xml-parser` under the `test-tool-packages` section. Supposedly that's meant to be equivalent, according to this : https://github.com/chorny/XML-Parser/blob/master/README

However, I don't know how to verify this. It does install however.
I've done this as CPAN XML::Parser has a dependency for `libexpat1-dev` (or `libexpat-dev`) or `expat-dev` depending on the OS, that appears to not be available to SLES12-SP4 (The machine both @Haroon-Khel and I were testing on).

Note: Currently testing using the `vagrantPlaybookCheck.sh` script. https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/220/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/CPAN/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/CPAN/tasks/main.yml
@@ -47,6 +47,7 @@
 
 - name: Install XML::Parser
   command: cpanm --with-recommends --sudo XML::Parser
+  when: ansible_distribution != "SLES" and ansible_distribution_major_version != "12"
   tags:
     - cpan
     - xml_parser
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
@@ -91,9 +91,7 @@
   tags: nvidia_cuda_toolkit
 
 - name: Enable NVidia CUDA toolkit Repo for SLES12 on x86_64
-  command: |
-    rpm -i /tmp/sles12_cuda9_repo.rpm
-    sed 's/gpgcheck=1/gpgcheck=0/' -i /etc/zypp/repos.d/cuda.repo
+  command: rpm -i /tmp/sles12_cuda9_repo.rpm
   when:
     - cuda_installed.stat.islnk is not defined
     - ansible_architecture == "x86_64"
@@ -103,6 +101,19 @@
     - nvidia_cuda_toolkit
     #TODO: rpm used in place of yum or rpm_key module
     - skip_ansible_lint
+
+- name: Sed change gpgcheck for SLES12 on x86_64
+  command: sed 's/gpgcheck=1/gpgcheck=0/' -i /etc/zypp/repos.d/cuda.repo
+  when:
+    - cuda_installed.stat.islnk is not defined
+    - ansible_architecture == "x86_64"
+    - ansible_distribution == "SLES"
+    - ansible_distribution_major_version == "12"
+  tags:
+    - nvidia_cuda_toolkit
+    #TODO: rpm used in place of yum or rpm_key module
+    - skip_ansible_lint
+
 
 - name: Install NVidia CUDA toolkit for SLES12 on x86_64
   zypper: pkg=cuda state=latest update_cache=yes
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/SLES.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/SLES.yml
@@ -8,7 +8,7 @@
 #########################################
 - zypper_repository:
     name: devel-tools
-    repo: 'https://download.opensuse.org/repositories/devel:/tools:/scm/SLE_12_SP3/'
+    repo: 'https://download.opensuse.org/repositories/devel:/tools/SLE_12_SP4/'
     auto_import_keys: yes
     state: present
   when:
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
@@ -27,7 +27,7 @@ Build_Tool_Packages:
 
 Additional_Build_Tools_SLES12:
   - java-1_8_0-openjdk
-  - git
+  - git-core
   - libfreetype6
   - libXext6
   - libXi6                        # JDK12+ compilation
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Swap_File/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Swap_File/tasks/main.yml
@@ -9,22 +9,12 @@
 # Not installed: When swap_root.rc or swap_tmp.rc does NOT equal 0
 # Assume swap path of /swapfile override where applicable
 
-- name: Test swapfile path - /swapfile
-  command: ls /swapfile
+- name: Test if swapfile exists
+  command: grep swap /etc/fstab
   ignore_errors: yes
-  register: swap_root
+  register: swap_exists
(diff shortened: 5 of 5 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NVidia_Cuda_Toolkit/tasks/main.yml
```diff
@@ -104,6 +102,19 @@
     #TODO: rpm used in place of yum or rpm_key module
     - skip_ansible_lint
 
+- name: Sed change gpgcheck for SLES12 on x86_64
```
sxa555: Why is this needed - if we have to do this instead of doing a proper GPG check then it needs a comment explaining why

Willsparker: I'm not sure- It was part of the `Enable NVidia_Cuda_Toolkit for SLES x86_64` tasks, I just split it up as `rpm` was reading in every word as a package to install.

sxa555: Can you do some experimentation and see if it's still needed and if so what's the message - we should document it in this playbook

Willsparker:
@sxa555 The sed task does appear to be necessary. When not running it the `install NVidia_Cuda_Toolkit` task outputs the following: 

`fatal: [9.20.195.225]: FAILED! => {"changed": false, "cmd": ["/usr/bin/zypper", "--quiet", "--non-interactive", "--xmlout", "refresh"], "msg": "Zypper run command failed with return code 4.", "rc": 4, "stderr": "", "stderr_lines": [], "stdout": "<?xml version='1.0'?>\n<stream>\n<prompt id=\"14\">\n<text>\nNew repository or package signing key received:\n\n&lt;gpgkey-info&gt;&lt;repository&gt;cuda&lt;/repository&gt;&lt;key-name&gt;cudatools &amp;lt;cudatools@nvidia.com&amp;gt;&lt;/key-name&gt;&lt;key-fingerprint&gt;AE09FE4BBD223A84B2CCFCE3F60F4B3D7FA2AF80&lt;/key-fingerprint&gt;&lt;key-created time_t=&quot;1466808197&quot; text=&quot;2016-06-24T15:43:17-07&quot;/&gt;&lt;key-expires time_t=&quot;0&quot; text=&quot;0&quot;/&gt;&lt;rpm-name&gt;gpg-pubkey-7fa2af80-576db785&lt;/rpm-name&gt;&lt;/gpgkey-info&gt;\n\nDo you want to reject the key, trust temporarily, or trust always?</text>\n<option default=\"1\" value=\"r\" desc=\"Don&apos;t trust the key.\"/>\n<option value=\"t\" desc=\"Trust the key temporarily.\"/>\n<option value=\"a\" desc=\"Trust the key and import it into trusted keyring.\"/>\n</prompt>\n<message type=\"error\">Repository &apos;cuda&apos; is invalid.\n[cuda|http://developer.download.nvidia.com/compute/cuda/repos/sles122/x86_64] Valid metadata not found at specified URL\nPlease check if the URIs defined for this repository are pointing to a valid repository.\n</message>\n<message type=\"error\">Skipping repository &apos;cuda&apos; because of the above error.</message>\n<message type=\"error\">Some of the repositories have not been refreshed because of an error.</message>\n</stream>\n", "stdout_lines": ["<?xml version='1.0'?>", "<stream>", "<prompt id=\"14\">", "<text>", "New repository or package signing key received:", "", "&lt;gpgkey-info&gt;&lt;repository&gt;cuda&lt;/repository&gt;&lt;key-name&gt;cudatools &amp;lt;cudatools@nvidia.com&amp;gt;&lt;/key-name&gt;&lt;key-fingerprint&gt;AE09FE4BBD223A84B2CCFCE3F60F4B3D7FA2AF80&lt;/key-fingerprint&gt;&lt;key-created time_t=&quot;1466808197&quot; text=&quot;2016-06-24T15:43:17-07&quot;/&gt;&lt;key-expires time_t=&quot;0&quot; text=&quot;0&quot;/&gt;&lt;rpm-name&gt;gpg-pubkey-7fa2af80-576db785&lt;/rpm-name&gt;&lt;/gpgkey-info&gt;", "", "Do you want to reject the key, trust temporarily, or trust always?</text>", "<option default=\"1\" value=\"r\" desc=\"Don&apos;t trust the key.\"/>", "<option value=\"t\" desc=\"Trust the key temporarily.\"/>", "<option value=\"a\" desc=\"Trust the key and import it into trusted keyring.\"/>", "</prompt>", "<message type=\"error\">Repository &apos;cuda&apos; is invalid.", "[cuda|http://developer.download.nvidia.com/compute/cuda/repos/sles122/x86_64] Valid metadata not found at specified URL", "Please check if the URIs defined for this repository are pointing to a valid repository.", "</message>", "<message type=\"error\">Skipping repository &apos;cuda&apos; because of the above error.</message>", "<message type=\"error\">Some of the repositories have not been refreshed because of an error.</message>", "</stream>"]}`

Haroon-Khel:
Unfortunately the `perl-XML-parser` fails to install, with the following error:
```
    },
    "item": "perl-XML-parser",
    "msg": "Package '+perl-XML-parser' not found.",
    "rc": 104,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "<?xml version='1.0'?>\n<stream>\n<message type=\"error\">Package &apos;+perl-XML-parser&apos; not found.</message>\n</stream>\n",
    "stdout_lines": [
        "<?xml version='1.0'?>",
        "<stream>",
        "<message type=\"error\">Package &apos;+perl-XML-parser&apos; not found.</message>",
        "</stream>"
    ]
}
```

Willsparker:
> Unfortunately the `perl-XML-parser` fails to install, with the following error:

I saw that whilst testing the above NVidia_Cuda task. In that case, I'm not sure what to do regarding the `CPAN XML::Parser` task not working for SLES12. More detail: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1033#issuecomment-568457068

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/CPAN/tasks/main.yml
```diff
@@ -47,6 +47,7 @@
 
 - name: Install XML::Parser
   command: cpanm --with-recommends --sudo XML::Parser
+  when: ! (ansible_distribution == "SLES" and ansible_distribution_major_version == "12")
```
Haroon-Khel: This doesnt skip, as it should, on SLES 12. It may be a syntax error.
Even if skipping this task on SLES 12 is a temp solution, I think the condition `ansible_distribution != "SLES" and ansible_distribution_major_version != "12"` would be more suitable

Willsparker: Fix in 👍

Willsparker:
With the latest commit, I've removed `perl-XML-parser` from `AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml` however an alternative is yet to be found.

Haroon-Khel:
@Willsparker Turns out the playbook already installs `expat`, but only for SLES 11 (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/SLES.yml#L170) @sxa555 Would you happen to know if this is intentional or simply due to a failure to update the playbooks for SLES 12? 

Ive changed the condition to SLES 12. Consequently, the `Install XML::Parser` now gets further than before, but still fails. A dependency, `Time::Zone`, fails to install:
`FAIL Installing Time::Zone failed. See /root/.cpanm/work/1577975525.3460/build.log for details. Retry with --force to force install it.`

Willsparker:
> Ive changed the condition to SLES 12. Consequently, the `Install XML::Parser` now gets further than before, but still fails. A dependency, `Time::Zone`, fails to install:
> `FAIL Installing Time::Zone failed. See /root/.cpanm/work/1577975525.3460/build.log for details. Retry with --force to force install it.`

Ah okay, thanks! I'll look to see if there's any way around that :-)

PR #1064, opened 2019-12-23
Allow Debian8 as an option to vmDestroy.sh

sxa555:


Willsparker:
I was planning on doing this yesterday when I saw it didn't work- I'm probably going to try and do it like the `vagrantPlaybookCheck.sh` is done, so we don't have to keep changing it :-)

PR #1065, opened 2019-12-24
Made Deb8 VF find IP correctly

Willsparker:
`ifconfig` is unfortunately not a thing in Debian8, therefore I found a new way to output a list of IP addresses, like the rest of the Vagrantfiles do.

### DIFF
--- a/ansible/Vagrantfile.Debian8
+++ b/ansible/Vagrantfile.Debian8
@@ -27,7 +27,7 @@ echo "*** '$VAGRANT_MOUNT/playbooks/hosts' file for ansible-playbook contains **
 cat $VAGRANT_MOUNT/playbooks/hosts
 echo "---------------------------------------------------------------------------------------"
 # Get IPs of the VM, and output to shared folder
-ifconfig | grep "inet" | awk '{print $2}' | sed 's/addr://'  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
+ip address show | grep -e "\\binet\\b" | cut -d' ' -f6 | cut -d/ -f1  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
 # Put the host machine's IP into the authorised_keys file on the VM
 [ -r /vagrant/id_rsa.pub ] && mkdir -p $HOME/.ssh && cat /vagrant/id_rsa.pub >> $HOME/.ssh/authorized_keys
 SCRIPT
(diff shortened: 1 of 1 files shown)


Willsparker:
@sxa555 This works to retrieve the IP and put it in the shared folder. However, there's a weird quirk of just the Debian8 VM, in which it won't update the shared folder once it's provisioned itself, therefore `vagrantPlaybookCheck.sh` can't find the `hosts.tmp` file that should be there but isn't. I'll put a fix into this PR if I can find one.

Willsparker:
Looking at the output and a bit of googling later, I believe it's because the Vagrant Box doesn't come packaged with `Guest Additions` : 
```
==> adoptopenjdkD8: Machine booted and ready!
==> adoptopenjdkD8: Checking for guest additions in VM...
    adoptopenjdkD8: No guest additions were detected on the base box for this VM! Guest
    adoptopenjdkD8: additions are required for forwarded ports, shared folders, host only
    adoptopenjdkD8: networking, and more. If SSH fails on this machine, please install
    adoptopenjdkD8: the guest additions and repackage the box to continue.
    adoptopenjdkD8: 
    adoptopenjdkD8: This is not an error message; everything may continue to work properly,
    adoptopenjdkD8: in which case you may ignore this message.
==> adoptopenjdkD8: Setting hostname...
==> adoptopenjdkD8: Configuring and enabling network interfaces...
```
and : https://github.com/hashicorp/vagrant/issues/6940

I'll have to find an alternative Debian8 Vagrant box

Willsparker:
Using a different box that has guest additions pre-installed has seemed to work in making the playbook run : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Debian8,label=vagrant/lastBuild/console
I attempted to use the vagrant plugin `vb-guests` however the plugin was unable to install the guest additions as it couldn't find a particular apt-package : `linux-headers-3.16.0-9-amd64`

Willsparker:
So, the original box I was attempting to use was the following: https://app.vagrantup.com/generic/boxes/debian8
And the one I've found to work is this: https://app.vagrantup.com/roboxes/boxes/debian8
Which seem to be provided by the same company.
However, the `roboxes` Debian box comes with the `Guest Additions` already installed. 
@sxa555 also found that the `vagrant vb-guest` plugin was not working on the `generic` box as the kernel package version is `9` when it is meant to be `10`, as it is in the `robox` box. This means the plugin is unable to install the package above, which is a dependency,
In a perfect world, we'd want to use the `generic` box as they seem to be the most official provider of Vagrant boxes in most OSs, however to do that we'd need to find a way to update the kernel version to 10, before running `vagrant up` as the `vagrant vb-guest` plugin runs and attempts to install the `Guest Additions` (and therefore the dependency that isn't there) before the initial provision script.

PR #1068, opened 2019-12-30
ansible: Add adoptopenjdk8 to Debian

sxa555:
Hopefully fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1067


sxa555:
Will push this through without review since it's absence is causing a break and I've tested it adequately.

Issue #1061, opened 2019-12-23
Investigate FreeBSD and SLES Vagrantfiles for testing the playbooks


Willsparker:
At the request of @sxa555 
It'd be useful to have FreeBSD and SLES vagrantfiles that can be used with the `pbTestScripts` to determine that the playbooks work with these OSs, and are able to build a JDK.

Willsparker:
Hit a bit of a snag in which the two main Vagrant Boxes for FreeBSD12, that have any significant amount of downloads (https://app.vagrantup.com/boxes/search?utf8=%E2%9C%93&sort=downloads&provider=&q=freebsd12) both mount the shared folder with `smb`, (https://www.vagrantup.com/docs/synced-folders/smb.html) which, although not an issue within itself, will require user input of a Username and password. When attempting to use the fields found in the link, (i.e. `smb_password` ) within the Vagrantfile, the following error message is shown: 
```
==> adoptopenjdkFBSD12: Preparing SMB shared folders...
Vagrant SMB synced folders require the account password to be stored
in an NT compatible format. Please update your sharing settings to
enable a Windows compatible password and try **again**
```
However, this seems to be only an issue on Mac host, as I didn't have this issue on the Ubuntu18 `infra-vagrant-1` jenkins node.

Willsparker:
As of current, the vagrant boxes provided on the vagrant website don't have a version of SLES12SP4 , which I believe is the version of SLES used for our testing machines. I''ve tried with the SLES12SP2 vagrant box located here: https://app.vagrantup.com/suse/boxes/sles12sp2
However, this causes results in the following error: 
```
localhost:ansible williamparker$ vagrant up
Bringing machine 'adoptopenjdkC6' up with 'virtualbox' provider...
==> adoptopenjdkC6: Importing base box 'suse/sles12sp2'...
==> adoptopenjdkC6: Matching MAC address for NAT networking...
==> adoptopenjdkC6: Checking if box 'suse/sles12sp2' version '0.0.1' is up to date...
==> adoptopenjdkC6: Setting the name of the VM: ansible_adoptopenjdkC6_1577454538499_56505
==> adoptopenjdkC6: Fixed port collision for 22 => 2222. Now on port 2202.
==> adoptopenjdkC6: Clearing any previously set network interfaces...
==> adoptopenjdkC6: Preparing network interfaces based on configuration...
    adoptopenjdkC6: Adapter 1: nat
[...]

PR #1073, opened 2020-01-03
VFs: Add FreeBSD12 and SLES12 Vagrantfiles

Willsparker:
fixes: #1061 

This PR will allow us to test the unix playbooks for the above OSs, so we don't run into a series of unexpected errors such as #1033 . 

FreeBSD12: I have managed to get the `generic` box working, which is the FreeBSD12 box with the most downloads and looks the most trustworthy. https://app.vagrantup.com/generic/boxes/freebsd12. FreeBSD12 doesn't support the `vboxsf` file type, so between the 2 alternatives of `smb` and `rsync`, `rsync` has been used as `smb` is not supported on linux hosts. Due to `rsync`s inability to share files from the guest VM to the host machine, the plugin `vagrant-rsync-back` has had to be used to do this.

SLES12: I've got this working using a Vagrantfile provided by SUSE. Unfortunately, this is only SLES12SP1, however there were no SLES12SP3 or SP4 vagrant boxes to speak of. https://app.vagrantup.com/suse/boxes/sles12sp1 . This vagrantfile also has to use the `rsync` file type as this box doesn't have the `Virtualbox Guest Additions` installed to them that are required for the `vboxsf` type. See the above issue for more detail. Therefore this will also make use of the `vagrant-rsync-back` plugin.

### DIFF
--- a/ansible/Vagrantfile.FreeBSD12
+++ b/ansible/Vagrantfile.FreeBSD12
@@ -4,6 +4,7 @@
 $script = <<SCRIPT
 # Doesn't come pre-installed in the vagrant box
 sudo pkg install -y python
+sudo ln -sf /usr/local/bin/python /usr/bin/python
 # Get IPs of the VM, and output to shared folder
 ifconfig | grep "inet" | awk '{print $2}' | sed 's/addr://'  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
 # Put the host machine's IP into the authorised_keys file on the VM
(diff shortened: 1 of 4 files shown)


Willsparker:
FreeBSD12 VF works in the `vagrantPlaybookCheck` job : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=FreeBSD12,label=vagrant/260/
I know it's failed, but this PR is about getting the Vagrantfiles to work in the that context. I'll open a separate PR to address the issues with the playbook.

Willsparker:
SLES12 VF now works too: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=SLES12,label=vagrant/261
Again, this has failed, though, this should be mostly fixed with #1062 - however, I am concerned about the fact this is SP1, not SP4 which is what that PR has been based upon.

PR #1083, opened 2020-01-08
UnixPB: Fixes for FreeBSD12

Willsparker:
A couple of issues I found whilst running through the FreeBSD12 playbook locally.

- if `gtar` is not installed, then creating the symlink later on isn't possible.
- `wget` is required to be installed, otherwise the `ccache` role will fail
- the `ansible_processor_cores` variable is undefined at the point it runs the `ccache` role. The `ansible_processor_vcpus` is FreeBSD's equivalent.
- The override task to turn `amd64` to `x86_64` was testing and setting the wrong variable. FreeBSD is the first instance of it being an issue, as the `ansible_architecture` variable is set to `amd64`. If not corrected, the `adoptopenjdk_install` role doesn't download or install the JDK, and then fails at the `chown` task.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
@@ -6,11 +6,11 @@
 ###########################
 # Overide amd64 to x86_64 #
 ###########################
-- name: If ansible_distribution is amd64 set it to x86_64
+- name: If ansible_architecture is amd64 set it to x86_64
   set_fact:
-    ansible_distribution: "x86_64"
+    ansible_architecture: "x86_64"
   when:
-    - ansible_distribution == "amd64"
+    - ansible_architecture == "amd64"
 
 ##################
 # Set root group #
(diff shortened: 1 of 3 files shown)


Willsparker:
Testing the branch at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/294/

(The link may not work for awhile as it's in a queue to be run)

Willsparker:
Just confirmed that a build of JDK8 works on a FreeBSD12 VM set up with this playbook, using the same arguments as in this Jenkins Job: https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-freebsd-x64-hotspot/20/

Willsparker:
Unfortunately, I've ran into an issue with the JDK11 build, where, despite using the same arguments as the last succeeding Jenkins build ( https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-freebsd-x64-hotspot/9/ ) , It output the the error: 
```
configure: Toolchain type clang is not valid on this platform.
configure: error: Cannot continue.
configure: Valid toolchains: .
configure exiting with result code 1
```
Specifying toolchain to be `gcc` or not specifying at all output with similar problems, either that `gcc` is not a valid toolchain type, or that a C compiler can't be found (though `gcc48` is installed).

gdams:
@Willsparker what source code did the job clone from? was it definitely the freebsd one rather than the generic AdoptOpenJDK source?

Willsparker:
` BUILD_CONFIG[REPOSITORY]="https://github.com/battleblow/openjdk-jdk11u" ` 
I believe the FreeBSD one

Issue #1090, opened 2020-01-13
"ansible_processor_vcpus" reported as '***Undefined***' on s390x RHEL7


sej-jackson:
Running playbooks on new Rhel7.7 system (s390x) which fell over in gmake task because they are reporting `ansible_processor_vcpus` as undefined.

>TASK [gmake : Compile and install make 4.1 from source] ************************
>fatal: [jsvt302.hursley.ibm.com]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible_processor_vcpus' is undefined\n\nThe error appears to be in '/home/sej42282/gitibm/runtimes/infra-can/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml': line 40, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Compile and install make 4.1 from source\n  ^ here\n"}

`/proc/cpuinfo` contains:
```
# cat /proc/cpuinfo
vendor_id       : IBM/S390
# processors    : 2
bogomips per cpu: 21881.00
max thread id   : 0
features        : esan3 zarch stfle msa ldisp eimm dfp edat etf3eh highgprs te vx vxd vxe gs sie 
facilities      : 0 1 2 3 4 6 7 8 9 10 12 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 30 31 32 33 34 35 36 37 38 40 41 42 43 44 45 46 47 48 49 50 51 52 53 54 55 57 58 59 60 73 74 75 76 77 80 81 82 128 129 130 131 133 134 135 146 147 156 168
cache0          : level=1 type=Data scope=Private size=128K line_size=256 associativity=8
cache1          : level=1 type=Instruction scope=Private size=128K line_size=256 associativity=8
cache2          : level=2 type=Data scope=Private size=4096K line_size=256 associativity=8
cache3          : level=2 type=Instruction scope=Private size=2048K line_size=256 associativity=8
cache4          : level=3 type=Unified scope=Shared size=131072K line_size=256 associativity=32
cache5          : level=4 type=Unified scope=Shared size=688128K line_size=256 associativity=42
processor 0: version = FF,  identification = 313F07,  machine = 3906
processor 1: version = FF,  identification = 313F07,  machine = 3906
```

[...]

PR #1092, opened 2020-01-14
UnixPB: Set vars for s390x and remove unneeded tasks

Willsparker:
fixes: #1090 

The addition of the task in the `../Common` role will stop there being any point of having so many tasks that are specific to `s390x` that just replace the `ansible_processor_vcpus` variable with `ansible_processor_cores` variable.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
@@ -12,6 +12,15 @@
   when:
     - ansible_architecture == "amd64"
 
+###############################
+# Set ansible_processor_vcpus #
+###############################
+- name: If ansible_processor_cpus is not defined, set to ansible_processor_cores or 1
+  set_fact:
+    ansible_processor_vcpus: "{{ ansible_processor_cores | default(1) }}"
+  when:
+    - ansible_processor_vcpus is not defined
+
 ##################
 # Set root group #
 ##################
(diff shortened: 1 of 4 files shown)


sxa555:
> stop there being any point of having so many tasks that are specific to s390x

Hmmm looking at this PR we haven't done it in a global way and we have modified each individual role - is there any way we can do it in a way that doesn't require changing each role as that's quite messy.

jdekonin:
I think how it is originally is correct, without specifying a tag for this task.  I believe in this manner it will always run even if someone only specifies a `cmake` tag.  Putting it under a `debug` tag would imply it has to run that tag as well as `cmake` in order to be set.

Haroon-Khel:
Yes, the `debug` tag is not required. But I would recommend that the task runs before the Debug role

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
```diff
@@ -12,6 +12,15 @@
...
 
+#########################################
+# Set ansible_processor_vcpus for s390x #
+#########################################
+- name: If ansible_architecture is s390x set ansible_processor_vcpus to ansible_processor_core
+  set_fact:
+    ansible_processor_vcpus: "{{ ansible_processor_cores }}"
+  when:
+    - ansible_architecture == "s390x"
+
```
sxa555: Agree with @Haroon-Khel on this - if we can add `ansible_processor_vpus is undefined` or whatever the valid syntax is to this `when` section that would be good ... In fact it could replace the `s390x` check :-)

Haroon-Khel: Yes, as this problem only occurs on RHEL (i think) and not every distribution on `s390x`

Willsparker:
In the latest commit, I've ensured that the `vcpus` is only set to `cores`, if `vcpus` is undefined. If `cores` is also undefined, then it defaults to `1` (I noticed in testing that this is the case with FreeBSD).

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
Willsparker: Commit in :-)

PR #1094, opened 2020-01-14
unixPB: Enable protobuf for OpenJ9 JITserver

sxa555:
Jan 2020 OpenJ9 releases on Linux/x64 will be built with JITserver support enabled, which requires protobuf.

Discussed in https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/

Testing at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/323/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -40,6 +40,7 @@
     - gcc_48
     - gcc_7                       # OpenJ9
     - cmake                       # OpenJ9 / OpenJFX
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
@sxa555 Is the Protobuf role required for tests? If not, then it would be convenient if it were tagged with `build_tools`

sxa555:
> @sxa555 Is the Protobuf role required for tests? If not, then it would be convenient if it were tagged with `build_tools`

Very good question. And it's a definite "No". The protobuf code gets statically linked in I believe so it should have that tag added too.

PR #1095, opened 2020-01-15
pbTests: Update buildJDK.sh to support FreeBSD & openSUSE

Willsparker:
Now that FreeBSD12 passes the playbook, `buildJDK.sh` needs to be altered to be able to set the correct environment variables for it to build the correct JDK.

I've also added in the ability to pass in JDKs, so we have the option to build more than JDK8 on these playbooks. 
As we don't build openJ9 on FreeBSD, that defaults to Hotspot, however linux distributions are built in `openJ9` as they require more dependencies be installed.

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -32,8 +32,13 @@ export JDK7_BOOT_DIR=$(find /usr/lib/jvm/ -name java-1.7.0-openjdk.x86_64)
 # Differences in openJDK8 name between Ubuntu and CentOS
 export JAVA_HOME=$(find /usr/lib/jvm/ -name java-1.8.0-openjdk-\*)
 if [ -z "$JAVA_HOME" ]; then
-  export JAVA_HOME=$(ls -1d /usr/lib/jvm/adoptopenjdk-8-* | head -1)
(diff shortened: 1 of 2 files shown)


Willsparker:
vagrantPlaybookCheck Run :
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/327/

sxa555:
Failed with:
`12:32:05 ls: /usr/lib/jvm/adoptopenjdk-8-*: No such file or directory`

Willsparker:
Ah, apologies - I forgot to set `JAVA_HOME` in FreeBSD's case. Latest commit should fix it : 
`vagrantPlaybookCheck` run: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/328/OS=FreeBSD12,label=vagrant/

Willsparker:
As I'm currently trying to get the playbooks for openSUSE working, I thought I'd update the PR to include the use case of building on it, to save making an additional PR.

sxa555:
Still not seeing a passing FreeBSD run :-)

Willsparker:
@sxa555 https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=FreeBSD12,label=vagrant/331/console
Latest commit got it :-)

PR #1100, opened 2020-01-17
Added build tags to appropriate roles

Haroon-Khel:
Related to https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/683

The tags `build_tools`, `build_tools_openj9` and `build_tools_openjfx` have been added to certain roles/tasks so that these can be skipped if creating only a test machine.

~~This is a work in progress, please do not merge yet~~

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -5,7 +5,7 @@
 # Groups can be passed in as a command-line variable in Ansible playbook.
 # It can be defined as 'all' or a specific group which the host belongs to.
 # For example, it can be 'all' or 'x86' for when a host is in the group 'x86'.
-- hosts: all
+- hosts: "{{ Groups | default('localhost:build:test:!*zos*:!*win*:!*aix*') }}"
   gather_facts: yes
   tasks:
     - block:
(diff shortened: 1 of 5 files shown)


Haroon-Khel:
@sxa555 This playbook has been tested on Centos 7, RHEL 7, Ubuntu 16 and 18, all x64. Its still a bit finicky on SLES 12; the playbook runs successfully, but git will not work well on the machine. This issue was present on the other distributions. but was resolved by installing libcurl-devel. I haven't found this fix to work with SLES 12.

Haroon-Khel:
On the machines on which the playbook ran successfully, I ran a significant number of jdk8 tests using jtreg. There were some failures, but these did not differ in number than a 'control' machine (a centos 7 fyre machine on which the playbook ran without skipping any build tags)

Willsparker:
@Haroon-Khel I'll test this locally on my machine tomorrow. I can't use `vagrantPlaybookCheck` as it will need to be modified to add an option to use these new tags :-)

Willsparker:
Got round to testing it - when running `ansible-playbook  ... --skip-tags adoptopenjdk,jenkins,build_tools,build_tools_openJ9 ...` the playbook ran through as expected on an Ubuntu 16 x86_64 machine, and was able to test a downloaded JDK8 version (by running a modified `./pbTestScripts/testJDK.sh` script).

Haroon-Khel:
@Willsparker @sxa555 Do you think further Java tests are required on the machines on which this playbook can run, for us to safely say these build tags are satisfactory?

karianna:
> @Willsparker @sxa555 Do you think further Java tests are required on the machines on which this playbook can run, for us to safely say these build tags are satisfactory?
> 
> Or are jdk8 tests sufficient?

I'll call it good enough testing.

PR #1102, opened 2020-01-17
unixPB: Remove duplicate protobuf install

sxa555:
We wish to use the separate role in order to force the use of GCC7 for the OpenJ9 builds as the code is statically linked in during build time.


sxa555:
Validating at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/345/

PR #1110, opened 2020-01-24
Set ansible_processor_vcpus for cmake role

vsebe:
Set ansible_processor_vcpus to ansible_processor_cores when defined or
default it to 1.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
@@ -20,6 +20,8 @@
     ansible_processor_vcpus: "{{ ansible_processor_cores | default(1) }}"
   when:
     - ansible_processor_vcpus is not defined
+  tags:
+    - cmake
 
 ##################
 # Set root group #


jdekonin:
This same block of code was removed via https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1092 and made generic under Common/tasks/main.yml.

Can you supply more information on how you hit an error and this was needed?

vsebe:
I run cmake tag only and got: `The error was: 'ansible_processor_vcpus' is undefined`.

jdekonin:
To me this is a valid use case.  Thoughts @sxa555?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
```diff
@@ -40,6 +40,21 @@
...
+# ansible_processor_vcpus is not defined on Linux s390x
+# set it to {{ ansible_processor_cores }} if defined or to 1
+- name: Set ansible_processor_vcpus variable
+  set_fact:
+    ansible_processor_vcpus: "{{ ansible_processor_cores | default(1) }}"
+  when:
+    - ansible_processor_vcpus is not defined
+    - (cmake_installed.rc != 0 ) or (cmake_installed.rc == 0 and cmake_version.stdout is version_compare('3.11.4', operator='lt') )
+    - (ansible_architecture == "x86_64") or (ansible_architecture == "ppc64le") or (ansible_architecture == "s390x")
+
```
Haroon-Khel: If you're running the `cmake` role only, then there should be a `cmake` tag for this task no?

jdekonin: I think the right fix might be to add the tag, and any others that might rely on it, to the task in ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml.

Haroon-Khel: Within the Unix playbook, the following are tags of the roles, excluding Common, that use ansible_processor_vcpus:

- `git_source`
- `ccache`
- `cmake`
- `gcc-4.8`
- `local_srcinstall` (though this role is to be removed via https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1102)
- `nagios_plugins`

Haroon-Khel: Within the Unix playbook, the following are tags of the roles, excluding Common, that use ansible_processor_vcpus:

- `git_source`
- `ccache`
- `cmake`
- `gcc-4.8`
- `local_srcinstall` (though this role is to be removed via https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1102)
- `nagios_plugins`

Haroon-Khel:
@vsebe instead of adding the task of setting `ansible_processor_vcpus` to the `cmake` role, it would be worth adding the multiple tags, discussed in the earlier comment, to https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/261cc6a405a855c9c3602b5bf268f17194a4d1d3/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml#L22

PR #1113, opened 2020-01-27
winPB: Update checksums for win_get_url module

Willsparker:
Ref: #1058 

@Haroon-Khel found that as of ansible 2.8, the `win_get_url` module separates the checksum and checksum algorithm fields, however (weirdly) not Unix's `get_url` module. Before ansible 2.8, `checksum` wasn't in the documentation at all.
https://docs.ansible.com/ansible/latest/modules/win_get_url_module.html   <-- latest
https://docs.ansible.com/ansible/2.7/modules/win_get_url_module.html       <-- 2.7

However, before ansible 2.8, it must've been planned as putting the checksum option in (as it is now) doesn't cause crashes. This wasn't noticed as the Jenkins nodes that runs the `vagrantPlaybookCheck` job are on ansible 2.5.1, where (presumably) the checksum option was just ignored.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/7-Zip/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/7-Zip/tasks/main.yml
@@ -9,7 +9,8 @@
     url: https://www.7-zip.org/a/7z1805-x64.exe
     dest: 'C:\temp\7z.exe'
     force: no
-    checksum: sha256:c1e42d8b76a86ea1890ad080e69a04c75a5f2c0484bdcd838dc8fa908dd4a84c
+    checksum: c1e42d8b76a86ea1890ad080e69a04c75a5f2c0484bdcd838dc8fa908dd4a84c
+    checksum_algorithm: sha256
   tags: 7zip
 
 - name: Install 7-Zip
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
@@ -20,7 +20,8 @@
     url: https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.5-bin.zip
     dest: c:\temp\ant.zip
     force: no
-    checksum: sha256:9028e2fc64491cca0f991acc09b06ee7fe644afe41d1d6caf72702ca25c4613c
+    checksum: 9028e2fc64491cca0f991acc09b06ee7fe644afe41d1d6caf72702ca25c4613c
+    checksum_algorithm: sha256
   when: (ant_installed.stat.exists == false) and (ant_download.stat.exists == false)
   register: ant_download
   tags: ANT
@@ -57,7 +58,8 @@
     url: https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.zip
     dest: c:\temp\ant-contrib.zip
     force: no
-    checksum: sha256:22bae6c3ddf1a464b285784599eef8698f64dde24378c77e42522a536b88cbbc
+    checksum: 22bae6c3ddf1a464b285784599eef8698f64dde24378c77e42522a536b88cbbc
+    checksum_algorithm: sha256
   when: (ant_contrib_installed.stat.exists == false)
   tags: ANT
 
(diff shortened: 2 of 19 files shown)


Haroon-Khel:
@Willsparker I am currently testing this new checksum_algorithm parameter. Some of the tasks, like `cygwin` and `MSVS_2017` still return checksum errors. For example 
```
"msg": "The checksum for https://aka.ms/vs/15/release/vs_community.exe did not match '14559f581aa0d2bb841f236d00e02d8f743f881da9077dea630ecef3658c6593', it was '14e078f4da89990e3c05775283adb8802d937f69a08ac8d0b6c683dc3ab6a0e8'",
```

Also, do you know if this `checksum_algorithm` parameter is backwards compatible? ie, will the playbook run for ansible versions less that 2.8?

Haroon-Khel:
I am also currently testing the windows playbook without the `checksum_algorithm` parameter on a 2019 windows machine, using **ansible 2.7.0**. If the same checksum errors, for the tasks listed in my previous comment, occur here, then we will know these checksum errors aren't associated with the omission of the `checksum_algorithm` parameter, but rather something else

Willsparker:
> @Willsparker I am currently testing this new checksum_algorithm parameter. Some of the tasks, like `cygwin` and `MSVS_2017` still return checksum errors. For example, for the task `MSVS_2017 : Download Visual Studio Community 2017`

Ah, in that case the downloadable will have a different checksum to what I put in. I'll get back to you on that. 

> Also, do you know if this `checksum_algorithm` parameter is backwards compatible? ie, will the playbook run for ansible versions less that 2.8?

Almost definitely not, but I will again check - I'm in the process of upgrading `infra-vagrant-T541` to ansible 2.9.4, so I'll let you know then if that fixes the win_get_url error. If it does, I'll run through a `vagrantPlaybookCheck` job with this branch to see if the checksums are wrong, as well as local testing.

Willsparker:
> I am also currently testing the windows playbook without the `checksum_algorithm` parameter on a 2019 windows machine, using **ansible 2.7.0**. If the same checksum errors, for the tasks listed in my previous comment, occur here, then we will know these checksum errors aren't associated with the omission of the `checksum_algorithm` parameter, but rather something else

According to documentation, the checksum algorithm defaults to `sha1`, whereas I've put all the checksums in `sha256` format. If that parameter is omitted, then it'll assume the checksum is `sha1`

Willsparker:
Alright, so I tested `2.9.4` and the `win_get_url` module works properly - I tested it on `GIT`, a task I know that was the issue. Having put the fix that I made in the PR in, it worked fine. This is the same case with `MSVS_2010`, with the current checksum.

Willsparker:
@Haroon-Khel The checksums should be correct now, with the latest commit :-)

Haroon-Khel:
@Willsparker Thanks. Ill test it on a windows 2019 machine using both 2.8.5 and 2.7.0 versions of ansible

Willsparker:
Latest Commit is to change the Freemarker URL, as this was a consistent error message: 
```
TASK [Freemarker : Download freemarker] ****************************************
fatal: [172.28.128.250]: FAILED! => {"changed": false, "dest": "C:\\openjdk\\freemarker.jar", "msg": "Error downloading 'http://central.maven.org/maven2/freemarker/freemarker/2.3.8/freemarker-2.3.8.jar' to 'C:\\openjdk\\freemarker.jar': The remote server returned an error: (501) Not Implemented.", "status_code": 501, "url": "http://central.maven.org/maven2/freemarker/freemarker/2.3.8/freemarker-2.3.8.jar", "win_get_url": {"dest": "C:\\openjdk\\freemarker.jar", "url": "http://central.maven.org/maven2/freemarker/freemarker/2.3.8/freemarker-2.3.8.jar"}}
```

With this, the Windows Playbook should run through to it's entirety (now with the checksums to ensure the downloads are correct). Test with `vagrantPlaybookCheck` : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/346/

Willsparker:
Nice - I'm just having a run through of the Jenkins job too :-)

Haroon-Khel:
I wasnt able to run the cygwin and the MSVS_2017 tasks, as their install took too long, to the point where ansible's connection timed out. Im not sure if youll run into the same problems with jenkins

Willsparker:
> I wasnt able to run the cygwin and the MSVS_2017 tasks, as their install took too long, to the point where ansible's connection timed out. Im not sure if youll run into the same problems with jenkins

If the download worked, then the checksum should be correct :-) It's currently running through now : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/349/console

Willsparker:
The run through succeeded 👍

PR #1114, opened 2020-01-28
unixPB: Pass CC/CXX variables to Protobuf build task

Willsparker:
Problem found as of upgrading Ansible to 2.9.4 - Using the `set_fact` module doesn't actually set the variables for the environment anymore, therefore they need to be passed in on the task. 
This wouldn't be an issue as supposedly GCC 4.8 can build Protobuf, as well as GCC-7, however in the instances when the default GCC is <4.8 (i.e. the OpenSUSE Vagrant VM), this is required.


Willsparker:
This fixes the issue I was seeing in OpenSUSE : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=SUSE12,label=vagrant/350/ (the build failed due to a current issue with the build scripts, not the playbook)

sxa555:
FYI @jdekonin - the previous method doesn't appear to properly take effect at least with newer ansible versions

PR #1116, opened 2020-01-28
Remove zulu-9 from Additional_Build_Tools var lists

### LINKED ISSUE #1108: Remove zulu-9 from the Linux/x86 playbooks
This is causing confusion on the Ubuntu systems (and possible CentOS/RHEL) as it seems to be setting it as a default. Since java 9 is out of support this is not ideal. Also we should not require zulu 9 on the machines as we have a valid adoptopenjdk9 available if required now.

Since I can't see any trace of zulu 7 being installed either it may also be possible to remove the zulu repositories entirely from t he playbooks.

sej-jackson:
The inclusion of zulu-9 in the Additional_Build_Tools variables lists causes the `/etc/alternatives/java` symlink to be redirected to the zulu-9 version of java on those systems where it is installed.

Removing zulu-9 from the list fixes #1108

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
@@ -20,27 +20,6 @@
     state: latest
   tags: patch_update
 
-#########################################
-# Configure Repos and Update the system #
-#########################################
-- name: Import AZUL public key
-  rpm_key:
-    state: present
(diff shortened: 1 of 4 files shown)


Willsparker:
Running the playbooks through `vagrantPlaybookCheck` :-) Can't see why this would be an issue though, as we _shouldn't_ need any version of JDK9 as we don't build 9/10 anymore. 
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/351/

(The builds may still fail due to the --macos option issue FYI)

Issue #1059, opened 2019-12-23
Investigate qemu for validating playbooks on Linux on ppc64le/s390x


sxa555:
See e.g.
- Official qemu ppc64 info at https://wiki.qemu.org/Documentation/Platforms/POWER
- https://djanotes.blogspot.com/2018/09/running-ubuntu-ppc64le-virtual-machine.html
- https://wiki.alpinelinux.org/wiki/Ppc64le
- https://stackoverflow.com/questions/53259899/is-it-possible-to-run-vm-with-ppc64le-architecture-on-a-host-machine-with-x86-64

Willsparker:
I had a look at ci.adoptopenjdk.net, and it appears we need the following iso images, if only looking at the ppc64le, s390x and arm architectures: 

| U16 | ppc64le, s390x, armv7, armv8 |
|-----|-------------------------------|
| U18 | ppc64le, armv8               |
| C7  | ppc64le, armv8               |

Am I missing any?

sxa555:
I think that since we have good coverage across distributions on x64 I'm happy with just going for just one on each architecture for now. Let's go Ubuntu18.04 where available, and Ubuntu16.04 otherwise for now.

Willsparker:
Okay - unfortunately the `armv7l` architecture is available for either Ubuntu's, so I'm looking at Raspbian instead , as per your suggestion.

Willsparker:
https://worthdoingbadly.com/aixqemu/ 

Apparently AIX is possible too!

Willsparker:
According to https://superuser.com/questions/297162/how-to-access-internet-from-qemu qemu is setup to access the internet by default.
I've confirmed this by running `qemu-system-ppc64 -M pseries -m 1024 -cdrom U18-ppc64el.iso -boot d -hda  U18-ppc66le.dsk `
Once running through the installation and then rebooting the system from the disk (by removing the `-cdrom` and `-boot` option), I was able to run `git clone https://github.com/adoptopenjdk/openjdk-infrastructure` , `sudo apt install ansible` and begin running the playbook with the following output : 

![Screenshot 2020-01-21 at 10 28 30](https://user-images.githubusercontent.com/52783864/72797320-ce961980-3c38-11ea-87bd-0cfbab9851a7.png)

Willsparker:
[...]

PR #1119, opened 2020-01-30
pbTests: Add script to test U18 on s390x/ppc64le QEMU VMs

Willsparker:
A script that is able to build a Qemu VM to run the playbook, build a JDK and test that JDK on. With Qemu, we're able to test the playbooks on other architectures by emulating them (admittedly at the expense of speed). 
As well as this, `buildJDK.sh` has been modified to allow for different architectures.

The script (currently) is able to create an s390x and ppc64le machine with Ubuntu server running on them, port forward to allow for ssh connections, and begin running the playbook and build scripts on them. They currently use a set of images that are already stored on the machine - @sxa555 suggested we keep these in a standardised place on Jenkins nodes so we can just copy them over for each new run.

fixes: #1059


Willsparker:
With the latest commit, the PPC64LE architecture works in the job : 
https://ci.adoptopenjdk.net/job/Will-QEMUplaybookTest/ARCHITECTURE=ppc64le,label=infra-vagrant-1/6/console

For some reason, not the `s390x` architecture. I feel this may be due to it taking longer than 2 minutes to boot the qemu VM

Willsparker:
The latest commit will pull my `qemu` branch to get the Unix playbook to run. This so I can use the `Will-QEMUPlaybookTest` to test the changes, without having to have anything merged. However, when this PR is merged, this will need to be changed back to the standard `openjdk-infrastructure` repo.

Willsparker:
I believe with the latest `force push`, this is ready to be taken out of `[WIP]`. The latest `QEMUPlaybookTest` ( https://ci.adoptopenjdk.net/job/Will-QEMUplaybookTest/18/console ) run came back with ppc64le and s390x able to run the playbook and build JDK8-J9. The test on the JDK didn't run as this branch is older and couldn't find `MachineInfo` as a `BUILD_TARGET` - the latest `push` fixed this, as `MachineInfo` was changed to `BUILD_TARGET=system`.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ant/tasks/main.yml
```diff
@@ -20,7 +20,6 @@
     mode: 0440
     timeout: 25
     validate_certs: no
-    checksum: sha512:2e48f9e429d67708f5690bc307232f08440d01ebe414059292b6543971da9c7cd259c21533b9163b4dd753321c17bd917adf8407d03245a0945fc30a4e633163
```
sxa555: Why remove this checksum?

Willsparker: There was an error message that I thought it was causing - I checked yesterday and I was wrong so I'll put it back in :)

Willsparker: Running again just to ensure the error isn't there: https://ci.adoptopenjdk.net/job/Will-QEMUplaybookTest/20/

Issue #1125, opened 2020-02-05
Update playbooks to handle RHEL 8


sej-jackson:
Now that RHEL 8 is available, let's get the playbooks updated to handle it.

Known issues so far:
- `ntp` package is no longer available (use `chrony` instead - should be installed by default)
- `java-1.7.0-openjdk-devel` 
- Asian locales need prereq packages: `glibc-locale-source` `glibc-langpack-XX`
- various problems setting up CPAN, but apparently no longer required (so skip it?)
- gmake build fails - but can use native version 4.2.1 instead

Willsparker:
Due to CentOS and RHEL being very similar, we can look into doing them both at the same time.

PR #1129, opened 2020-02-06
updates to support RHEL 8 playbooks

sej-jackson:
Added tasks for RHEL8 to install Common packages, and skip CPAN & gmake roles for issue #1125.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/CPAN/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/CPAN/tasks/main.yml
@@ -32,6 +32,8 @@
 # All
 - name: Install JSON
   command: cpanm --with-recommends --sudo JSON
+  when:
+    - not ((ansible_distribution == "RedHat" or ansible_distribution == "CentOS") and ansible_distribution_major_version == "8")
   tags:
     - cpan
     - json
@@ -40,6 +42,8 @@
 
 - name: Install Text::CSV
   command: cpanm --with-recommends --sudo Text::CSV
+  when:
+    - not ((ansible_distribution == "RedHat" or ansible_distribution == "CentOS") and ansible_distribution_major_version == "8")
   tags:
     - cpan
     - text_csv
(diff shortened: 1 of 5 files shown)


sej-jackson:
I believe the `libdwarf-devel` and `libmpc-devel` packages need to come from another repository - possibly `ftp3-CRB`?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/CPAN/tasks/main.yml
```diff
@@ -32,6 +32,8 @@
 # All
 - name: Install JSON
   command: cpanm --with-recommends --sudo JSON
+  when:
+    - not ((ansible_distribution == "RedHat" or ansible_distribution == "CentOS") and ansible_distribution_major_version == "8")
```
gdams: @smlambert Am I correct in saying that we no longer need these cpan packages at all? If so we should just remove this role.

sxa555:
We should ideally try and put these into `CentOS.yml` files in the `Common` role too to keep them in sync with RHEL

sej-jackson:
> We should ideally try and put these into `CentOS.yml` files in the `Common` role too to keep them in sync with RHEL

Yes, @Willsparker is going to do them - he's just helping me with my changes at the moment....

Willsparker:
I've been having issues with CentOS8, namely this: 
```
# sudo yum upgrade
CentOS-8 - Base                                                                                           1.2 kB/s | 2.9 kB     00:02    
Error: Failed to synchronize cache for repo 'BaseOS'
```
Not sure what the issue is, but I'll put the CentOS8 changes into a separate PR

Willsparker:
If nobody objects, I'm going to open a separate PR for CentOS8 changes , as I'll take the opportunity to add a CentOS8 Vagrantfile at the same time :-)

sej-jackson:
> If nobody objects, I'm going to open a separate PR for CentOS8 changes , as I'll take the opportunity to add a CentOS8 Vagrantfile at the same time :-)

Go for it. :)

Willsparker:
Squashed and merged - the non-Common role changes will be tested regularly with the CentOS8 PR is merged with the `VagrantPlaybookCheck` Jenkins job. The `Common` role changes were tested before on an internal machine too :)

PR #1131, opened 2020-02-07
UnixPB: Add support for CentOS8

Willsparker:
Ref #1129  
Fixes: #1125 

Additions to the playbooks to allow for CentOS8 to build, after 1129 is merged. Included the addition of a Vagrantfile so CentOS8 can be tested with the `vagrantPlaybookCheck` job.

PR #1136, opened 2020-02-11
 UnixPB: Remove CPAN role

Willsparker:
Initially bought around from @gdams comment https://adoptopenjdk.slack.com/archives/C53GHCXL4/p1580993701028200
talking about this PR #1129


Willsparker:
Just in the process of removing additional code George suggested to me, I'll put it through the job once that's in :-)

Willsparker:
Playbook job is running through : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/357/

gdams:
@smlambert do we still need the perl deps? I wasn't sure if they could be ditched too??

smlambert:
Mentioned in Slack, tests still need core perl (modules that come with standard perl install), the extra CPAN modules are no longer a prereq.  This PR looks right.

karianna:
@Willsparker merge conflict

Willsparker:
The Jenkins job above has completed, with most OSs succeeding - the ones that failed are due to the Vagrantfile not being there ( CentOS8, will be fixed by #1131 ) and the strange recurring bug I have with Vagrant VMs turning each other off when a playbook finishes. Removing the CPAN role doesn't affect building JDK8-J9 anyway 👍

PR #1138, opened 2020-02-11
unixPB: Alter adoptopenjdk_install role to allow for jdk8

Willsparker:
Based off the `adoptopenjdk_install` role. I did this as it seems there's a lot of discrepancy between the linux distributions about whether `openjdk8` is available using their specific package manager. If there is, they can end up in different directories, with different names.

With this role, we can ensure all distros have a version of JDK8, and they are in a standardised place with a standardised name. This can also mitigate any issues when trying to build JDK8 without the package managers having JDK7 available.


Haroon-Khel:
@Willsparker The role successfully installs openjdk8 when there isn't one available in `/usr/lib/jvm`, but it also installs if it is already present. For example, `java-1.8.0-openjdk-1.8.0.242.b07-1.el6_10.x86_64` was already in `/usr/lib/jvm`, and yet the playbook downloaded and installed `jdk8u242-b08` into that directory

sxa555:
What is this doing that requires a separate role from `adoptopenjdk_install` that means we need a new role? Also I'm not sure this quite helps is put it in a "standardised place" as it's subject to whatever the latest release is at the time of the playbook execution, so we still have to specify the location in the jenkins configuration

Willsparker:
I initially attempted to add it to the `adoptopenjdk_install` role, but due to the naming scheme `jdk8u...` as opposed to `jdk-XX...`, there would have the be a duplication of the `Checking if it's installed` and `Get full path name` tasks - one for when `jdk_version == 8` and one for when `jdk_version != 8`. 
This also caused some issues in conditionals - for example, the initial `Check if it's installed` tasks assigned a variable `adoptopenjdk_installed`. If the JDK8 specific version of the task following it is skipped, for example, then that variable becomes undefined, and will fail for `install latest release` tasks. Therefore it can only succeed for one of `jdk_version == 8` and `jdk_version != 8` (depending on the way round you assign the variable)
If we assign a different variable to that i.e. `adoptopenjdk8_installed` and then put the conditional ` when (adoptopenjdk8_installed.rc != 0) or (adoptopenjdk_installed.rc != 0)`onto the install task then one of those will be undefined and will error.

At that point, there'd need to be JDK8 specific version of the task for every task in the `adoptopenjdk_install` role - therefore we should just create a separate role.

Willsparker:
> @Willsparker The role successfully installs openjdk8 when there isn't one available in `/usr/lib/jvm`, but it also installs if it is already present. For example, `java-1.8.0-openjdk-1.8.0.242.b07-1.el6_10.x86_64` was already in `/usr/lib/jvm`, and yet the playbook downloaded and installed `jdk8u242-b08` into that directory.
> 
> This was tested on x86_64

The first one you mentioned `java-1.8.0-openjdk ...` will have been installed on the machine via the package manager (i.e. `apt install java-1.8.0-openjdk` or `yum install openjdk-8-jdk`) which is fine to have in addition to the adoptopenjdk one. The reason I want to add this role is because some distributions of linux wont necessarily be able to install any version of openjdk8 / java8 using their package manager

gdams:
@Willsparker I think I'm right in saying that this will run on macOS which you probably want to skip as we use homebrew. Also you should be able to expand the architecture variable inn the URL rather than having a single task for each architecture

Willsparker:
@gdams Good point ... I'll add those conditions. And I'll do some testing with the architecture variable, but agreed, that'd be a much cleaner way to do it :-)

Willsparker:
Running through it with the `vagrantPlaybookCheck` job : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/358/

sxa555:
Ref the reasons above for having the role separate, can I propose modifying `adoptopenjdk_install` to have a check at the top to set a `jdk_path` variable based on whether `jdk_version` is equal to 8, and set it differently if it is not later than (not equal would also work) to 8. Using this new `jdk_path` variable instead of `/usr/lib/jvm/jdk-{{ jdk_version }}` would resolve the above concerns I believe.

sxa555:
Thanks for the updates - verifying at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/371

PR #1140, opened 2020-02-12
UnixPB: Add support for SLES15-SP0

Willsparker:
Requested by @sej-jackson 

Additions to the `../common/*/SLES.yml` files, to allow for the playbook to run through on SLES15 and build JDK8-J9 on it. Most of the added packages were available from the default repos on the machine, however `devel-tools` was added for `libdwarf-devel` to be installed. 
Tested on a SLES15 SP0 internal machine. This initially didn't have `sudo` installed however - the `git_source` role will fail if this is the case, however I assumed that _actual_ machines would already have this on the machine?

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
@@ -17,7 +17,6 @@ Build_Tool_Packages:
   - libdw1
   - libelf1
   - make
-  - ntp
   - pkg-config
   - systemtap-sdt-devel
   - unzip
@@ -45,8 +44,8 @@ Additional_Build_Tools_SLES15:
 Additional_Build_Tools_SLES12:
   - java-1_8_0-openjdk
   - git-core
-  - libfreetype6
   - libelf0
+  - libfreetype6
   - libXext6
   - libXi6                        # JDK12+ compilation
   - libXrandr2                    # JDK12+ compilation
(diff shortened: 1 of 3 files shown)


Willsparker:
ref #1141

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/SLES.yml
```diff
@@ -45,6 +63,7 @@ Additional_Build_Tools_SLES11:
 Additional_Build_Tools_SLES_x86:
   - glibc-32bit                   # a dependency required for executing a 32-bit C binary
   - glibc-devel-32bit             # a dependency required for executing a 32-bit C binary
+  - libelf0
```
Haroon-Khel: Ran the playbook on a x86_64 SLES 15 SP0 fyre machine. This package failed to install, with an error message along the lines of No provider of libelf0 found.

Willsparker: AH! Sorry, I must've put it in the wrong place.

Willsparker:
Latest commit should address both requested changes :-)

Haroon-Khel:
@Willsparker The build packages that youve added run fine on my x64 sles 15 test machine

Willsparker:
@sej-jackson That's okay! I'll add SLES15 to the list of conditions for the task we made for RHEL8, and move ntp to the SLES11 and SLES12 specific package lists

PR #1147, opened 2020-02-18
UnixPB: remove sudo from GIT_SOURCE role

Willsparker:
Noticed whilst fixing the playbooks for SLES15 SP1, that the `GIT_SOURCE` role has tasks that use `sudo`. The SLES15 SP1 environment I was running didn't have `sudo` installed, and so the `./configure and make` task failed. However, `sudo` isn't required to be used as `become: yes` is used. 

VagrantPlaybookCheck Run: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/365/


Willsparker:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1141#issuecomment-587387607

PR #1152, opened 2020-02-19
pbTests: Change buildJDK to build from other URLs

Willsparker:
Requested by @sxa555 

Update `buildJDK.sh` to be able to build JDKs other than JDK8-J9, as well as give it the option to use other branches of the openjdk-build repository.

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -1,6 +1,63 @@
 #!/bin/bash
 set -eu
 
+processArgs() {
+	while [[ $# -gt 0 ]] && [[ ."$1" = .-* ]] ; do
+		local opt="$1";
+		shift;
+		case "$opt" in
+			"--version" | "-v" )
+				JAVA_TO_BUILD="$1"; shift;;
+			"--URL" | "-u" )
+				GIT_URL="$1"; shift;;
+			"--openj9" | "-j9" )
+				VARIANT=openj9;;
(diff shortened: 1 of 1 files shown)


ansible/pbTestScripts/buildJDK.sh
```diff
@@ -1,26 +1,118 @@
...
+		echo "No branch detected"
+		git clone $GIT_URL $WORKSPACE/openjdk-build
+	fi
+}
+
+# Default values
+export JAVA_TO_BUILD=jdk8u
+export PATH=/usr/local/bin/:$PATH
 export TARGET_OS=linux
-export VARIANT=openj9
```
sxa555: We should leave this as openj9 as it has the higher level of prereqs so should be the one to use unless there's a good reason not to

ansible/pbTestScripts/buildJDK.sh
```diff
@@ -54,13 +146,10 @@ echo "DEBUG:
         VARIANT=$VARIANT
         JDK7_BOOT_DIR=$JDK7_BOOT_DIR
         JAVA_HOME=$JAVA_HOME
-        WORKSPACE=$WORKSPACE"
+        WORKSPACE=$WORKSPACE
+	GIT_URL=$GIT_URL"
```
sxa555: Incorrect indentation

Willsparker: Sorted 👍

PR #1158, opened 2020-02-20
playbook: Update gcc-7 role to install gcc 7.5

sxa555:
This should resolve the failing CentOS8/SUSE12 playbook executions - issue introduced by https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1148


sxa555:
Forgot to to remove the conditional - now fixed.
The same GCC version is used for all architectures now which is why s390x does not require its own section.

PR #1168, opened 2020-02-21
Playbook to configure machines for TRSS

Haroon-Khel:
The `AdoptOpenJDK_Unix_Playbook/trss.yml` contains **most** of the manual steps to configure machines to run the TRSS service. 
The playbook currently only supports Ubuntu 18 on x86_64.
The manual steps that were omitted from the playbook are:

- `forever-service install TRSSFrontend -e "NODE_ENV=production" -f " --workingDir /home/jenkins/openjdk-test-tools/TestResultSummaryService" --script /home/jenkins/openjdk-test-tools/TestResultSummaryService/frontend.js  -o " --configFile=path-to-trssConf.json"`
- `forever-service install TRSSBackend -e "NODE_ENV=production" -f " --workingDir /home/jenkins/openjdk-test-tools/TestResultSummaryService" --script /home/jenkins/openjdk-test-tools/TestResultSummaryService/backend.js  -o " --configFile=/dev/mongodb/credentials/path-to-trssConf.json"`
- Credentials for the database
 
The forever commands were omitted because they require the trssConf.json file to exist, which will contain the database username and password. I thought this file, as well as the creation of those database credentials, could be best done manually.

The installation of `nginx` can be a bit finicky. Though this issue didnt appear on the new TRSS machine, when testing this playbook on a fyre machine the nginx service tries to start automatically upon installation, that is, it attempts to listen on port 80. If this port is in use, or for any other reason, this will cause an error and the playbook will crash.

Thoughts?

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
@@ -20,9 +20,30 @@
           apt: upgrade=safe update_cache=yes
           tags: patch_update
 
-        - name: Add Node.js v12.x
-          raw: "curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -"
-          tags: dependencies
+        - name: Download Nodev12.16.1 tarball
+          get_url:
+            url: https://nodejs.org/dist/v12.16.1/node-v12.16.1-linux-x64.tar.xz
+            dest: /tmp/node-v12.16.1-linux-x64.tar.xz
+            mode: 0440
+          tags: node_install
+        
+        - name: Extract Node tarball
+          unarchive:
+            src: /tmp/node-v12.16.1-linux-x64.tar.xz
+            dest: /usr/local/lib/
+            copy: False
+          tags: node_install
+
+        - name: Create symlinks for node, npm and npx from node install directory to /usr/local/bin/
+          file:
+            src: /usr/local/lib/node-v12.16.1-linux-x64/bin/{{ item }}
+            dest: /usr/local/bin/{{ item }}
+            state: link
+          with_items:
+            - node
+            - npm
+            - npx
+          tags: node_install
 
         - name: Import MongoDB 4.2 key
           raw: "wget -qO - https://www.mongodb.org/static/pgp/server-4.2.asc | sudo apt-key add -"
@@ -38,7 +59,6 @@
             update_cache: yes
           with_items:
             - git
-            - nodejs
             - mongodb-org
             - make
             - build-essential
@@ -101,7 +121,6 @@
             regex: 'root /var/www/html;'
             line: 'root /home/{{ Jenkins_Username }}/openjdk-test-tools/test-result-summary-client/build/;'
             state: present
-          when: nginx_installed.rc == 0
           tags: nginx_conf
 
         - name: Add server name trss.adoptopenjdk.net /etc/nginx/sites-enabled/default
@@ -110,7 +129,6 @@
             regex: 'server_name _;'
             line: 'server_name trss.adoptopenjdk.net;'
             state: present
-          when: nginx_installed.rc == 0
           tags: nginx_conf
 
         - name: Add Location /api/ to /etc/nginx/sites-enabled/default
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
```diff
@@ -0,0 +1,157 @@
...
+          hostname:
+            name: trss.adoptopenjdk.net
+          tags: hostname
+
+        - name: OS update -- apt-get upgrade
+          apt: upgrade=safe update_cache=yes
+          tags: patch_update
+
+        - name: Add Node.js v12.x
+          raw: "curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -"
```
sxa555: I'm not comfortable with using that way to install Node.
If we want it regularly patched with security releases via `apt-get upgrade` then I'd prefer us to use the ansible operations which add that repository manually instead of using that shell script from an unofficial source.
If we don't want it regularly updated, then grabbing it directly from the nodejs.org site would be preferable

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
```diff
@@ -0,0 +1,157 @@
...
+          shell: cd /home/{{ Jenkins_Username }}/openjdk-test-tools/TestResultSummaryService/ && npm ci
+
+        - name: Install nginx seperately
+          apt:
+            name: nginx
+            state: present
+          tags: nginx_conf
+          register: nginx_installed
+
+        - name: Add app root directorty to /etc/nginx/sites-enabled/default
```
sxa555: Typo `directorty`

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
```diff
@@ -0,0 +1,157 @@
...
+          lineinfile:
+            path: /etc/nginx/sites-enabled/default
+            regex: 'root /var/www/html;'
+            line: 'root /home/{{ Jenkins_Username }}/openjdk-test-tools/test-result-summary-client/build/;'
```
sxa555: I presume the username gets expanded ok in there with the single quotes around it?

Haroon-Khel: Yes. The saved line becomes `root /home/jenkins/openjdk-test-tools/test-result-summary-client/build/;`

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
Haroon-Khel: Those installation instructions were found from a link on the nodejs.org website, https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions-enterprise-linux-fedora-and-snap-packages

I cant seem to find a way to add the repository using 'ansibley', but an alternative could be to download and configure/make from source code. Would this be viable?

sxa555: No reason to compile from source, you can just pull down a tarball from e.g. https://nodejs.org/dist/v12.16.1/node-v12.16.1-linux-x64.tar.xz

Although something like https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml#L37 is an example of adding a third party apt repository

Haroon-Khel:
@sxa555 The unofficial nodejs source has been removed in favour of downloading the tarball from the nodejs website, as you suggested

Haroon-Khel:
Concerning any of the manual steps that have not been included in this playbook, I will add them to ansible/README.md under a new section dedicated to running this playbook

sxa555:
Might be worth printing out out a message at the end of the playbook with a list of extra steps (or detecting if they may have already been done if possible otherwise this looks ok - thanks. I'll merge. :-)

PR #1172, opened 2020-02-24
unixPB: Fix conditions to gmake

Willsparker:
ref: #1129 , https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1106#issuecomment-590349778

Fixing the conditionals added in from here. In the current situation, a system will run these roles if it is both SLES/SUSE12 or Ubuntu14 on x86_64, AND CentOS/RHEL6/7, which isn't possible.


sxa555:
Verifying at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/380/

Issue #697, opened 2019-02-11
Ubuntu 16.04 arm32 playbook issues


sxa555:
`openjdk-7-jdk` missing
`Chown /usr/lib/jvm/jdk-9.*` in `adoptopenjdk9` role doesn't work as directory is `/usr/lib/jvm/jdk-9+181`

sxa555:
No particularly high priority on this one Will, but if you want an excuse to play with a raspberry pi, this could be your chance :-)

sxa555:
ping @Willsparker - any update on this?

Willsparker:
Not yet - I'll prioritise this tomorrow when I have access to the pi :)

Willsparker:
Okay, I've managed to run through the playbook on Ubuntu1804 armv7l, with only the `adoptopenjdk_install` role being an issue as doesn't like the architecture being called `armv7l`. A fix for that will be to just add an additional role specifically for armv7l / armhf

The `chown` issue is no longer relevant as the `adoptopenjdk9` role isn't there anymore.

As for `openjdk-7-jdk` missing, that is unfortuantely still true, even when adding the `ppa:openjdk-r/ppa` repo. However, I'd ask is this entirely necessary as we're able to build with `JDK8` ? @sxa555

sxa555:
Yeah under the circumstances I'm ok with just JDK8 being on there (for Ubuntu 18+) I would normally expect `arm` as the descriptor to mean arm32

PR #1183, opened 2020-03-02
UnixPB: Arm specific adoptopenjdk_install task

Willsparker:
fixes: #697 

The issues listed on the #697  are no longer relevant to the current playbooks as the `adoptopenjdk9` role no longer exists, and we're not able to add the `openjdk-7-jdk` package as the `openjdk-r` repository doesn't have it available for `arm`, however we can build `jdk8` with `jdk8` so this isn't an issue.

 However, whilst running through the playbook to ensure it worked, the `adoptopenjdk_install` role doesn't work as the `ansible_architecture` variable is set as `armv7l`, not `arm` as APIv3 requires, so I've made an ARM specific role for it.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
@@ -66,7 +66,22 @@
     - ansible_os_family != "Darwin"
   tags: adoptopenjdk_install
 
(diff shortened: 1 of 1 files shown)


sxa555:
Could we set a api_architecture variable and use that instead to avoid duplication. Set it to ansible_architecture by default, and override if on arm32?

Willsparker:
Changes put in - this does still have duplication of tasks, but I agree that it's probably the better way to do it.

sxa555:
Testing at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/385/

Issue #685, opened 2019-01-24
Rework documentation about running ansible (esp.with Vagrant)


sxa555:
Related: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/379 https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/379 https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/423 https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/681 https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/656
New updates put in under https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1063 to at least make the information a bit more accurate but will likely need a second round of updates :-)

Willsparker:
Some updates put in:

https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/828

Willsparker:
@sxa555 asked that some documentation on how to install vagrant with KVM be put in.

Willsparker:
@karianna Any suggestions on additions or reworking, in particular? :-)

karianna:
> @karianna Any suggestions on additions or reworking, in particular? :-)

I need to go through it from scratch as a new user :-).  I'll try to get that this week.

sxa555:
Also included the basic execution in the new FAQ being added at https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1066

karianna:
@Willsparker I immediately get:

```
A host only network interface you're attempting to configure via DHCP
already has a conflicting host only adapter with DHCP enabled. The
DHCP on this adapter is incompatible with the DHCP settings. Two
host only network interfaces are not allowed to overlap, and each
host only network interface can have only one DHCP server. Please
reconfigure your host only network or remove the virtual machine
```

Willsparker:
@karianna Is this following the instructions in `../ansible/README.md` ? I can't say I've seen that error message before, was it on `vagrant up` ?

karianna:
@Willsparker Yep on the vagrsant up...

Willsparker:
https://github.com/hashicorp/vagrant/issues/3083 

At a guess, this may be the cause - if it's a fresh install of Virtualbox and no `vagrant up`s have been ran before on your machine? 
The issue says that running 
[...]

PR #1184, opened 2020-03-03
ansible: Refactor ansible/README.md Vagrant section

Willsparker:
fixes: #685 

Changes to add sections for running ansible remotely, pointing at a Linux Vagrant VM. I also moved & overhauled the section about running ansible remotely pointing at a Windows Vagrant VM to under the `Running via Vagrant and Virtualbox` section.


sxa555:
We do need to ensure that we have information in there that is valid for the use case of a user who wants to run against a real windows system (i.e. not Vagrant)

Willsparker:
The latest commit comes from suggestions made by @M-Davies going through the updated `README.md`

PR #1167, opened 2020-02-21
winPB: Switch adoptopenjdk download to use APIv3

sxa555:
Signed-off-by: Stewart Addison <sxa@uk.ibm.com>

PR #1185, opened 2020-03-05
WinPB: Make Java8 role use APIv3

Willsparker:
Ref: #1167 

Ensuring Java8 uses APIv3 as well

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
@@ -16,7 +16,7 @@
 
 - name: Download Java 8
   win_get_url:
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml
```diff
@@ -16,7 +16,7 @@
 
 - name: Download Java 8
   win_get_url:
-    url: https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl={{ bootjdk }}&os=windows&arch=x64&release=latest&type=jdk&heap_size={{ heapsize }}
+    url: https://api.adoptopenjdk.net/v3/binary/latest/8/ga/windows/x64/jdk/{{ bootjdk }}/normal/adoptopenjdk?project=jdk
```
karianna: Is changing behavior from heap size var to normal deliberate?

Willsparker: It was as when `bootjdk = hotspot` and `heapsize = large`, the API returns nothing - however I've just seen that we don't have Large heapsize for hotspot, so that use-case wouldn't happen. I'll add it back in :)

PR #1187, opened 2020-03-05
pbTests: Update buildJDKWin to specify which JDK to build

Willsparker:
Update to allow the VagrantPlaybookCheck Jenkins job to specify which jdk we want to build for both Unix and Windows, with defaults to JDK8 - OpenJ9.

### DIFF
--- a/ansible/pbTestScripts/vagrantPlaybookCheck.sh
+++ b/ansible/pbTestScripts/vagrantPlaybookCheck.sh
@@ -13,6 +13,7 @@ vmHalt=true
 cleanWorkspace=false
 newVagrantFiles=false
 skipFullSetup=''
+jdkToBuild=''
 
 # Takes all arguments from the script, and determines options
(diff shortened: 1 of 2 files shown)


ansible/pbTestScripts/vagrantPlaybookCheck.sh
```diff
@@ -56,6 +59,8 @@ usage()
 					--all | -a 				Builds and tests playbook through every OS
 					--retainVM | -r				Option to retain the VM and folder after completion
 					--build | -b				Option to enable testing a native build on the VM
+					--jdkToBuild | -jdk			Specify which JDK to build, if build is specified
```
sxa555: Isn't the parameter `--JDK-Version`?

Willsparker: 🤦‍♂

PR #1197, opened 2020-03-12
pbTests: Add option for Hotspot & improve JDK_BOOT_DIR detection

Willsparker:
Beforehand, both `buildJDKwin.sh` and `buildJDK.sh` had the ability to build with Hotspot, however this wasn't usable by the VagrantPlaybookCheck Jenkins job as it wasn't be passed through to them via `vagrantPlaybookCheck.sh`.

I also noticed that when building a JDK version other than 8 on linux, the `JDK_BOOT_DIR` was just being set as `JAVA_HOME` which was being set to JDK8 to run Gradle. I've improved the logic so it will first look for JDK_VERSION-1 , and if it can't find it, it will settle for JDK_VERSION. With the playbooks installing 8,10,12 and 13, this should suffice to build them all.

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -59,8 +59,8 @@ showJDKVersions() {
 }
 
 checkJDKVersion() {
-        local jdk=$1
-        case "$jdk" in
+	local jdk=$1
+	case "$jdk" in
                 "jdk8u" | "jdk8" | "8" | "8u" )
                         JAVA_TO_BUILD="jdk8u";;
                 "jdk9u" | "jdk9" | "9" | "9u" )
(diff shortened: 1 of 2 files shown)


ansible/pbTestScripts/buildJDK.sh
```diff
@@ -77,7 +77,26 @@ checkJDKVersion() {
...
+	esac
+	setBootJDK 
+}
+
+setBootJDK() {
+	local buildJDKNumber=$(echo ${JAVA_TO_BUILD//[!0-9]/})
+	local bootJDKNumber=$(($buildJDKNumber - 1));
+        # if building JDK8/9 look for 'jdk8u', not 'jdk-x'
+	if [[ $buildJDKNumber -eq 8 || $buildJDKNumber -eq 9 ]]; then
+		export JDK_BOOT_DIR=$(find /usr/lib/jvm -maxdepth 1 -name *jdk8*)
```
sxa555: You seem to have added a fixed `jdk8` search immediately after checking if it's either 8 or 9. To be honest, I'd ignore 9 entirely now

Willsparker: I did that specifically because the playbooks don't install `jdk9`, so the only other option to build jdk9 is going to be jdk8 - but fair enough - I know it's out of support but I thought if the option to make it able to build is there, we may as well put it in :-)

PR #1200, opened 2020-03-13
pbTests: Add options to specify infra repo/branch for qemu script

Willsparker:
Useful change so we can test new changes to the playbooks on different architectures, requested by @sxa555

### DIFF
--- a/ansible/pbTestScripts/qemu_test_script.sh
+++ b/ansible/pbTestScripts/qemu_test_script.sh
@@ -4,6 +4,8 @@
 
 ARCHITECTURE=""
 skipFullSetup=""
+gitURL="https://github.com/adoptopenjdk/openjdk-infrastructure"
+gitBranch="master"
 PORTNO=10022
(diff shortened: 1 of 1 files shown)


ansible/pbTestScripts/qemu_test_script.sh
```diff
@@ -120,7 +129,7 @@ local workFolder="$WORKSPACE/qemu_pbCheck"
 
 # Find/stop port collisions
 # while ps -aux | grep "$PORTNO" | grep -q -v "grep"; do
-while netstat -lp 2>/dev/null | grep "tcp.*:$PORTNO " > /dev/nulln; do
+while netstat -lp 2>/dev/nulln | grep "tcp.*:$PORTNO " > /dev/null; do
```
sxa555: nulln? You've switched them around but that seems wrong regardless

Willsparker: typo- my bad

Issue #1210, opened 2020-03-18
Cache installed cygwin in order to speed up Windows provisions


sxa:
I believe the cygwin installation is no more than splatting a load of files into `C:\cygwin64`. This is one of the slowest parts of the Windows playbooks. We should prototype storing the contents of that directory somewhere and extracting it when we perform a playbook deployment.

sxa:
The change merged was backed out in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1231 - reopening this issue

Willsparker:
I've had another crack at trying what I did last time -  `wget` the archive from `ci.adoptopenjdk.net/userContent/` and untar to `C:/cygwin64`. While that works generally, (i.e. I can access the terminal `mintty` and do basic operations ), attempting to `git clone https://github.com/adoptopenjdk/openjdk-infrastructure` returns the same error as when I tried it before:
```
fatal: unable to access 'https://github.com/adoptopenjdk/openjdk-infrastructure/': SSL: couldn't create a context: error:0E079065:configuration file routines:def_load_bio:missing equal sign
```
Reinstallation of `git` via `setup-x86_64.exe` (The installer for Cygwin) unfortunately didn't fix this issue either.

Willsparker:
Following https://stackoverflow.com/questions/11744472/how-to-replicate-a-cygwin-install-and-environment-on-another-machine unfortunately didn't give much luck either - same `git` issue. I'm going to get a fresh installation of `Cygwin` and repeat the things I did above, on the off-chance that it's the archive that is the issue.

AdamBrousseau:
Does the machine have Git for Windows installed? Is it first on the path? If so, try removing it.

Willsparker:
It doesn't have it installed - I'm using a clean machine at the minute as the Cygwin installation shouldn't need anything else :-)

Willsparker:
Judging from the error message, It is an issue with OpenSSL, not git. Therefore, I tried to run a basic SSL command to see if the error message given was anymore helpful (following a guide [here](https://www.poftut.com/use-openssl-s_client-check-verify-ssltls-https-webserver/)):
```
$ openssl s_client -connect poftut.com:443
34359738384:error:0E079065:configuration file routines:def_load_bio:missing equal sign:crypto/conf/conf_def.c:391:line 40
[...]

PR #1219, opened 2020-03-20
winPB: Update Cygwin to use a prebuilt folder

Willsparker:
fixes: #1210 

Fix-up the playbooks to use https://ci.adoptopenjdk.net/userContent/winansible/ to significantly reduce time of Cygwin installation.
Testing at:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/425/OS=Win2012,label=infra-vagrant-1/console

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -39,6 +39,7 @@
     - Jenkins                     # AdoptOpenJDK infrastructure
     - MSVS_2010
     - VS2010_SP1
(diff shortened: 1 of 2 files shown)


sxa555:
Are we confident that this is working? The linked check job looks to have failed

Willsparker:
The linked job succeeded with the Cygwin part of the playbook - that particular run I only wanted to test the playbooks, and it comes up as failed due to a weird Jenkins Agent issue (which I hope has been fixed by being restarted today)

sxa555:
We do need to be absolutely certain that the installation works - it would be hard for a simple unzip to fail :-)

PR #1224, opened 2020-03-23
Ansible pulls from userContent/winansible

Haroon-Khel:
On the windows playbook, I have changed download source of the openssl and llvm 32bit from `userContent` to `userContent/winansible` (https://ci.adoptopenjdk.net/userContent/winansible/), as per https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1215

The source for the cygwin download was already set to the winansible directory by a previous commit

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
@@ -16,7 +16,7 @@
     url: https://ci.adoptopenjdk.net/userContent/winansible/llvm-7.0.0-win64.zip
     dest: 'C:\temp\'
     force: no
-    checksum: 74b197a3959b0408adf0824be01db8dddfa2f9a967f4085af3fad900ed5fdbf6
+    checksum: e61c3d115778dbb54f91cec8ca7682cbad601d61e2c40a9a97e554aa1f1bb2b6
     checksum_algorithm: sha256
   when: not clang_64bit_installed.stat.exists
   tags: clang_64bit
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
@@ -3,6 +3,7 @@
 # LLVM/Clang 64bit #
 ####################
 # Clang 64bit is an OpenJ9 prerequisite.
+# The original source of Clang64 was https://releases.llvm.org/7.0.0/LLVM-7.0.0-win64.exe being executed with the /S parameter
 
 - name: Test if Clang 64bit is already installed (required by OpenJ9)
   win_stat:
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
@@ -6,7 +6,7 @@
 
 - name: Test if Clang 64bit is already installed (required by OpenJ9)
   win_stat:
-    path: 'C:\Program Files\LLVM\bin\clang.exe'
+    path: 'C:\Program Files\LLVM64\bin\clang.exe'
   register: clang_64bit_installed
   tags: clang_64bit
 
@@ -20,8 +20,11 @@
   when: not clang_64bit_installed.stat.exists
   tags: clang_64bit
 
-- name: Install Clang 64bit
-  win_command: 'c:\temp\llvm-7.0.0-win64.exe /S'
+- name: Install (unzip) Clang 64bit
+  win_unzip:
+    src: C:\temp\llvm-7.0.0-win64.zip
+    dest: C:\
+    creates: 'C:\Program Files\LLVM64\bin\clang.exe'
   when: not clang_64bit_installed.stat.exists
   tags: clang_64bit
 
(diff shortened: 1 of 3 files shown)


sxa555:
@Willsparker Can you run this through a playbookcheck job when you find a space in the schedule please?

Haroon-Khel:
I have just noticed that the llvm 64 bit task in the playbook downloads from https://releases.llvm.org/7.0.0/LLVM-7.0.0-win64.exe. Ill put in a change so that it uses https://ci.adoptopenjdk.net/userContent/winansible/llvm-7.0.0-win64.zip instead

sxa555:
The reason why it was done differently is covered in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/514#issuecomment-440270003 ... If you run the [32|64]-bit installer after installing the other one the it uninstalls the previous one, so we installed one of them via zip file.

In theory we could unzip both, and I suppose for consistency I'd be in favour of that assuming the 64-bit zip file in there works ok - A run of VagrantPlaybookCheck will be able to check that.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
```diff
@@ -21,7 +21,7 @@
   tags: clang_64bit
 
 - name: Install Clang 64bit
-  win_command: 'c:\temp\LLVM-7.0.0-win64.exe /S'
+  win_command: 'c:\temp\llvm-7.0.0-win64.exe /S'
```
sxa555: This needs to be an unzip operation on the zip file now that the download has been changed

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
```diff
@@ -38,7 +38,7 @@
 
 - name: Cleanup Clang 64bit installer
   win_file:
-    path: 'c:\temp\LLVM-7.0.0-win64.exe'
+    path: 'c:\temp\llvm-7.0.0-win64.exe'
```
sxa555: As above, this should clean up a `.zip` file not the `.exe`

Willsparker:
Running through VPC :-) https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/433/

Willsparker:
Rerunning - I set the wrong parameters 🤦‍♂ 
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/435/

sxa555:
I'm rather confused as to why (Since I'm pretty sure I just moved the file) but https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=Win2012,label=infra-vagrant-1/440/console is saying that the checksum is wrong on the 64-bit LLVM download ...

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
```diff
@@ -3,25 +3,29 @@
...
   win_stat:
-    path: 'C:\Program Files\LLVM\bin\clang.exe'
+    path: 'C:\Program Files\LLVM64\bin\clang.exe'
   register: clang_64bit_installed
   tags: clang_64bit
 
 - name: Download Clang 64bit
   win_get_url:
-    url: https://releases.llvm.org/7.0.0/LLVM-7.0.0-win64.exe
+    url: https://ci.adoptopenjdk.net/userContent/winansible/llvm-7.0.0-win64.zip
```
Willsparker: The checksum will need to be changed as well :-)

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
```diff
@@ -3,25 +3,29 @@
...
     force: no
     checksum: 74b197a3959b0408adf0824be01db8dddfa2f9a967f4085af3fad900ed5fdbf6
     checksum_algorithm: sha256
   when: not clang_64bit_installed.stat.exists
   tags: clang_64bit
 
-- name: Install Clang 64bit
-  win_command: 'c:\temp\LLVM-7.0.0-win64.exe /S'
+- name: Install (unzip) Clang 64bit
+  win_unzip:
```
Willsparker: I would be concerned about using `win_unzip`. The module requires `PSCX` to be installed which can be done using the `win_psmodule`, however that requires further installation of Powershell modules to work, which I couldn't get working with our vagrantfiles. 
https://docs.ansible.com/ansible/latest/modules/win_unzip_module.html#requirements

Willsparker:
> I'm rather confused as to why (Since I'm pretty sure I just moved the file) but https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=Win2012,label=infra-vagrant-1/440/console is saying that the checksum is wrong on the 64-bit LLVM download ...

Because now we're downloading the zip folder from UserContent, not the installer from the LLVM website

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
Haroon-Khel: I used `win_unzip` since some of the other roles in the windows playbook used `win_unzip`, like the `freemarker` and `Java_install`. Are there other ways of installing the `PSCX` module without using `win_psmodule`?

Haroon-Khel:
I'll change the checksum accordingly.
@sxa555 
I switched it to avoid confusion with the install directory of clang32. But after unzipping the .zip for test purposes, I see that it creates a `/Program Files/LLVM` automatically

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_64bit/tasks/main.yml
Willsparker: Ah, my mistake. It may've just been the situation I was using it, but I haven't seen any issues with either of those roles before regarding it, so go ahead :-)

Haroon-Khel:
The checksum for the Clang64 download has been updated
@Willsparker Could you run this through your playbook check job to see if this update works?

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=Win2012,label=infra-vagrant-1/446/console

Willsparker:
The playbook works- annoyingly it wasn't able to build due to an error that's fixed in #1231 , but the Playbook ran all the way through with no issues.

PR #1225, opened 2020-03-24
UnixPB: Make nasm pickup gcc7.5, revert opensuse gcc4.3 install task

Willsparker:
Issues noticed when running the VagrantPlaybookCheck job.

nasm was looking for gcc7.3, which has been taken out of the playbooks since #1158 ,
The OpenSUSE role had to be reverted back from #1190 , as replacing the `shell` module for the `zypper` module means the `--force-resolution` argument can't be put in. The `skip_ansible_lint` tag has been put in, with a comment for justification.


karianna:
@Willsparker change looked good but Travis failures?

Willsparker:
My bad :) should be good now

PR #1230, opened 2020-03-26
UnixPB: Add role for riscv building

Willsparker:
ref: https://github.com/eclipse/openj9/issues/7421

This adds a role that installs all the additional tools necessary to build JDK11/J9 for riscv64. Currently, the tasks have the `adoptopenjdk` tags on them as this isn't necessarily wanted as standard on the machines. This roles is also only for CentOS x86_64 machines, as the `riscv_toolchain_linux` was built for CentOS. 

The role is currently pulling from https://ci.adoptopenjdk.net/userContent/riscv/ to get the prebuilt toolchain and qemu fedora `sysroot` folder. The UserContent folder also includes instructions on how to setup a machine manually, and start that build. 

Given the tag, this role can't be tested on `VagrantPlaybookCheck`, however I manually tested on a CentOS6 VM made from our Vagrantfiles and was able to build the JDK :-)


sxa555:
@Willsparker Have you run this through VPC?

Willsparker:
@sxa555 I can't test it through VPC, it's got the `adoptopenjdk` tag- I tested it locally :-)

PR #1231, opened 2020-03-26
[wip] WinPB: Revert Cygwin cache, pin Make to v4.2.1-2

Willsparker:
I've taken the Cygwin cache PR ( #1219 ) out, as there were some issues with Git : 

```
Cloning into 'C://openjdk-build'...
fatal: unable to access 'https://github.com/adoptopenjdk/openjdk-build/': SSL: couldn't create a context: error:0E079065:configuration file routines:def_load_bio:missing equal sign
``` 
When installed normally, this issue doesn't occur.

I've also pinned Cygwin's Make version to `4.2.1-2` as there was an issue building JDK11 with Make version 4.3 : https://www.mail-archive.com/bug-make@gnu.org/msg12006.html
which causes this issue: 

```
make[3]: *** No rule to make target '/cygdrive/c/workspace/build/src/build/windows-x86_64-normal-server-release/buildtools/langtools_tools_classes/_the.BUILD_TOOLS_LANGTOOLS.vardeps', needed by '/cygdrive/c/workspace/build/src/build/windows-x86_64-normal-server-release/buildtools/langtools_tools_classes/_the.BUILD_TOOLS_LANGTOOLS_batch'.  Stop.
```

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -22,6 +22,7 @@
     force: no
     checksum: 4a3fc472b6051d8ec136831e64b9b4bbd328b587d66eef1eedf2a58a620bde97
     checksum_algorithm: sha256
+  when: (not cygwin_installed.stat.exists)
   tags: cygwin
 
 - name: Install Cygwin and its packages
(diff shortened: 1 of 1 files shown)


Willsparker:
Put back into work in progress as a VPC run has shown that make isn't getting installed properly.

Willsparker:
I should probably mention, the issue I'm having is
```
172.28.128.254 | FAILED | rc=1 >>
C:/cygwin64/bin/make.exe: error while loading shared libraries: cygguile-2.0-22.dll: cannot open shared object file: No such file or directory
```
That's after I had to manually extract the folder, so I'll have to redo that task as well

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/450/OS=Win2012,label=infra-vagrant-1/console

Can confirm this works - I know the run didn't succeed, but it was able to fully run the playbook and build JDK11 - I'll just need to tweak `testJDKWin.sh`. Merging

PR #1237, opened 2020-03-30
pbTests: Test script fixes

Willsparker:
Additional fixes to :
1) stop `testJDK.sh` finding the `debug-image` thats produced when building JDKs

2) Allow `testJDKWin.sh` to find built JDKs other than JDK8

### DIFF
--- a/ansible/pbTestScripts/buildJDKWin.sh
+++ b/ansible/pbTestScripts/buildJDKWin.sh
@@ -115,14 +115,13 @@ export VARIANT=openj9
 export PATH=/usr/bin/:$PATH
 export TARGET_OS=windows
 export ARCHITECTURE=x64
-export JAVA_HOME=/cygdrive/c/openjdk/jdk-8
 export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-8
 GIT_URL=https://github.com/adoptopenjdk/openjdk-build
 CLEAN_WORKSPACE=false
 
 processArgs $*
 cloneRepo
-
+export JAVA_HOME=/cygdrive/c/openjdk/jdk-8
 echo "DEBUG:
 	TARGET_OS=$TARGET_OS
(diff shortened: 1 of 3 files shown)


Willsparker:
Latest commit ensures JDK8 is built with JDK8 on Windows. When running the build with JDK_BOOT_DIR set to jdk7, it attempts to run Gradle with it, causing Gradle to fail. 
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/452/OS=Win2012,label=infra-vagrant-1/console

ansible/pbTestScripts/buildJDKWin.sh
```diff
@@ -63,7 +63,7 @@ checkJDKVersion() {
         case "$jdk" in
                 "jdk8u" | "jdk8" | "8" | "8u" )
                         JAVA_TO_BUILD="jdk8u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk7;;
```
sxa555: Why is this being changed from 7 to 8?

karianna:
What version of gradle dropped 7 support and do we need to execute gradle as part of our boot strapping?

sxa555:
Normally JAVA_HOME is set to 8 (or it's in the PATH by default) and we just set the JDK boot dir to the boot one, which should be 7. We don't want to change JDK8 to bootstrap with 8 - it's not what we do in the production builds, and so it would be wrong to do that here.

Willsparker:
Okay, I'll put it back and just ensure `JAVA_HOME` is jdk8

sxa555:
Can you do a VPC run to ensure this is ok @Willsparker ?

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/456/
I'll do a jdk8 run now too

Issue #1239, opened 2020-03-31
Replace win2008 machines with win2012


gdams:
Once https://github.com/AdoptOpenJDK/openjdk-build/pull/1638 lands we will no longer need win2008R2 build machines. 

My proposal is that we replace them with two new 2012 machines so that we have a larger pool of 2012.

Steps:

- [x] Merge https://github.com/AdoptOpenJDK/openjdk-build/pull/1638
- [x] Spin up 2 new 2012 machines (leave 2008 running for now)
- [x] Destroy old Softlayer JCK machine
- [x] Run Ansible on them and add to jenkins
- [x] Ensure that builds run on them
- [x] Remove 200R2 machines are a few weeks

gdams:
We do also have a JCK windows machine that we should re-purpose at the same time (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/inventory.yml#L190)

Willsparker:
I'm happy to run the playbooks on them and confirm builds run on them 👍

gdams:
@tellison has confirmed that the JCK machine still exists so I've requested that he re-provision it as a fresh image

tellison:
Have we checked with the OpenJ9 folks about their build requirements on Windows?

gdams:
@tellison all of the OpenJ9 builds are running on win2012 the only thing that was still using 2008 is JDK8u Hotspot

gdams:
two Azure 2012R2 machines have been handed to @Willsparker

gdams:
https://ci.adoptopenjdk.net/computer/build-azure-win2012r2-x64-1/
https://ci.adoptopenjdk.net/computer/build-azure-win2012r2-x64-2/

both added to jenkins

tellison:
The old Softlayer JCK machine has been decommissioned, and a shiny new test machine provisioned in its place.  Name is `test-win2012-2.adoptopenjdk.net` and details sent privately to @gdams in the first instance to share with infra team.

Willsparker:
The 2 new `build-azure-2012r2-x64` machines have had the playbook ran on them, and are connected to Jenkins

gdams:
Installer jobs failing with:

```bash
Expand-Archive : The term 'Expand-Archive' is not recognized as the name of a 
cmdlet, function, script file, or operable program. Check the spelling of the 
name, or if a path was included, verify that the path is correct and try again.
```

[...]

PR #1250, opened 2020-04-03
WinPB: Add playbook role to set shortpaths in 'Program Files (x86)'

Willsparker:
Ref: #1239 

When setting up the Azure machines, `short names` weren't enabled on the `C:/` drive, which stopped builds as it didn't like the spaces in directories such as `Microsoft Visual Studio`.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -67,3 +67,4 @@
     - Rust                        # IcedTea-Web
     - IcedTea-Web                 # For Jenkins webstart
     - WiX                         # For creating installers
+    - Shortnames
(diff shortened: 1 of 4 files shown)


sxa555:
Under what circumstances are these not being created, and what do we have relying on them?

Willsparker:
> Under what circumstances are these not being created, and what do we have relying on them?

They weren't created when @gdams and I setup the `azure-2012r2` machines, causing builds to fail. E.g: JDK11-HS x64 build yesterday ( https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-windows-x64-hotspot/496/console ):
```
checking for mt... mt
configure: Rewriting MT to "/cygdrive/c/progra~2/windows kits/10/bin/10.0.17763.0/x86/mt"
checking for rc... rc
configure: Rewriting RC to "/cygdrive/c/progra~2/windows kits/10/bin/10.0.17763.0/x86/rc"
```
As there was no shortpath, the space remained, causing this later error:

```
cl : Command line warning D9024 : unrecognized source file type 'kits/netfxsdk/4.6.1/include/um', object file assumed
cl : Command line warning D9027 : source file 'kits/netfxsdk/4.6.1/include/um' ignored
```
I believe this is just due to the difference in Images that certain machine providers given us- i.e. our Vagrantfiles create Windows VMs that already have it enabled.

PR #1255, opened 2020-04-06
winPB: Add WMF install role

Willsparker:
ref: #1239 

Updates Powershell to `v5.1`. Required for the `Expand-Archive` cmdlet.


sxa555:
For reference, this is required because the scripts in [openjdk-installer](https://github.com/adoptopenjdk/openjdk-installer) use `Expand-Archive`

PR #1258, opened 2020-04-06
winPB: Change Cygwin checksum

Willsparker:
It's changed again.

@sxa555 How'd you feel about taking the checksum off the Cygwin Setup?


sxa555:
> @sxa555 How'd you feel about taking the checksum off the Cygwin Setup?

"Somewhat mixed" :-) But them my preference is to try and figure out how to make a cached install work, so in the meantime I'll let this through as-is. Bear in mind we have the same issue with at least one of the Visual Studio installers too

Willsparker:
> "Somewhat mixed" :-) But them my preference is to try and figure out how to make a cached install work, so in the meantime I'll let this through as-is. Bear in mind we have the same issue with at least one of the Visual Studio installers too

I was worried you'd say that 😂 I checked the MSVS_2017 installer and that's fine- I think it was just a cygwin update

PR #1263, opened 2020-04-07
pbTest: set correct params running buildJDKWin w/JDK8

Willsparker:
When running VagrantPlaybookCheck Jenkins job and building JDK8, the Windows VM will fail as it attempts to run Gradle with jdk-7, as that was `JDK_BOOT_DIR` was set to. I've found that setting the `JDK8_BOOT_DIR` variable will run Gradle w/ JDK8, but keep the `JDK_BOOT_DIR` variable at jdk-7, thereby simulating what we do on our Jenkins build jobs.

### DIFF
--- a/ansible/pbTestScripts/buildJDKWin.sh
+++ b/ansible/pbTestScripts/buildJDKWin.sh
@@ -116,8 +116,6 @@ export PATH=/usr/bin/:$PATH
 export TARGET_OS=windows
 export ARCHITECTURE=x64
 export JAVA_HOME=/cygdrive/c/openjdk/jdk-8
-export JDK8_BOOT_DIR=/cygdrive/c/openjdk/jdk-8
-export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-8
 GIT_URL=https://github.com/adoptopenjdk/openjdk-build
 CLEAN_WORKSPACE=false
 
@@ -129,7 +127,6 @@ echo "DEBUG:
 	JAVA_TO_BUILD=$JAVA_TO_BUILD
         VARIANT=$VARIANT
         JDK_BOOT_DIR=$JDK_BOOT_DIR
-        JDK8_BOOT_DIR=$JDK8_BOOT_DIR
         JAVA_HOME=$JAVA_HOME
         WORKSPACE=$WORKSPACE
         GIT_URL=$GIT_URL"	
--- a/ansible/pbTestScripts/buildJDKWin.sh
+++ b/ansible/pbTestScripts/buildJDKWin.sh
@@ -63,25 +63,25 @@ checkJDKVersion() {
         case "$jdk" in
                 "jdk8u" | "jdk8" | "8" | "8u" )
                         JAVA_TO_BUILD="jdk8u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk7;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk7;;
                 "jdk9u" | "jdk9" | "9" | "9u" )
                         JAVA_TO_BUILD="jdk9u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-8;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-8;;
                 "jdk10u" | "jdk10" | "10" | "10u" )
                         JAVA_TO_BUILD="jdk10u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-10;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-10;;
                 "jdk11u" | "jdk11" | "11" | "11u" )
                         JAVA_TO_BUILD="jdk11u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-10;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-10;;
                 "jdk12u" | "jdk12" | "12" | "12u" )
                         JAVA_TO_BUILD="jdk12u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-12;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-12;;
                 "jdk13u" | "jdk13" | "13" | "13u" )
                         JAVA_TO_BUILD="jdk13u";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-12;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-12;;
                 "jdk14u" | "jdk14" | "14" | "14u" )
                         JAVA_TO_BUILD="jdk14";
-			JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-13;;
+                        export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-13;;
                 *)
                         echo "Not a valid JDK Version" ; showJDKVersions; exit 1;;
         esac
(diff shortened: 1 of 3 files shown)


sxa555:
Does it not work with `JAVA_HOME` set to the JDK8 directory?

Willsparker:
Unfortunately not - I wasn't sure why, either.

sxa:
OK so the home dir used by gradle is set from `JDK8_BOOT_DIR` as a preference and `JAVA_HOME` if that isn't used (See the logic in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L397) - would be good to double check whether we can do this without `JDK8_BOOT_DIR` is set and understand what is resetting `JAVA_HOME` if it isn't working with just that

Willsparker:
Removed `JDK8_BOOT_DIR` to allow for testing with `VPC` with the latest commit. 
Run through:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/487/

Willsparker:
@sxa Unfortunately, Gradle is still trying to run with jdk-7. 

I narrowed it down to why: https://github.com/AdoptOpenJDK/openjdk-build/blob/f65ae15988a92a7f33d4c8fad6d6029f6b983615/build-farm/platform-specific-configurations/windows.sh#L24

Willsparker:
Ref: The latest commit ( regarding the `FILENAME` variable ) 
Recently build JDK11/J9 on Windows, the variant hasn't been detected as openj9 (therefore defaulting to hotspot) and cloning the incorrect source repo, which has been causing some config arguments to not be accepted.
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/486/OS=Win2012,label=infra-vagrant-1/console

This was due to no name being passed into the `target-file-name` option, causing all arguments afterwards to be disregarded:

```
/cygdrive/c/openjdk-build/build-farm/../makejdk-any-platform.sh --clean-git-repo ... --with-toolchain-version=2017 --target-file-name  --freetype-version 2.5.3 --use-jep319-certs --build-variant openj9 jdk11u
Starting /cygdrive/c/openjdk-build/build-farm/../makejdk-any-platform.sh to configure, build (Adopt)OpenJDK binary
Parsing opt: --clean-git-repo
...
Parsing opt: --target-file-name
Possible opt arg: --freetype-version
Working dir is ./build/
```

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/489/ - JDK8 run
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/490/ - JDK11 run

Willsparker:
The latest commit is required for the JDK8/J9 builds. When Windows is building JDK8/J9, the following is added to the configure arguments: https://github.com/AdoptOpenJDK/openjdk-build/blob/995c157903452e2c10751ab196567f281ff7fcd4/sbin/build.sh#L319
As '$WORKSPACE' defaults to `C:/`, the variable `${addsDir}` ends up being `/cygdrive/c/cygwin64//cygdrive/c/workspace/./build//src/closed/adds`, which is _very_ wrong. 
Changing it to `/tmp` should alleviate this, as well as replicate what we do in the official builds.

Rerun of JDK8/J9 : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/491/

sxa:
> @sxa Unfortunately, Gradle is still trying to run with jdk-7.
> 
> I narrowed it down to why: https://github.com/AdoptOpenJDK/openjdk-build/blob/f65ae15988a92a7f33d4c8fad6d6029f6b983615/build-farm/platform-specific-configurations/windows.sh#L24
> 
> Well- that's the point that `$JAVA_HOME` is overwritten to `$JDK_BOOT_DIR`. Not sure why the builds at ci.adoptopenjdk.net doesn't do this, but I did note that they run Gradle with jdk-11, not 8... Again, not really sure why
> 
> edit: https://github.com/AdoptOpenJDK/openjdk-build/blob/6ce704f7ba1cf3dfa1e440a5eba36f06693cab2b/sbin/build.sh#L410
> 
> That's why the builds are using jdk11- they're setting `$JDK11_BOOT_DIR`, as far as I can see, which has higher priority than `$JAVA_HOME` and `$JDK8_BOOT_DIR`. At this point, I think it's fair to say that setting `JDK8_BOOT_DIR` in my script isn't incorrect..

@johnoliver Do you know why we set `JAVA_HOME=$JDK_BOOT_DIR` [on Windows](https://github.com/AdoptOpenJDK/openjdk-build/blame/f65ae15988a92a7f33d4c8fad6d6029f6b983615/build-farm/platform-specific-configurations/windows.sh#L24)? - @Willsparker I'm tempted to suggest trying to remove that line and run a VPC test build with it for JDK8 and JDK11 and see if it causes any other issues

Willsparker:
> @Willsparker I'm tempted to suggest trying to remove that line and run a VPC test build with it for JDK8 and JDK11 and see if it causes any other issues

I'll start them off once my current runs are finished (I'll have to remove the `JDK8_BOOT_DIR `variable from this PR to test that the line change actually fixes it).
If it's of any bearing, removing that line allowed me to start a build properly on a Windows VM on my laptop

johnoliver:
It may have been set from a while ago, as gradle detects the java dir separately I dont think it will need JAVA_HOME set here, so can probably be removed safely

sxa:
@johnoliver We force the gradle home dir in getGradleHome() - but `JAVA_HOME` is one of the variables it sources from, which unfortunately means that if you're set up with `JAVA_HOME` pointing to java 7 in order to bootstrap a JDK8 build (which is what that problematic line in `windows.sh` is doing) it fails. Hopefully it's no longer needed, and we can leave `JAVA_HOME` as it was.

sxa:
@Willsparker I couldn't see any run that you'd kicked off for this, so I've got https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/494/ queued up

Willsparker:
Reviewing the stuff that ran last night - JDK11 seems okay, when not setting JAVA_HOME to JDK_BOOT_DIR (Thanks @sxa) - it failed on the test (intermittent problem) , but it did the build fine.
JDK8 is still an issue with the `add-source-root` option - its still putting it at `/cygdrive/c/cygwin64//cygdrive/c/workspace/./build//src/closed/adds` despite setting the `$WORKSPACE` variable (and starting the build) to `/tmp`. This may require a change in `vagrantPlaybookCheck.sh`

Willsparker:
Re-runs:
 https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/495/
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/496/.  (JDK11, not massively required for the most recent change but if I can get it green, I'd prefer it to be)

johnoliver:
If you have no luck with those current runs, reordering the order of preference for detecting the gradle java, so that JDK8_BOOT_DIR/JDK11_BOOT_DIR have higher priority would probably be fine.

Willsparker:
> If you have no luck with those current runs, reordering the order of preference for detecting the gradle java, so that JDK8_BOOT_DIR/JDK11_BOOT_DIR have higher priority would probably be fine.

Thanks! I believe that is already the case:
 https://github.com/AdoptOpenJDK/openjdk-build/blob/995c157903452e2c10751ab196567f281ff7fcd4/sbin/build.sh#L397

I think @sxa was saying that he'd prefer those weren't necessarily set in my `buildJDKWin.sh` script as the only variable that _should_ be setting `gradleJavaHome` is `JAVA_HOME` - in a perfect world where that wouldn't ruin the usual build jobs :-)

Willsparker:
Alright, the `add-source-root` path is still messing me up..
@johnoliver What does `add-source-root` do? It keeps finding the wrong path - `--with-add-source-root=/cygdrive/c/cygwin64//cygdrive/c/tmp/workspace/./build//src/closed/adds ` when I started the build at `/tmp`

https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/495/OS=Win2012,label=infra-vagrant-1/console

sxa:
Hi @johnoliver Can you comment on the final comment above please - there isn't much description in the original PR (https://github.com/AdoptOpenJDK/openjdk-build/pull/635) as to what the original error was that we were , and I'm inclined to agree with @karianna 's comment (unfortunately made after I merged it) that this may not be the right place to fix the issue. It appears to stop things working if we're not building in quite a specific location on the machine.

Willsparker:
Last commit changes the path that the `testJDKWin.sh` script finds the newly built JDK, as it was incorrect as per https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/499/OS=Win2012,label=infra-vagrant-1/

sxa:
In light of the [success of the JDK8 builds](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/506/parameters/), can this now be taken out of draft state?

Willsparker:
Yep! I'd like @johnoliver 's PR to be merged first, as this does rely on that fix.
And I'm yet to get a JDK11 run succeeding the `test` phase, however, I don't think that's necessarily fixable by changing these scripts - it finishes a build and seems to start the test

Willsparker:
@sxa Ready to be merged :-)

PR #1269, opened 2020-04-10
winPB: Updated OpenSSL role

olvap377:
OpenSSL-1.1.1f

**Update:** Now OpenSSL-1.1.1g

Pav.Salimon@ibm.com

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
@@ -4,57 +4,57 @@
 ###########
 - name: Check if OpenSSL 32bit VS2013 installed
   win_stat:
-    path: C:\openjdk\openssl-1.1.1g-x86_32-VS2013
+    path: C:\openjdk\OpenSSL-1.1.1g-x86_32-VS2013
   register: openssl32_installed
   tags: openssl
 
 - name: Check if OpenSSL 64bit VS2013 installed
   win_stat:
-    path: C:\openjdk\openssl-1.1.1g-x86_64-VS2013
+    path: C:\openjdk\OpenSSL-1.1.1g-x86_64-VS2013
   register: openssl64_vs2013_installed
   tags: openssl
 
 - name: Check if OpenSSL 64bit VS2017 installed
   win_stat:
-    path: C:\openjdk\openssl-1.1.1g-x86_64-VS2017
+    path: C:\openjdk\OpenSSL-1.1.1g-x86_64-VS2017
   register: openssl64_vs2017_installed
   tags: openssl
 
-- name: Download openssl-1.1.1g
+- name: Download OpenSSL-1.1.1g
   win_get_url:
     url: https://www.openssl.org/source/openssl-1.1.1g.tar.gz
-    dest: C:\temp\openssl-1.1.1g.tar.gz
+    dest: C:\temp\OpenSSL-1.1.1g.tar.gz
     checksum: ddb04774f1e32f0c49751e21b67216ac87852ceb056b75209af2443400636d46
     checksum_algorithm: sha256
   when: (not openssl32_installed.stat.exists) or (not openssl64_vs2013_installed.stat.exists) or (not openssl64_vs2017_installed.stat.exists)
   tags: openssl
 
-- name: Unpack openssl-1.1.1g for installation
+- name: Unpack OpenSSL-1.1.1g for installation
   win_shell: |
     cd C:\temp
-    C:\7-Zip\7z.exe x C:\temp\openssl-1.1.1g.tar.gz
-    C:\7-Zip\7z.exe x C:\temp\openssl-1.1.1g.tar
+    C:\7-Zip\7z.exe x C:\temp\OpenSSL-1.1.1g.tar.gz
+    C:\7-Zip\7z.exe x C:\temp\OpenSSL-1.1.1g.tar
   when: (not openssl32_installed.stat.exists) or (not openssl64_vs2013_installed.stat.exists) or (not openssl64_vs2017_installed.stat.exists)
   tags: openssl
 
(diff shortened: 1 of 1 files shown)


jdekonin:
For reference this was requested by @pshipton via https://github.com/eclipse/openj9/issues/9089

pshipton:
Just fyi, the OpenJ9 project would like to stick with 1.1.1e for the 0.20.0 release, but the nightly builds can use 1.1.1f.

karianna:
Let's hold off merging this until after this week's release then

jdekonin:
This is just adding the package to the machines and should have no bearing on the builds.  It does not remove previously installed versions.

gdams:
if this is going to be regularly updated It may be worth creating a variable at the top with the version?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -4,57 +4,57 @@
...
   args:
     chdir: 'C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC'
     executable: cmd
   when: (not openssl64_vs2013_installed.stat.exists)
   tags: openssl
 
-- name: Install OpenSSL-1.1.1e 64-bit (VS2017)
-  win_shell: set PATH=C:\Strawberry\perl\bin;C:\openjdk\nasm-2.13.03;%PATH% && .\vcvarsall.bat AMD64 && cd C:\temp\openssl-1.1.1e && perl C:\temp\openssl-1.1.1e\Configure VC-WIN64A --prefix=C:\openjdk\OpenSSL-1.1.1e-x86_64-VS2017 && nmake install > C:\temp\openssl64.log && nmake -f makefile clean
+- name: Install OpenSSL-1.1.1f 64-bit (VS2017)
+  win_shell: set PATH=C:\Strawberry\perl\bin;C:\openjdk\nasm-2.13.03;%PATH% && .\vcvarsall.bat AMD64 && cd C:\temp\OpenSSL-1.1.1f && perl C:\temp\OpenSSL-1.1.1f\Configure VC-WIN64A --prefix=C:\openjdk\OpenSSL-1.1.1f-x86_64-VS2017 && nmake install > C:\temp\openssl64.log && nmake -f makefile clean
```
Willsparker, line 57: Needs to create a new log for OpenSSL-64 bit (VS2017), as it will current just add onto the OpenSSL-64bit (VS2013) log.

gdams: does the log not get cleaned up each time?

Willsparker: Nope, stays in `C:\temp\`

sxa:
Looks like we're now expecting a 1.1.1g [next Tuesday](https://mta.openssl.org/pipermail/openssl-announce/2020-April/000170.html) - since we're not going to implement this until after the release it's probably worth waiting for that to be available and modify this PR to take that version.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
olvap377: @Willsparker Does it add onto or overwrite? We can have different logs for the different VS (2013 & 2017) but if we will be keeping the logs at the end in the event they need to be looked over, then I guess we don't need a cleanup job for them at the end.

Willsparker: Looking at some machines that have recently been setup, it looks like it just overwrites the file, so we've lost the log for 64-bit VS2013. I think it's worth keeping the logs around, if not for anything else, then because we can compare logs if it succeeds building on one system, but fails on another :-)

sxa:
@olavp377 Can you update for 1.1.1g please?

olvap377:
@sxa Updated for 1.1.1g and fixed log file overwrite as discussed with @Willsparker

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/OpenSSL/tasks/main.yml
```diff
@@ -4,57 +4,57 @@
 ###########
 - name: Check if OpenSSL 32bit VS2013 installed
   win_stat:
-    path: C:\openjdk\OpenSSL-1.1.1e-x86_32-VS2013
+    path: C:\openjdk\openssl-1.1.1g-x86_32-VS2013
```
Willsparker, line 7: Is there any particular reason you're changing `OpenSSL` to `openssl`?

Willsparker, line 7: Is there any particular reason you're changing `OpenSSL` to `openssl` ?

olvap377: No particular reason, does it need to be `OpenSSL`?

Willsparker: Yeah, the build scripts will look for them at the specific name: https://github.com/AdoptOpenJDK/openjdk-build/blob/3b9f2ec5883e52b6474d5ac1e9e6772dbe1dcb5b/build-farm/platform-specific-configurations/windows.sh#L78

I'm not entirely sure if the capitalisation is *required* , but if we don't have to change it, I don't think we should :-)

olvap377: Good call will fix up now

PR #1273, opened 2020-04-14
pbTests: Force vagrant halt for Windows runs

Willsparker:
Small changes I've been meaning to get round to.
1) Remove `SLES-12` to run `vagrant rsync-back`, as we no longer have a SLES vagrantfile
2) Remove MSVS_2017 from 'fastmode', as it's required for Window's OpenSSL role now, and OpenJ9 requires MSVS_2017 for some (if not all) builds.
  - Add VS2012_SP1 to fastmode as MSVS_2010 is required to install it, but is ignored when fastmode is specified.
3) Replace `vagrant reload` for `vagrant halt --force && vagrant up` and add `--force` on standalone `vagrant halt`s. This is due to a bug where the Windows Vagrant VM hangs when `vagrant halt` tries to shutdown the VM 'gracefully'. 
4) Added a note explaining why the VM needs to be restarted via Vagrant

### DIFF
--- a/ansible/pbTestScripts/vagrantPlaybookCheck.sh
+++ b/ansible/pbTestScripts/vagrantPlaybookCheck.sh
@@ -44,7 +44,7 @@ processArgs()
 			"--new-vagrant-files" | "-nv" )
 				newVagrantFiles=true;;
 			"--skip-more" | "-sm" )
-				skipFullSetup=",nvidia_cuda_toolkit,MSVS_2010,MSVS_2017";;
+				skipFullSetup=",nvidia_cuda_toolkit,MSVS_2010,VS2010_SP1";;
 			"--build-repo" | "-br" )
 				buildURL="--URL $1"; shift;;
 			"--build-hotspot" )
@@ -194,8 +194,8 @@ startVMPlaybook()
 	rm -f id_rsa.pub id_rsa
 	ssh-keygen -q -f $PWD/id_rsa -t rsa -N ''
 	vagrant up
-	# FreeBSD12 & SLES12 uses a different shared folder type- required to get hosts.tmp from VM
-	if [[ "$OS" == "FreeBSD12" || "$OS" == "SLES12" ]]; then
+	# FreeBSD12 uses a different shared folder type- required to get hosts.tmp from VM
+	if [[ "$OS" == "FreeBSD12" ]]; then
                vagrant rsync-back
 	fi
 	# Generate hosts.unx file for Ansible to use, remove prior hosts.unx if there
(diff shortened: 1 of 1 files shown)


sxa:
What state is the VM in without `halt --force` - what does the console look like. Do we know what circumstances the associated problem occurs in?

Willsparker:
> Hmmm adding another Visual Studio increases the time by about half an hour. Can we get a bit more smart about this - potentially only install the version we need for the specific build? As mentioned we can add/remove openSSL build variants depending on which Visual Studio tags have been skipped too (although I'd prefer to get back to using pre-built OpenSSL as that's a bit of a time sink too now that we're building several versions0

I feel we should try and keep to having `vagrantPlaybookCheck` run as close to what we run on actual build machines as possible - it's a better test of the playbooks, and make us less likely to run into issues when we're setting up actual machines. 

> What state is the VM in without halt --force - what does the console look like. Do we know what circumstances the associated problem occurs in?

According to `vagrant status` the machine is in the `aborted` state. I haven't ran into the issue locally, so it's either due to the machine, or the Jenkins link. The issue appears to occur when the playbook has been ran on a machine, and `vagrant halt` or `vagrant reload` is executed- anything that tries to make the VM shutdown 'gracefully'.

sxa:
> I feel we should try and keep to having vagrantPlaybookCheck run as close to what we run on
> actual build machines as possible - it's a better test of the playbooks, and make us less likely
> to run into issues when we're setting up actual machines.

That is precisely what it does by default and it should be the default, however, the fast mode is intended to let you test small changes that are only likely to affect one role, or to quickly validate a new role across operating systems, in as fast a time as possible without the full cycle which currently takes, I believe the best part of ten hours which means you will frequently have a full day delay on this.

sxa:
> According to vagrant status the machine is in the aborted state. I haven't ran into the issue 
> locally, so it's either due to the machine, or the Jenkins link. The issue appears to occur when
> the playbook has been ran on a machine, and vagrant halt or vagrant reload is executed- 
> anything that tries to make the VM shutdown 'gracefully'.

But have you tried replicating it on `infra-vagrant-1` instead of locally (which I assume means your laptop)?

Willsparker:
>That is precisely what it does by default and it should be the default, however, the fast mode is intended to let you test small changes that are only likely to affect one role, or to quickly validate a new role across operating systems, in as fast a time as possible without the full cycle which currently takes, I believe the best part of ten hours which means you will frequently have a full day delay on this.

Fair enough. I'll remove that stuff from this PR then, and open a new one once I've figured out how to put it in. I'll go off what the build-scripts do on ci.adoptopenjdk.net to determine the various combinations of things to skip.

> But have you tried replicating it on infra-vagrant-1 instead of locally (which I assume means your laptop)?

Running in a screen session on `infra-vagrant-1` now

Willsparker:
@sxa I've been able to recreate the issue running on `infra-vagrant-1` (not through Jenkins). Looking at `top`, it seems to be a virtualbox issue - `/usr/lib/virtualbox/VBoxSVC --auto-shutdown` has a time of nearly 12 hours.

PR #1275, opened 2020-04-14
UnixPB: Correct gcc_7 conditions

Willsparker:
Noticed the discrepancy awhile ago. I'm going to try running the change through the `vagrantPlaybookCheck` job as I'm not sure the condition is necessary required anymore. As can be seen from the below link, the gcc_7 task is run despite it not being intended to, but the build still succeeded: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=CentOS6,label=infra-vagrant-1/464/consoleFull

It's been succeeding in CentOS7 and CentOS8, so it's worth testing them too.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_7/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_7/tasks/main.yml
@@ -21,7 +21,7 @@
     mode: 0644
   when:
     - ansible_distribution == "RedHat" or ansible_distribution == "CentOS" or ansible_distribution == "openSUSE"
-    - ansible_architecture != "CentOS" and ansible_architecture != "x86-64"
+    - ansible_architecture != "CentOS" and ansible_architecture != "x86_64"
     - gcc7_installed.rc != 0
   tags: gcc-7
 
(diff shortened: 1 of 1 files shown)


Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/478/

Willsparker:
The above `VPC` run through failed as `nasm` wasn't installed, due to not being able to detect a version of `gcc7`. I checked the vagrant VMs and they haven't got them installed, so I'll remove this line entirely from the conditions: `- ansible_architecture != "CentOS" and ansible_architecture != "x86_64"`

Willsparker:
Rerunning `VagrantPlaybookCheck` : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/479/

I also took the opportunity to enter a comment explaining why we don't have a checksum for the gcc_7 binary we download.

Willsparker:
All is well with the above `VPC` run

Willsparker:
> The only change seems to be the comment?

I've also removed the ` - ansible_architecture != "CentOS" and ansible_architecture != "x86-64"` from the tasks :-)

sxa:
@karianna Do the two previous comments resolve your rejection of this?

PR #1292, opened 2020-04-24
unixPB: Add openj9 libraries, remove self-built zlib for OpenSUSE

Willsparker:
Adding all libraries that were in `Common/vars/centos.yml`, and removing the self built `zlib` as it was causing issues with finding `libpng` in the build: 
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/507/OS=SUSE12,label=infra-vagrant-1/console

VPC Run: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/515/


Willsparker:
VPC run succeeded 👍

PR #1298, opened 2020-04-27
UnixPB:IPv6

olvap377:
Added IPv6 Role

Enables IPv6 via sysctl.conf on RHEL systems.

Allows for IPv6 testing.

Pav.Salimon@ibm.com

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/IPv6/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/IPv6/tasks/main.yml
@@ -1,0 +1,23 @@
+##############################
+# Enable IPv6 in sysctl.conf #
+##############################
+
+- sysctl:
+    name: net.ipv6.conf.all.disable_ipv6
+    value: '0'
+    sysctl_set: yes
+    state: present
+    reload: no
+  when:
+    - ansible_distribution == "RedHat"
+  tags: IPv6_sysctl
+
+- sysctl:
+    name: net.ipv6.conf.default.disable_ipv6
+    value: '0'
+    sysctl_set: yes
+    state: present
+    reload: yes
+  when:
+    - ansible_distribution == "RedHat"
+  tags: IPv6_sysctl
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/IPv6/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- sysctl:
+    name: net.ipv6.conf.default.disable_ipv6
+    value: '0'
+    sysctl_set: yes
+    state: present
+    reload: yes
+  when:
+    - ansible_distribution == "RedHat"
+  tags: IPv6_sysctl
```
sxa, line 23: Looks like this is missing a newline character at the end

olvap377: On our machines all CentOS have this set by default while all RHEL needed this PB run to allow IPv6 to function. But yes should work for CentOS the same way so we could add that in the `ansible_distribution` as well as I don't think it would do any harm.

I can add the `adoptopenjdk` tag no problem.

Good eye on the newline will fix that up as well.

sxa: Looks like the travis check has failed as you've put a full blank line in at the end (The i.e. it ends with two new line characters where you had zero before - it needs one :-)

sxa:
I'm not expecting any problems with this, but I'll run it through [our test job](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/526/) and once that plus the above issue has been addressed I'll merge it

olvap377:
Looks like travis wants the tasks named. Ill add that as well

Willsparker:
@sxa The Jenkins run didn't run the new playbook role because it skips the `adoptopenjdk` tag - I ran it on a CentOS6 vagrant VM though and it went through okay 👍

Issue #1287, opened 2020-04-22
Re-enable parallel jobs on VagrantPlaybookCheck / infra-vagrant-1


sxa:
Now that Windows VMs in Vagrant are looking more stable after [this change](https://github.com/AdoptOpenJDK/openjdk-build/pull/1689), we should see if we can get the parallelism up since it takes a long time at the moment to test across all platforms. We can use this issue to log progress and what issues we have trying to turn it back on.

We had this [enabled previously](https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/923) but had to turn it off again as things were very unstable with it enabled.

Willsparker:
Would you mind increasing the executors on `infra-vagrant-1` ? I suppose we could see if having `2` shows the issues we had before? :-)

sxa:
I have granted @Willsparker configure access to [infra-vagrant-1](https://ci.adoptopenjdk.net/computer/infra-vagrant-1/) in jenkins (he's the only person who uses it and has full access to the underlying machine already) so is able to make this change now.

Willsparker:
Executors have been put up to 3, with a JDK8 job (fastmode, run build, but no Windows, no test)
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/523/
Just as a test to ensure the error is still there, mainly.

Willsparker:
A rebuild of the above run showed the issue is still there: 
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/526/
Each of the failing OSs have failed with error messages that would point to the machine being turned off half way through execution ( i.e., `Failed to connect to the host via ssh: Shared connection to ___ closed `  or  `A prompt was encountered while running a command, but no input data was specified`)

Running `vagrant global-status --prune` shows the machines that failed, were left on the machine, so other runs of VPC aren't accidentally destroying them. Their state varies between `poweroff` and `aborted`.

It seems that the end of `ansible` commands are correlated to the failure of a different OS, i.e.
CentOS6 failed at around 22:58:00
CentOS8 finished it's build at around 22:56:36 (the build started with `ansible all .... `

Rerunning without the `build` option as the jenkins job will hang for awhile if half way through a build and the VM 'turns off'. It should give us a better idea of if this correlation is correct. If so, it may be helped by #1303

Willsparker:
Rerunning without using `build` or `test` has given a better show of the correlation: 

[...]

PR #1313, opened 2020-05-07
pbTests: Add BUILD_ID variable for Jenkins

Willsparker:
Fixes: #1287 

See above issue for more in depth details.
This variable should effectively stop Jenkins from killing the process that keeps the vagrant VMs running - and allowing the `VagrantPlaybookCheck` job to run in parallel.
https://wiki.jenkins.io/display/JENKINS/ProcessTreeKiller


Willsparker:
FYI, in case anyone did want it, I've put a run through `vagrantPlaybookCheck` . 
The `vagrantPlaybookCheck.sh` script can't actually be tested by putting it as the `REPOSITORY_URL` parameter in the Jenkins job, the job needs to be reconfigured to pull from this branch, itself. (hence why on the `status` page of the run, it says 'origin/parallel' )

https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/569/

PR #1314, opened 2020-05-07
pbTests: Debian10 rsync folder changes

Willsparker:
Turns out the Debian 10 box I'm using, uses rsync folders and not the standard `vboxsf` that most boxes have. It's not a massive issue as we already support `FreeBSD12` that uses `rsync` folders, so it's just the addition of the condition.

### DIFF
--- a/ansible/Vagrantfile.Debian10
+++ b/ansible/Vagrantfile.Debian10
@@ -7,7 +7,7 @@ sudo apt-get install software-properties-common -y
 sudo apt-get install gpg -y	
 sudo apt-get update
 # Get IPs of the VM, and output to shared folder
-ip address show | grep -e "\\inet\\b" | cut -d' ' -f6 | cut -d/ -f1  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
+ip address show | grep -e "\\binet\\b" | cut -d' ' -f6 | cut -d/ -f1  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
 # Put the host machine's IP into the authorised_keys file on the VM
 if [ -r /vagrant/id_rsa.pub ]; then
         mkdir -p $HOME/.ssh && cat /vagrant/id_rsa.pub >> $HOME/.ssh/authorized_keys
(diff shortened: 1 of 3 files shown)


Willsparker:
Ref: latest commit
Forgot to add this condition to #1309

Willsparker:
Ref: latest commit
Forgot to add this condition to #1309

ansible/Vagrantfile.Debian10
```diff
@@ -7,7 +7,7 @@ sudo apt-get install software-properties-common -y
 sudo apt-get install gpg -y	
 sudo apt-get update
 # Get IPs of the VM, and output to shared folder
-ip address show | grep -e "\\binet\\b" | cut -d' ' -f6 | cut -d/ -f1  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
+ip address show | grep -e "\\inet\\b" | cut -d' ' -f6 | cut -d/ -f1  >> /vagrant/playbooks/AdoptOpenJDK_Unix_Playbook/hosts.tmp
```
sxa, line 10: DId you mean to remove this `b`?

Willsparker: Weirdly the give the same output - I'll change it back to keep it consistent with the Deb8 VF

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
```diff
@@ -116,6 +116,7 @@
   with_items: "{{ gcc_compiler }}"
   when:
     - (ansible_distribution_major_version != "9" and ansible_architecture != "armv7l")
+    - ansible_distribution_major_version != "10"
```
sxa, line 119: This doesn't seem related to the PR title.
Also shouldn't we be saying which distribution this is? When we hit RHEL/Cent10 this conditional will trigger

Willsparker: Nope, this is within the `Common/tasks/Debian.yml` file ; which only runs on Debian guests. However, I am going to remove this because I've done it a better way in my own testing, and it relates to a PR I'll be doing later on today.

PR #1317, opened 2020-05-11
fix playbooks on macOS

gdams:
a few extra skips were required to make the playbook fully run on macOS

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -50,18 +50,24 @@
       tags: [build_tools, build_tools_openj9]
     - role: adoptopenjdk_install
       jdk_version: 8
+      when: ansible_distribution != "MacOSX"
       tags: build_tools
     - role: adoptopenjdk_install  # JDK11 Build Bootstrap
       jdk_version: 10
+      when: ansible_distribution != "MacOSX"
       tags: build_tools
     - role: adoptopenjdk_install
       jdk_version: 11
+      when: ansible_distribution != "MacOSX"
       tags: build_tools
     - role: adoptopenjdk_install  # JDK13 Build Bootstrap
       jdk_version: 12
+      when: ansible_distribution != "MacOSX"
       tags: build_tools
     - role: adoptopenjdk_install  # JDK14 Build Bootstrap
       jdk_version: 13
+      when: ansible_distribution != "MacOSX"
+      tags: build_tools
     - OpenSSL102                  # OpenJ9
     # - Nagios_Plugins              # AdoptOpenJDK Infrastructure
     # - Nagios_Master_Config        # AdoptOpenJDK Infrastructure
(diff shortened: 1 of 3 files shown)


gdams:
@sxa added a prefix,

Using the API would get difficult on macOS, especially as the bundle needs to be notarized, sticking with the casks is by far the easiest approach here.

sxa:
> Using the API would get difficult on macOS, especially as the bundle needs to be notarized

Can you clarify what is it that doesn't work for a boot JDK here if it's not notarized? And what else would be difficult?

gdams:
gatekeeper would not allow the binary to be executed which would cause issues.

sxa:
> gatekeeper would not allow the binary to be executed which would cause issues.

@gdams Can you clarify/document from a naive non-mac user perspective what these issues would be? My understanding is that it's possible to extract into `/usr/local` as an adminstrator and have java execute as a normal user from there.

gdams:
@sxa for starters the macos bundles have to be extracted to:

`/Library/Java/JavaVirtualMachines/adoptopenjdk-<version>.jdk/` in order for the system to correctly identify them.

On a Catalina machine or above any bundle that isn't notarized will cause the following error:

image

essentially as soon as the java process is created it gets destroyed by gatekeeper.

Using homebrew is definitely the way to go here, it also allows us to update the openjdk versions easily.

aahlenst:
As a side note only: If you download via curl or wget, the Gatekeeper attributes aren't added. You can move the extracted tarball to `/Library/Java/JavaVirtualMachines` and launch java. Gatekeeper etc. only kicks in if you use Apple tools like Safari that add the quarantine attributes. Or Homebrew, which does that, too 🙄 

Ubuntu 20.04 ISO downloaded via Safari:

```
$ xattr ubuntu-20.04-live-server-amd64.iso
com.apple.lastuseddate#PS
com.apple.macl
com.apple.metadata:kMDItemDownloadedDate
com.apple.metadata:kMDItemWhereFroms
com.apple.quarantine
```

AdoptOpenJDK 14 downloaded via wget:

```
$ xattr OpenJDK14U-jdk_x64_mac_hotspot_14.0.1_7.tar.gz
```

sxa:
Consistent with what @aahlenst is suggesting I have confirmed with someone on Catalina that they are able to extract the tarball into `/usr/local` on their machine and have it run as a normal user, so I'm not sure what the inmplications of the extraction location you mention or the Gatekeeper error is with respect to extracting the tarball, but that does not appear to be correct.

gdams:
okay @sxa if it works then I have not personal objections. As mentioned before I would suggest extracting them to the correct location `/Library/Java/JavaVirtualMachines`

There will be some script changes needed to support this as the macOS tarballs are in the form of a bundle as well. Let's get this PR in and then we can work out the best approach

sxa:
As discussed on our call last week if the changes go in to make the extraction consistent with the other platforms (I'm ok with the location you suggest!) then I'll approve and merge this.

karianna:
@gdams also merge conflict now

PR #1318, opened 2020-05-11
UnixPB: Installer role and playbook

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1283

Includes a role, `installer`, which installs the prerequisite build tools for the linux installer job, and a playbook, `installer-setup.yml`, which runs this role alongside the `Jenkins_User` and `Superuser` role to configure machines to be able to execute the Linux installer Jenkins job.

Tested by running Rpm and Deb build tasks, see https://github.com/AdoptOpenJDK/openjdk-installer/tree/master/linux#deb-packages, on a ubuntu 18.04 machine on which the playbook had been run. Build tasks were successful

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/installer-setup.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/installer-setup.yml
@@ -1,0 +1,15 @@
+- hosts: all
+  gather_facts: yes
+  tasks:
+    - block:
+      # Set standard variables
+        - name: Load AdoptOpenJDKs variable file
+          include_vars: group_vars/all/adoptopenjdk_variables.yml
+
+  #########
+  # Roles #
+  #########
+  roles:
+    - installer
+    - Jenkins_User
+    - Superuser
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/linux_installer_setup.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/linux_installer_setup.yml
@@ -10,6 +10,6 @@
   # Roles #
   #########
   roles:
-    - installer
+    - linux_installer
     - Jenkins_User
     - Superuser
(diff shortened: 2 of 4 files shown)


Haroon-Khel:
Currently only supports Ubuntu 18.04 on x86

gdams:
Is there a reason that you don't just want to install these dependencies on all our build machines?

karianna:
> Is there a reason that you don't just want to install these dependencies on all our build machines?
> 
> creating an additional playbook for this seems like overkill IMO

I think it's about using Ansible to be able to replicate installing those dependencies if we lose machines?

sxa:
@gdams it was more around any signing keys/authentication tokens etc. - while that can be done using jenkins secrets, we're not currently doing that and from a security/paranoia perspective I'm more keen to do this on machines we can keep a bit more locked down.

@karianna I believe George was suggesting adding them into the main build/test playbooks as opposed to objecting to adding this stuff into the playbooks at all.

Haroon-Khel:
@gdams I included the `linux_installer_setup.yml` playbook in the case where we need a sole machine for linux installer role, but if its considered overkill I can remove it.
I still feel that it would be convenient to have the **role** which installs only the linux installer prerequisites, as per https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1283#issuecomment-614663778, since this role can be included in the main Unix playbook if needed.

Willsparker:
@Haroon-Khel I'd agree with having the individual role in the main playbook (personally)- as that would keep consistency with the Windows playbook ( https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/WiX/tasks/main.yml ).

sxa:
@Haroon-Khel Have you been able to verify whether a setup done with this playbook is suitable for running all the operations in the install jobs or does that still have to be verified?

Haroon-Khel:
Ive tested only the gradle tasks `buildDebPackage` and `buildRpmPackage`. I wasnt able to test the upload tasks as I do not have access to the artifactory user and password

PR #1321, opened 2020-05-12
UnixPB: Alter Deb Common file & pb script to build JDK8/HS on Deb10

Willsparker:
Several changes to allow building JDK8 / HS on Debian10:
1) Change `buildJDK.sh` to detect `JAVA_HOME`. correctly for Debian Buster
2) Add a condition to ensure it doesn't attempt to install gcc-4.8 (as this is unavailable for Debian 10's repos)
3) Symlink `/usr/bin/gcc-7` and `/usr/bin/gcc` and the same for `g++`. As `gcc-4.8` is unavailable to Debian 10, the gcc to use to build JDK8/HS should be the lowest version available - `gcc-7.4`. Tested and it does build JDK8/HS now.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
@@ -116,6 +116,7 @@
   with_items: "{{ gcc_compiler }}"
   when:
     - (ansible_distribution_major_version != "9" and ansible_architecture != "armv7l")
+    - not (ansible_distribution_major_version == "10" and ansible_architecture == "x86_64")
   tags: build_tools
 
 - name: Install GCC G++ on Raspian Buster
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -150,8 +150,7 @@ if [ -z "$JAVA_HOME" ]; then
 	export JAVA_HOME=$(ls -1d /usr/lib/jvm/adoptopenjdk-8-* | head -1)
 fi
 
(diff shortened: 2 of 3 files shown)


Willsparker:
`VPC` run : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/586/

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
```diff
@@ -198,3 +199,19 @@
     name: zh_TW.UTF-8
     state: present
   tags: locales
+
+#####################
+# GCC for Debian 10 #
+#####################
+
+- name: Set gcc/g++ 7 to /usr/bin/gcc and /usr/bin/g++ on Debian10
```
sxa, line 207: This makes me nervous as it could overwrite any existing `/usr/bin/gcc` that the user might have due to having another version installed. Can we avoid this and set `CC`/`CXX` variables when required?

If we can't do it another way it needs the `dont_remove_system` tag to allow us to guard against this, or maybe also `adoptopenjdk` so that end users don't run into trouble with it

sxa:
Can you squash the commits and simplify the commit message please? Something like `unixPB: Complete support for building on Debian10`

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
Willsparker: I'll investigate and update the PR As required :)

Willsparker:
@sxa Rebased and reworded. Setting `CC` and `CXX` worked on my local Debian10 VM, so I removed the symlink task 👍 Good call

PR #1322, opened 2020-05-12
WinPB: Remove Windows-SDK-10.0 installation

Willsparker:
I've removed this as this should be installed with MSVS_2017, so it would always skip, unless `VagrantPlaybookCheck` is being ran to build JDK8/J9 with 'fastmode' on.
e.g.:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/571/OS=Win2012,label=infra-vagrant-1/console     (Failed at the WiX role)
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/572/OS=Win2012,label=infra-vagrant-1/console    (Got past the WiX role)

They ran with the same parameters, however the second one didn't run with `fastmode` on.
In current normal use, this wouldn't have been noticed as the Windows machines that run the installer jobs are also build machines ( I assume ), that will have MSVS_2017 installed anyway, so unless we have/get Windows machines specifically for making the installer, this is unnecessary.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/WiX/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/WiX/tasks/main.yml
@@ -32,16 +32,3 @@
     state: absent
   ignore_errors: yes
   tags: Wix
-
-- name: Test if Windows 10 SDK is already installed
-  win_stat:
-    path: 'C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x64\signtool.exe'
-  register: win10_sdk_installed
-  tags: Wix
-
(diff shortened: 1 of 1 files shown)


Willsparker:
Closing due to inability to test this on any Windows system - @gdams suggested that it's okay for this Vagrant VM, however may not be for different providers. For the sake of `VagrantPlaybookCheck`, I'll just add WiX to the skip tags as it isn't required to build or test a JDK.

sxa:
@gdams @aahlenst  I've got a feeling we may have discussed this before but before closing this permanently can you clarify under what circumstances you believe this may not work with the version included in VS2017? in general I'd rather not install stuff unnecessarily if possible.

Willsparker:
( apologies to keep opening and closing this- see the above PR as I feel this is the best fix, if we decide to keep the installation of the task )

PR #1323, opened 2020-05-12
pbTests: All Wix to skip tags for windows run

Willsparker:
ref: #1322 

As I'm unable to test if what I described in the above PR, is the case for other versions of Windows / MSVS_2017, I've opted to just skip `Wix` instead as it isn't required for a build / test of a JDK, and so is inconsequential to `VagrantPlaybookCheck`.


Willsparker:
(See above PR, as this is a better solution to the problem that this PR was attempting to solve)

PR #1325, opened 2020-05-12
UnixPB: Add Zulu7 for JDK8 Build bootstrap, make buildJDK.sh use JDK7

Willsparker:
ref: https://github.com/AdoptOpenJDK/openjdk-build/issues/1732

Add in Zulu-7 to Unix playbook, as it can be a consistent JDK-7 amongst all the distributions, as some new linux distros are unable to install `openjdk-7` with their package manager.

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -82,21 +82,26 @@ checkJDKVersion() {
 }
 
 setBootJDK() {
-	local buildJDKNumber=$(echo ${JAVA_TO_BUILD//[!0-9]/})
-	local bootJDKNumber=$(($buildJDKNumber - 1));
-        # if building JDK8/9 look for 'jdk8u', not 'jdk-x'
-	if [[ $buildJDKNumber -eq 8 || $buildJDKNumber -eq 9 ]]; then
-		export JDK_BOOT_DIR=$(find /usr/lib/jvm -maxdepth 1 -name *jdk8*)	
-		return
-	else
-		export JDK_BOOT_DIR=$(find /usr/lib/jvm -maxdepth 1 -name *jdk-$bootJDKNumber*)
-	fi
-
-	if [ -z "${JDK_BOOT_DIR}" ]
-	then
-		echo "Can't find jdk$bootJDKNumber to build JDK, looking for jdk$buildJDKNumber"
-		export JDK_BOOT_DIR=$(find /usr/lib/jvm -maxdepth 1 -name *jdk-$buildJDKNumber*)
-	fi
+        local buildJDKNumber=$(echo ${JAVA_TO_BUILD//[!0-9]/})
+        local bootJDKNumber=$(($buildJDKNumber - 1));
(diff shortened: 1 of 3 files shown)


Willsparker:
FYI, zulu-7 is installed using the shell module, using `wget` and `tar` as both the `unarchive` and `get_url` ansible modules would throw the following errors, even with `validate_certs: false`.

`
"msg": "Failed to validate the SSL certificate for cdn.azul.com:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine  (the python executable used (/usr/bin/python) is version: 2.6.6 (r266:84292, Jun 20 2019, 14:14:55) [GCC 4.4.7 20120313 (Red Hat 4.4.7-23)]) or you can install the `urllib3`, `pyOpenSSL`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible. The exception msg was: [Errno 1] _ssl.c:492: error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol."
`
However, running `wget` manually on the machine did work, hence why `shell` works.

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/573/
VagrantPlaybookCheck run ^

Willsparker:
With the latest commit, the PR has include fairly significant changes to `buildJDK.sh`: 
1) remove code to find the distribution specific versions of JDK8 to set `JAVA_HOME` with as it's unnecessary as we install jdk8 with adoptopenjdk_install
2) Remove setting `JDK7_BOOT_DIR`, as  setting `JDK_BOOT_DIR` takes priority in the build, when building JDK8
3) Set `JDK_BOOT_DIR` to be JDK7 when building JDK8 (it was previously setting JDK8).

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/581/OS=SUSE12,label=infra-vagrant-1/console
Unfortunately, SUSE/SLES12 IS new enough to not have openjdk-7 in it's package repos, but does have a version of python low enough to cause issues with the `unarchive` repo. For now, I've removed it from the zulu-7 install, as we don't have any SLES build machines (and only one test machine), it shouldn't matter if it uses jdk-8 to build JDK-8 for the playbook check. 
That being said, @sxa If you want, I can put back in that `shell` module task and make it run only for SLES/OpenSUSE.

Willsparker:
Between https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=CentOS8,label=infra-vagrant-1/584/
and
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=Ubuntu1804,label=infra-vagrant-1/582/
It appears to work for the other distributions.

PR #1338, opened 2020-05-26
inv: Update Test/infra machine sections

Willsparker:
Fixes: #634 

This should get the `inventory.yml` more accurately reflecting Jenkins. This PR is primarily ensuring `inventory.yml` is in alphabetical order and adds in a couple machines that are on Jenkins but not here.
I have omitted from `inventory.yml` the machines with the `-sxa` prefix, on request of @sxa, as they're only meant to be temporary machines.
Currently `test-inspira-solaris10u11-sparcv9-1` is commented out as I'm unable to find the IP address for the machine (ping @gdams )

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -18,7 +18,7 @@ hosts:
 
       - softlayer:
           ubuntu1604-x64-3: {ip: 158.176.72.83, description: jckservices.adoptopenjdk.net}
-          vagrant-x64-1: {ip: 52.117.150.132, description: Bare metal machien to run VM playbook testing}
+          ubuntu1804-x64-1: {ip: 52.117.150.132, description: Bare metal machine to run VM playbook testing}
 
       - godaddy:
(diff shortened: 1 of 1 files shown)


ansible/inventory.yml
```diff
@@ -183,8 +187,7 @@ hosts:
           ubuntu1604-x64-1: {ip: 51.15.76.107}
 
       - softlayer:
-          ubuntu1604-x64-1: {ip: 169.55.150.155}
           rhel74-x64-1: {ip: 169.55.170.70}
           rhel69-x64-1: {ip: 169.55.150.147}
-          win2012r2-x64-1: {ip: 169.55.170.72, user: admin}
+          ubuntu1604-x64-1: {ip: 169.55.150.155}
```
sxa, line 192: Windows machine `169.55.170.72` is still live in softlayer and allocated to us so I don't think these should be removed

Willsparker: I'll reinstate it :)

PR #1340, opened 2020-05-27
UnixPB: Failing NTP_TIME role for Ubuntu 20

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1226

Skips the `Set timedatectl set-ntp no` task for Ubuntu 20 as it causes the playbook to break, but instead runs an alternative, but equivalent, task of disabling the `systemd-timesyncd` service. This alternative task can also be run on distros that are not Ubuntu 20, since, after some inspection, our machines have the `systemd-timesyncd` service disabled by default.

However this service is not available on Centos or Redhat, hence the alternative task is skipped for those distros.


sxa:
Need to validate this on Debian ... Although at the moment the playbook check job is failing badly - ping @Willsparker to test this once the server has been recovered

Willsparker:
Unfortunately I'm unable to put this through `VPC` for Ubuntu1804 is `apt-get upgrade` is hanging. I tried this locally and the role ran through okay. And here's the `VPC` run for Ubuntu1604 - https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=Ubuntu1604,label=infra-softlayer-ubuntu1804-x64-1/619/
I've started a run for all of the CentOS Versions to make sure they skip the new task okay : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/629/
.. however the conditions to skip look correct to me.

PR #1342, opened 2020-05-28
inv: Put back test-softlayer-win2012r2-x64-1

Willsparker:
Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1338#discussion_r431779754


sxa:
I guess this was removed as it wasn't in jenkins. We should probably see if we can get into it and run some tests on it to increase our testing capacity (maybe after searching issues to see if there was a reason it was shut down and/or repurposed)

Issue #1344, opened 2020-05-29
Ubuntu1804 Vagrantfile apt-upgrade hanging


Willsparker:
Noticed in several VagrantPlaybookCheck runs that Ubuntu1804 will hang when running the `apt-get upgrade` task in the `Common` role.
I created a local Vagrant VM and ran the command manually, in which a prompt for `grub-pc` came up:
<img width="1423" alt="Screenshot 2020-05-29 at 10 21 10" src="https://user-images.githubusercontent.com/52783864/83243671-2835c400-a196-11ea-8b70-127a3544823a.png">
(apologies for the screenshot)

I assume this is what is causing the hang when using Ansible.
Selecting the first option and continuing doesn't seem to cause anymore issues.
Here's the log :
```
vagrant@adoptopenjdkU18:~$ sudo apt-get upgrade
...
snapd.snap-repair.service is a disabled or a static unit, not starting it.
Setting up libdns-export1100 (1:9.11.3+dfsg-1ubuntu1.12) ...
Setting up apport (2.20.9-0ubuntu7.15) ...
Installing new version of config file /etc/init.d/apport ...
apport-autoreport.service is a disabled or a static unit, not starting it.
Setting up ubuntu-minimal (1.417.4) ...
Setting up libdns1100:amd64 (1:9.11.3+dfsg-1ubuntu1.12) ...
Setting up git (1:2.17.1-1ubuntu0.7) ...
Setting up libpython3.6-stdlib:amd64 (3.6.9-1~18.04ubuntu1) ...
Setting up ruby2.5 (2.5.1-1ubuntu1.6) ...
Setting up curl (7.58.0-2ubuntu3.8) ...
Setting up grub-pc (2.02-2ubuntu8.15) ...

(prompt shows)

Installing for i386-pc platform.
Installation finished. No error reported.
Sourcing file `/etc/default/grub'
Generating grub configuration file ...
Found linux image: /boot/vmlinuz-4.15.0-54-generic
Found initrd image: /boot/initrd.img-4.15.0-54-generic
Found linux image: /boot/vmlinuz-4.15.0-45-generic
Found initrd image: /boot/initrd.img-4.15.0-45-generic
done
Setting up open-vm-tools (2:11.0.5-4ubuntu0.18.04.1) ...
Installing new version of config file /etc/vmware-tools/vm-support ...
Setting up python3.6-minimal (3.6.9-1~18.04ubuntu1) ...
...

```

[...]

PR #1345, opened 2020-05-29
VF: Force no-prompt upgrade to grub-pc on U18

Willsparker:
Fixes: #1344 

Fixes an issue specific to the Vagrant box that we use, where `grub-pc` needs to be upgraded with the `DEBIAN_FRONTEND=noninteractive` variable, to stop a prompt coming up on the `apt-get upgrade` task in the playbook. 

In draft until I can run it through `VPC` properly.

### DIFF
--- a/ansible/Vagrantfile.Ubuntu1804
+++ b/ansible/Vagrantfile.Ubuntu1804
@@ -6,6 +6,7 @@ sudo apt-get install tree -y
 sudo apt-get install software-properties-common
 sudo apt-add-repository ppa:ansible/ansible
 sudo apt-get update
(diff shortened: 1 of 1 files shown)


Willsparker:
https://askubuntu.com/a/147079
The first reply to this answer mentioned that if `sudo` is put after, it won't set the variable properly. It worked when ir an it on my mac :-)

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=Ubuntu1804,label=infra-softlayer-ubuntu1804-x64-1/637/console
Works! 👍

PR #1366, opened 2020-06-04
UnixPB: Don't validate certs for wget-ting the git source

Willsparker:
ref: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=SUSE12,label=vagrant/658/console

SUSE has been consistently failing due to the SSL cert not being able to be validated. Unfortunately, attempting to install the python modules suggested failed due to not being able to find them, therefore I opted for not `validating_certs` - as we have the checksum in place, (AFAIK) there's no added security risk of this.


Willsparker:
Closing as a new PR will be made when a solution is found.

PR #1367, opened 2020-06-04
cleanup leftover tarballs and dockerfiles

gdams:
fixes: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1312

### DIFF
--- a/ansible/Dockerfile.CentOS6
+++ b/ansible/Dockerfile.CentOS6
@@ -12,7 +12,9 @@ RUN echo "localhost ansible_connection=local" > /ansible/hosts
 
 RUN set -eux; \
  cd /ansible; \
- ansible-playbook -i hosts -s ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --skip-tags="debug,hosts_file,hostname,adoptopenjdk,jenkins,nagios,superuser,docker,swap_file,crontab,adoptopenjdk_install"
+ ansible-playbook -i hosts -s ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --skip-tags="debug,hosts_file,hostname,adoptopenjdk,jenkins,nagios,superuser,docker,swap_file,crontab,nvidia_cuda_toolkit,adoptopenjdk_install"
+
+RUN rm -rf /ansible; yum remove ansible; yum clean all
 
 RUN groupadd -g 1000 ${user}
 RUN useradd -c "Jenkins user" -d /home/${user} -u 1000 -g 1000 -m ${user}
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
@@ -2,6 +2,11 @@
 ##########
 # ccache #
 ##########
+
+- name: Set ccache version
+  set_fact: ccacheVersion=3.4.2
+  tags: ccache
+
 - name: Check if ccache is already installed
   stat:
     path: /usr/local/bin/ccache
@@ -9,7 +14,7 @@
   tags: ccache
 
 - name: Download ccache.tar.gz
-  command: wget -O /tmp/ccache.tar.gz https://github.com/ccache/ccache/releases/download/v3.4.2/ccache-3.4.2.tar.gz
+  command: wget -O /tmp/ccache.tar.gz https://github.com/ccache/ccache/releases/download/v{{ ccacheVersion }}/ccache-{{ ccacheVersion }}.tar.gz
     warn=False
   when: ccache_status.stat.isdir is not defined
   tags: ccache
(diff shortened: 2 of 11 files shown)


gdams:
@sxa some cleanups are done inline and others are done in the cleanup task.

I much prefer the first because its part of the role and ensures that people don't skip the cleanup tag by accident. It also means that when we bump a version, it's much easier to remember to bump it in the same role rather than in some cleanup role later on.

This was the case with cmake which was bumped from `3.7.2` to `3.11.4` but the cleanup role was never updated.

I'd be happy to update this PR by removing the cleanup role completely and making the cleanup part of each specific role?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
```diff
@@ -42,6 +47,7 @@
     state: absent
   with_items:
     - /tmp/ccache.tar.gz
-    - /tmp/ccache-3.4.2
+    - /tmp/ccache-{{ ccacheVersion }}
+    - /home/{{ Jenkins_Username }}/ccache.tar
```
sxa, line 51: Where is this being creasted? Not sure /home/jenkins/ccache.tar should ever exist

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/freemarker/tasks/main.yml
```diff
@@ -28,3 +28,13 @@
   command: mv /tmp/freemarker-2.3.8/lib/freemarker.jar /Users/{{ Jenkins_Username }}
   when: not freemarker.stat.exists and ansible_distribution == "MacOSX"
   tags: [freemarker, adoptopenjdk]
+
+- name: Remove downloaded packages for freemarker.jar
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /tmp/freemarker-2.3.8
```
sxa, line 37: Maybe abstract out 2.3.8 as you have for other roles here?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml
```diff
@@ -10,7 +10,11 @@
 # Addresses https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/499
 # For RHEL 8, use gmake 4.2.1 already installed.
 
-- name: Test if self-built make 4.1 is available
+- name: Set make version
+  set_fact: makeVersion=4.1
+  tags: goodmake_source
+
+- name: Test if self-built make is available
```
sxa, line 17: Would be nice to keep the version number in the ansible execution log so add {{ makeVersion }} into this

ansible/Dockerfile.CentOS6
```diff
@@ -12,7 +12,9 @@ RUN echo "localhost ansible_connection=local" > /ansible/hosts
 
 RUN set -eux; \
  cd /ansible; \
- ansible-playbook -i hosts -s ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --skip-tags="debug,hosts_file,hostname,adoptopenjdk,jenkins,nagios,superuser,docker,swap_file,crontab,adoptopenjdk_install"
+ ansible-playbook -i hosts -s ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --skip-tags="debug,hosts_file,hostname,adoptopenjdk,jenkins,nagios,superuser,docker,swap_file,crontab,nvidia_cuda_toolkit,adoptopenjdk_install"
```
sxa, line 15: Was the removal of nvidia_cuda_toolkit from this because you do'nt expect the dockerfiles to be run on a machine that supports it from a testing perspective?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
gdams: I'm not sure either? I left it though to be safe

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml
gdams: AFAIK you can't use variables in ansible task names: https://stackoverflow.com/a/29001553/5838529

ansible/Dockerfile.CentOS6
gdams: The main reason right now is that it bloats the docker file. The cuda toolkit installation is over 10gb. If we ever have a use case where we need the cuda kit here we can create a separate layer.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/freemarker/tasks/main.yml
gdams: done

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml
gdams: actually I take that back, managed to make it work.

ansible/Dockerfile.CentOS6
sxa: Side note: We should probably add a comment to the top of the Dockerfile indicating that they are for codebuild and other such places, but it would depend on whether the host hardware supports it as to whether it can ebe used. Saying that, I'm ok with not having it in there for now - it certainly does bloat it!

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/ccache/tasks/main.yml
sxa: I say delete it - especially now that the download, building and removal is all in one place it looks completely obsolete. I suspect at one point in the past we tried to extract from the tar.gz into a .tar and so it's historic as we don't do that any more. If you can do that I'll happily approve :-)

gdams: removed

sxa:
(Full testing of this is effectively blocked at present due to https://github.com/AdoptOpenJDK/openjdk-build/issues/1863 not allowing the builds to go through but so far the playbooks have worked ok)

Issue #1001, opened 2019-11-18
Ensure JDK13 is on  all build machines for bootstrapping JDK14


sxa:
Updates made in the playbooks but JDK13 is not yet deployed on many of the machines (which is preventing many of the builds in https://ci.adoptopenjdk.net/view/Tooling/job/JavaFutures/ from working (fixing that will just require execution on the two machines that are being used by that job but it needs to be done everywhere.

sxa555:
Added to both machines used for building the [JavaFutures](https://ci.adoptopenjdk.net/view/Tooling/job/JavaFutures/) codebases for now, which should resolve the immediate problem.

sxa:
Now that we are building JDK head which is JDK15 we should also install a JDK14 boot JDK on each build machine

Willsparker:
From the machines I have access to:

| Build Machine                     | JDK13 | JDK14 |
|-----------------------------------|-------|-------|
| build-azure-win2012r2-x64-1       | Y     | N     |
| build-azure-win2012r2-x64-2       | Y     | N     |
| build-digitalocean-centos69-x64-1 | Y     | N     |
| build-digitalocean-centos69-x64-2 | N     | N     |
| build-osuosl-centos74-ppc64le-1   | N     | N     |
| build-osuosl-centos74-ppc64le-2   | Y     | N     |
| build-packet-centos74-armv8-1     | N     | N     |
| build-softlayer-centos6-x64-1     | Y     | N     |
| build-softlayer-win2012r2-x64-1   | N     | N     |
| build-softlayer-win2012r2-x64-2 *  | N     | N     |

* also doesn't have JDK10 on it.
I'll get to work putting them on these machines 👍

sxa:
> @sxa What machines are left, and can I get access to them? :-)

You tell me - you've been updating the inventories recently!

Also bear in mind that the appropriate `JDK*_BOOT_DIR` definitions in jenkins for each machine will need updated to point to the install location

Willsparker:
I've updated the `JDK*_BOOT_DIR` variables for each machine I've done :-) 
Here's a list of machines at `ci.adoptopenjdk.net` that I haven't been able to access to check if they have JDK13/14 on them.
```
build-godaddy-rhel7-x64-1 
build-inspira-solaris10u11-sparcv9-1 (D)(Not sure how to ping)
build-inspira-solaris10u11-sparcv9-2		
build-linaro-centos74-armv8-1 (D)		
[...]

PR #1375, opened 2020-06-08
PBs: Update which JDKs to install

Willsparker:
Ref: #1001 

Stop install JDK12 on machines as we no longer build JDK13, start installing JDK14 for JDKNext.


Willsparker:
@Haroon-Khel https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/686/console :)

Issue #1106, opened 2020-01-21
Ensure make 4.1 is the default on Linux/ppc64le build machines


sxa:
I've just replaced `/usr/bin/make` with a symlink to `/usr/local/bin/make` after hitting the issue described in #499 again. Ideally we'd modify the playbooks to remove the system installed make (as we do for the `GIT_Source` role unless the `dont_remove_system` tag is skipped on the playbooks (Hmmm that should possibly have been called `remove_system`

Since it affected the builds today (Maybe as a result of whatever's causing https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1097) I've adjusted it again so hopefully that will have resolved the immediate problem.

Willsparker:
I'll look into this whilst I wait for `qemu` stuff :-)

Willsparker:
I've had a look - putting in the conditional to run the `gmake` stuff when the `make` version is too low wasn't difficult, however when adding a task to remove system `make`, an error message came up : 
```
TASK [gmake : Remove system git if needed (yum)] **************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error: Trying to remove \"yum\", which is protected\n
```
It seems `yum` is dependent on `make`, therefore on CentOS/RHEL (at least) we can't remove it. I can add a task to symlink it from `/usr/bin/make` to `/usr/local/bin/make`, as this will make the system `make` inaccessible?

sxa555:
This leaves us with three options:
1. Force remove `make`, breaking the `yum` dependency
2. Find/produce a make 4.1 RPM package and use that instead
3. As you say, replace `/usr/bin/make` with a symlink, but that will be overridden if yum decides to update `make` ... TO avoid that we can add an exclude for `make` into `/etc/yum.conf`
4. Modify our scripts to ensure that `/usr/local/bin/make` is used if it exists instead of just `make`

4 may be the best option, although of the others 3 is likely best

sxa555:
In fact option 4 might be quite trivial ... https://github.com/AdoptOpenJDK/openjdk-build/blob/master/configureBuild.sh#L262

Willsparker:
Just for updating-sake, @sxa555 and I have come to the conclusion that there is something in the build scripts that are causing the builds to use the wrong version of `make` as https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/602 and https://github.com/AdoptOpenJDK/openjdk-build/pull/759
[...]

PR #1393, opened 2020-06-16
UnixPB: Exclude make from updating when gmake is built

Willsparker:
Fixes: #1106 

This change will exclude `make` from auto-updating after creating the symlink to `gmake`. With the CentOS / RedHat system, the `yum` package of `make` is a dependency of `yum` itself and therefore can't be uninstalled, however if the updates are stopped, the symlink created won't be broken. 
This should stop machines reverting back to using an older version of `make`.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml
@@ -86,6 +86,9 @@
   ignore_errors: yes
   tags: goodmake_source
 
+
+# Disabiling updates to system 'make' to ensure the symlink above doesn't break.
+# See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1106
 - name: Check if /etc/yum.conf exists
   stat:
     path: /etc/yum.conf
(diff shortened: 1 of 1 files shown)


Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/689/ - putting through VPC

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gmake/tasks/main.yml
```diff
@@ -85,3 +85,40 @@
     - /tmp/make-{{ makeVersion }}.tar.gz
   ignore_errors: yes
   tags: goodmake_source
+
```
sxa, line 88: Could this be simplified down to two blocks:
```
if yum.conf does not exist
   create it with one line [main]
```
Then have the extra block to add `exclude=make` that gets executed regardless of whether it did the initial create or if it was already there?

Willsparker: I can make it so it: (each bullet is a single 'task')
- Check if /etc/yum.conf exists
- Creates /etc/yum.conf with [main] in it, if it doesn't exist
- Put `exclude=make` into /etc/yum.conf 
Is that what you meant? It'd be the same amount of tasks

sxa: Yes but it avoids duplicating the exclude line in two sections, so that is preferable. Please proceed :-)

Willsparker: Fixed :-)

PR #1405, opened 2020-06-22
pbTests: Hardcode JDK_BOOT_DIR in buildJDK.sh

Willsparker:
Ref: #1387 

I didn't realise that as part of the above PR we now symlink the JDK versions to `/usr/lib/jvm/jdk-XX` ( or `/usr/lib/jvm/jdk8`), like we do with the Windows playbooks. Due to this, the `find` functions that are being replaced in this PR pick up multiple entries, which confuse the build-scripts. (see all the Ubuntu runs and C6 / C7 in this `VPC` run: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/692/ )
Hardcoding this fixes this :-)

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -84,6 +84,7 @@ checkJDKVersion() {
 setBootJDK() {
         local buildJDKNumber=$(echo ${JAVA_TO_BUILD//[!0-9]/})
         local bootJDKNumber=$(($buildJDKNumber - 1));
+        # Refer to 8 as 'jdk8'. Anything else is 'jdk-XX'
         [[ $bootJDKNumber != "8" ]] && bootJDKNumber="-$bootJDKNumber"
         if [[ $buildJDKNumber -eq 8 ]]; then
                 # CentOS JDK7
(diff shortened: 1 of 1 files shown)


Willsparker:
Running through `VPC` on the OSs that it failed with in the run linked above: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/693/

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/694/
Seems fix the issue :-)

ansible/pbTestScripts/buildJDK.sh
```diff
@@ -84,6 +84,7 @@ checkJDKVersion() {
 setBootJDK() {
         local buildJDKNumber=$(echo ${JAVA_TO_BUILD//[!0-9]/})
         local bootJDKNumber=$(($buildJDKNumber - 1));
+	# Refer to 8 as 'jdk8'. Anything else is 'jdk-XX'
```
sxa, line 87: Indent on this looks wrong

Willsparker: Fixed :-)

PR #1406, opened 2020-06-22
Add gcc 7.5.0 install to arm32 Ubuntu16 machines (for build)

sxa:
Last nights builds with this change look good, so merging this which will bring arm32 builds up to the level of gcc used on other Linux platforms. NOTE that this will not install on raspbian at present, although I have tested with a tweaked version of this to ensure that it works

Fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/487


sxa:
Although the PR hasn't yet been merged this is now on the machines and activated, therefore I am closing.

Issue #1401, opened 2020-06-18
winPB: Make removal of speech mark in %PATH it's own task in Cygwin


Willsparker:
ping @lumpfish 

Currently, this task: https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/3143979688df2c97ea6b41347feaef8438809885/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml#L113

Will remove the stray speech mark in the Windows path before then adding cygwin's folder to the front of the path, however on some Windows systems this errors, with the following error message:

```
TASK [cygwin : Add c:\cygwin64\bin to front of %PATH%] *************************
fatal: [IP]: FAILED! => {"changed": true, "cmd": "setx /M PATH \"C:\\cygwin64\\bin;%PATH:\\\"=%\"", "delta": "0:00:00.218777", "end": "2020-06-17 01:27:23.008818", "msg": "non-zero return code", "rc": 1, "start": "2020-06-17 01:27:22.790040", "stderr": "ERROR: Invalid syntax. Default option is not allowed more than '2' time(s).\r\nType \"SETX /?\" for usage.\r\n", "stderr_lines": ["ERROR: Invalid syntax. Default option is not allowed more than '2' time(s).", "Type \"SETX /?\" for usage."], "stdout": "", "stdout_lines": []}
```
Running the command that is written in the task on the windows machine itself also didn't work, however running this manually did: `set NEWPATH=%PATH:"=% && setx /M PATH "C:\cygwin64\bin;%NEWPATH%"`  which can be 'ansibilised' as two separate tasks. This has the side benefit of being much clearer to read and understand what's going on.

Willsparker:
I've found a solution that would fix this issue- as in split out the tasks, as this makes it more readable for people who aren't as familiar with `cmd`/ Windows functions. I confirmed this does remove the speech marks as intended.
```
- name: Remove speech mark from the Path
  win_shell: 'setx /M PATH "%PATH:\"=%"'
  args:
    executable: cmd
  tags:
    - cygwin
```
[...]

PR #1408, opened 2020-06-23
winPB: Separate remove speech mark & add cygwin to front of %PATH% task

Willsparker:
Fixes: #1401 

FYI @lumpfish 

Separates the task `Add c:\cygwin64\bin to front of %PATH%` and `Remove speech marks from %PATH%`,  as it makes it clearer how each task is being done, for people who don't know `cmd` / Windows.
As well as this, making sure that these changes are made only as we install cygwin, as the path changes are only needed directly after it's installed, and rerunning the current `Add c:\cygwin64\bin to front of %PATH%` task causes an error.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -106,20 +106,30 @@
       - 'C:\cygwin64\bin'
     scope: machine
     state: absent
+  when: (not cygwin_installed.stat.exists)
   tags: cygwin
 
 # NOTE: This construct is because an SQL Server entry is being added to
 # the path which has an erroneous quote in it which causes problems for SETX
-- name: Add c:\cygwin64\bin to front of %PATH%
-  win_shell: 'setx /M PATH "C:\cygwin64\bin;%PATH:\"=%"'
+# See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1401
+- name: Remove speech mark from the %PATH%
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -106,20 +106,30 @@
...
     scope: machine
     state: absent
+  when: (not cygwin_installed.stat.exists)
   tags: cygwin
 
 # NOTE: This construct is because an SQL Server entry is being added to
 # the path which has an erroneous quote in it which causes problems for SETX
+# See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1401
+- name: Remove speech mark from the %PATH%
+  win_shell: 'setx /M PATH "%PATH:\"=%"'
```
lumpfish, line 116: Running from a command line on my test machine
```
setx TESTPATH "%TESTPATH:\"=%"
```
did not work (the double quote I had in TESTPATH was not removed).
```
setx TESTPATH "%TESTPATH:"=%"
```
(no backslash) worked (the double quote I had in TESTPATH was removed).

Willsparker: The `\` has to be in there to escape the `"` after it, otherwise Ansible interprets it as a literal `"` and fails. It effectively runs on the Windows machine as the second example you gave :-)

Willsparker:
Yep, ran through a VM under the same circumstances as @lumpfish 's, I was able to recreate the problem, and then implementing these changes stopped it :-)

PR #1411, opened 2020-06-24
aixPB: download packages from yum instead

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1391

Ive replaced the bullfreeware download links for perl, cmake, libiconv and libunistring with the usual yum install. Besides the cmake link, the other 3 were not working so needed to be replaced.

The packages download, without error, on one aix 7.1 machine on which ive tested.

### DIFF
--- a/ansible/playbooks/aix.yml
+++ b/ansible/playbooks/aix.yml
@@ -191,6 +191,10 @@
             - grep
             - libXrender-devel
             - libffi-devel
+            - libiconv
+            - libunistring
+            - perl
+            - cmake
             - make
             - m4
             - pcre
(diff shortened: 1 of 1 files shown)


sxa:
@Haroon-Khel Are you able to try running one of the system test suites on the machine (which make use of perl) to see if they can be run on a machine set up with the updated playbooks?

Haroon-Khel:
Following the steps [here](https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally#openjdk-tests), at the `make compile` step I keep getting this error 
```
configure-ant-1.10.2:
     [echo] Executing macro download-file
     [echo] File to download: http://archive.apache.org/dist/ant/binaries/apache-ant-1.10.2-bin.zip
     [echo] Destination: /tmp//apache-ant-1.10.2/apache-ant-1.10.2-bin.zip
     [echo] Download tool: wget
     [exec] Could not load program wget:
     [exec]     Dependent module /usr/lib/libiconv.a(libiconv.so.2) could not be loaded.
     [exec]     Member libiconv.so.2 is not found in archive 
     [exec] Result: 255
```

The machine has the module `libiconv` installed, and the directory `/usr/lib/libiconv.a` certainly exists on the machine. `libiconv` is one of the packages whose installation ive changed from a download link to yum in this pr, but this package, at version `1.16-1` was already present on the machine without the need for a yum install

Haroon-Khel:
After setting an appropriate LIBPATH variable, I managed to run a sanity.system test. Many of the tests failed due to insufficient space in `/home`. Despite this, the aim of this was to see if a system test could be run with a machine using perl v5.28.1 and it can

sxa:
Did the new machine described in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1434#issuecomment-659001278 not hit the same `libinconv` issue running `wget`?

Haroon-Khel:
It did, which prompted me to add `/opt/freeware/lib/` to the `LIBPATH`. Ill add something to the playbooks which modifies the `LIBPATH` accordingly, as I don't think there is currently anything in them that does this on aix machines

sxa:
I'm not comfortable with changing any system default on the `LIBPATH`. Doing things like that can make the testing less valid by hiding real issues that end users may see. We don't have it set on the existing machines so we should identify another way to resolve this

sxa:
> needs 3g more space @sxa if youre interested in adding more space to aoaix4

aoaix4 has 5Gb in /home now

Haroon-Khel:
Duplicated by https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1704

PR #1420, opened 2020-06-30
add cuda headers

gdams:
from a conversation with @keithc-ca at https://github.com/AdoptOpenJDK/openjdk-build/pull/1935#issuecomment-652039954

This allows us to build OpenJ9 using the dockefile/codebuild setup


karianna:
See https://github.com/AdoptOpenJDK/TSC/issues/88 - I'm not sure we can bundle these headers as we're not the final end user.

gdams:
closing as we cannot add CUDA to the public docker image for legal reasons

PR #1421, opened 2020-07-02
remove jenkins slave and add centos7

gdams:
We no longer need the slave image

Adds Centos7 image for later use


gdams:
@sxa updated PTAL

PR #1422, opened 2020-07-03
ansible: add which to centos packages

gdams:
The centos7 dockerfile doesn't come with which by default


gdams:
It's this line here in the jdk source: https://github.com/AdoptOpenJDK/openjdk-jdk15/blob/19c5bfe42ffe9ae217f2dcc514951146a7d08a8b/make/autoconf/configure#L114

Annoyingly it doesn't spawn a bash shell

PR #1429, opened 2020-07-06
travis: switch to bionic

gdams:


gdams:
merging as this will unblock the docker build failures

PR #1434, opened 2020-07-07
aixPB: Breakdown aix playbook into roles

Haroon-Khel:
Similar to https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/660, but more up to date

I've segmented the aix playbook into roles, similar to the Unix playbook, removing its monolithic structure. This makes it easier for users to make additions to the playbook, and moves towards a consistent directory structure across our Unix, Windows and Aix playbook directories.

Ive kept the original aix playbook file while I test this pr. Once this is ready to merge, I will delete the aix.yml file.


sxa:
What's the status of testing of this? If it's capable of setting up a machine from scratch and then a subsequent build and basic test case works then that's probably enough :-)

Haroon-Khel:
The playbook runs. Though I am getting issues with the `yum` role, which I suspect isn't related to changes I've been making. I get the error whenever I try to install a package which has `gettext-0.20.1-1.ppc` as a dependency, in this case `git`:
```
Error: Package: gettext-0.20.1-1.ppc (installed)
           Requires: libncurses.a(libncurses.so.5)
           Removing: ncurses-6.1-3.ppc (installed)
               libncurses.a(libncurses.so.5)
           Updated By: ncurses-6.2-1.ppc (AIX_Toolbox)
               Not found
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```
Any ideas?

Haroon-Khel:
Besides git being unable to be installed through yum, the rest of the playbook runs fine with the exception of the final swap role, which gives what I think to be a machine specific error:
```
        "0516-404 allocp: This system cannot fulfill the allocation request.",
        "\tThere are not enough free partitions or not enough physical volumes ",
        "\tto keep strictness and satisfy allocation requests.  The command",
        "\tshould be retried with different allocation characteristics."
```

Haroon-Khel:
@sxa On a new machine, I was able to get the playbook to run completely, without the failing `yum` task as above, though the `swap` role still fails in the same way as it does above. Do you know the use of the `swap` role? Has it ever worked?

While trying to build a jdk binary, during the stage where the build script clones the source code, I get the following memory error:

```
git clone --depth=1 https://github.com/adoptopenjdk/openjdk-jdk11u.git /home/openjdk-build/workspace/./build//src
Cloning into '/home/openjdk-build/workspace/./build//src'...
remote: Enumerating objects: 81820, done.
remote: Counting objects: 100% (81820/81820), done.
remote: Compressing objects: 100% (41007/41007), done.
remote: Total 81820 (delta 44882), reused 52150 (delta 33832), pack-reused 0
Receiving objects: 100% (81820/81820), 139.57 MiB | 2.62 MiB/s, done.
Resolving deltas: 100% (44882/44882), done.
fatal: Out of memory, malloc failed (tried to allocate 415685 bytes)
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

Do you think this is at all connected to the failing `swap` role?

Haroon-Khel:
On the plus side, basic tests cases do work on the machine. Sanity openjdk and system tests ran fine using a jdk11 openj9 binary

sxa:
Sorry it's taken me so long to follow up on this. I suspect the `swap` role is dependent on how the underlying disks are configured - if there's no space to allocate them using the commands in the swap role then it won't be able to set them up. As to whether the allocation error is related to that, my gut feel is "unlikely" but it would be useful to know whether (1) The machine has any swap space configured as you have it (2) How you're running that - if it's direct from the command line can you see if setting `LDR_CNTRL=MAXDATA=0x80000000` in the environment ([we do it for the builds here](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/platform-specific-configurations/aix.sh#L45))

Haroon-Khel:
Closing due to https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1633

PR #1438, opened 2020-07-08
WinPB: Update Cygwin_checksum

Willsparker:
Spotted by @lumpfish

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -20,7 +20,7 @@
     url: https://cygwin.com/setup-x86_64.exe
     dest: 'C:\temp\cygwin.exe'
     force: no
-    checksum: 2431de4597a2162f3e1d60af90f638b41677265b07cc66fcb0231fdde452c841
+    checksum: 4dd4d4531e8e63ade849daaaf587ba1c1430368701772c8ee42a27f4e8c373e4
     checksum_algorithm: sha256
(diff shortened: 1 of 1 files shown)


lumpfish:
Note also that the latest version of cygwin (3.1.5) cannot be used to build openj9 - see https://github.com/eclipse/openj9/issues/10109

Willsparker:
@karianna Yeah, this is 3.15, but as @lumpfish said it is currently broken for OpenJ9 - however I'm not really able to find an older 3.14 version - it seems Cygwin doesn't keep the older versions of `setup-x86_64.exe`.

lumpfish:
What I found is that there appears to be just one setup-x86_64.exe and using that you can choose which package versions you want to install - and cygwin and cygwin-devel are two such packages.  So I think that in theory you could just get the latest setup-86_64.exe but then alter the commands to get the specific versions of the packages .  Never done it on a command line though (and of course if you hard coded that into your ansible scripts you'd not be automatically upgraded to the latest cygwin version when installing a new machine).

Willsparker:
That is possible - we already do that with `make` and `grep` (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/c9902f959118f2f66d3470a31a7809213c904b6f/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml#L72) 
Unfortunately there is no way to specify versions via CLI and the specific packages need to be extracted to the correct place; I'm not so sure this is doable with the version of `cygwin` / `cygwin-devel`, though it's worth some investigation. Alternatively, we could wait until Cygwin 3.16, but there's no indication of when that will release.
For now though, this PR will allow me to keep using / testing `vagrantPlaybookCheck`, and at least the remainder of the playbook :-)

Issue #1444, opened 2020-07-14
Fix QEMU VMs for QEMUPlaybookCheck; Investigate adding ARM64 too


Willsparker:
Ref: https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/ , #1059 

For awhile the s390x machine at the referenced Jenkins job has not been able to complete its playbook due to what looks like missing packages (From https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/41/ARCHITECTURE=s390x,label=vagrant/ ):

```
install 'gnupg2'' failed: E: Failed to fetch http://gb.ports.ubuntu.com/ubuntu-ports/pool/main/o/openldap/libldap-common_2.4.45+dfsg-1ubuntu1.4_all.deb  404  Not Found [IP: 91.189.88.150 80]\nE: Failed to fetch http://gb.ports.ubuntu.com/ubuntu-ports/pool/main/o/openldap/libldap-2.4-2_2.4.45+dfsg-1ubuntu1.4_s390x.deb  404  Not Found [IP: 91.189.88.150 80]\nE:
```

ppc64le has also been failing but reporting a false positive ( from https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/41/ARCHITECTURE=ppc64le,label=vagrant/ ). It appears the build is failing due to `FILENAME` not being defined before running `make-adopt-build-farm.sh`
```
01:46:05 BRANCH:  (For release either BRANCH or TAG should be set)
01:46:05 TAG: 
01:46:05 FILENAME must be set in the environment
01:46:05 + [[ true == true ]]
```
As `QEMUPlaybookCheck` uses the same script to build the [JDK](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/pbTestScripts/buildJDK.sh) as `vagrantPlaybookCheck` , this should be fixed with #1442

When this Jenkins job was initially put in, I was unable to get `ARM64` to work as `ssh`ing to the machine wasn't working, so I'll aim to give this another go too.

Willsparker:
With S390x, I was able to replicate the problem, and fixed it by running `sudo apt-get update` before attempting to install `gnupg2`. I'll add `update_cache: yes` to the task to do this.

Willsparker:
QEMUPlaybookCheck run to determine any other issues with s390x: 
https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/42/

[...]

PR #1456, opened 2020-07-21
UnixPB: Install docker.io for Debian, Add gcc-7 to Common role for aarch64 

Willsparker:
ref: #1444  ; https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/46/ARCHITECTURE=arm64,label=vagrant/

For Debian10/aarch64 , `docker-ce` isn't an available package, however `docker.io` is : https://packages.debian.org/buster/arm64/docker.io/download

Running through QEMUPlaybookCheck here: https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/47/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Debian.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Debian.yml
@@ -11,9 +11,11 @@ Build_Tool_Packages:
   - build-essential
   - cpio
   - curl
-  - flex                          # Openj9
+  - flex                          # OpenJ9
   - g++
+  - g++-7                         # OpenJ9
   - gcc
+  - gcc-7                         # OpenJ9
   - gettext
   - libasound2-dev
   - libcups2-dev
@@ -74,18 +76,12 @@ Additional_Packages_Debian8:
 Additional_Build_Tools_x86_64:
   - libnuma-dev
   - numactl
-  - gcc-7
-  - g++-7
 
 Additional_Build_Tools_ppc64le:
   - libnuma-dev
   - numactl
-  - gcc-7                         # OpenJ9
-  - g++-7                         # OpenJ9
 
 Additional_Build_Tools_s390x:
-  - gcc-7                         # OpenJ9
-  - g++-7                         # OpenJ9
   - numactl
   - libfreetype6-dev              # Needed by test state=installed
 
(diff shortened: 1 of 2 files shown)


sxa:
Any reason why this needs to be specific to aarch64? If we can use `docker.io` on all architectures I'd suggest doing so

Willsparker:
No reason other than I haven't tested other architectures yet - I know `docker-ce` at least works for Debian10 on x86_64 (looking at `VPC`) , so I'll investigate if `docker.io` works too :-)

aahlenst:
`docker-ce` usually refers to the upstream package (see https://docs.docker.com/engine/install/debian/), `docker.io` is usually bundled with the distributions. `docker-ce` is available on significantly less platforms (amd64, armhf, arm64).

Willsparker:
Alright, makes sense to just use `docker.io` then! Cheers

Willsparker:
Marked the PR as a draft as I'll just add to this PR until we get a clean ARM64 run. 
In it's current state, Debian10 aarch64 doesn't install any version of gcc-7 / g++-7. This is easily rectified by just adding it to the list of packages in the `Common` role.

Willsparker:
With the latest commit, I've moved `gcc-7` and `g++-7` to the common build tools, as the addition of it to `aarch64` means it would appear in every architecture specific list of packages, which is inefficient

Willsparker:
> Can we remove the section on `- name: Add Docker repo for Debian x86_64` with this change?
> As long as it still works on Debian 8 we shoule ok

Unfortunately, Debian 8 (x86_64) needs it: 
```
vagrant@adoptopenjdkd8:~$ sudo apt install docker.io
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package docker.io is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```
I tried running `sudo apt update` too, to no avail

sxa:
Could we lock the installation of the repo down to `Debian >=10` ?

Willsparker:
Done with the latest commit - I've also restricted added the GPG key as well as that isn't needed if the repo isn't being added :-)

sxa:
Running again at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/742/ to test the changes across both Debian versions

sxa:
Looks good - happy for this to be approved if taken out of draft

Willsparker:
Ah ...Okay, I thought Debian 8 had `docker.io` ...

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/749/ 
Both Debians have got passed the `Docker` role

PR #1464, opened 2020-07-27
UnixPB: Install cmake on aarch64

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1462

This pr includes a condition in the cmake role to install cmake on aarch64

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
@@ -31,7 +31,7 @@
     checksum: sha256:8f864e9f78917de3e1483e256270daabc4a321741592c5b36af028e72bff87f5
   when:
     - (cmake_installed.rc != 0 ) or (cmake_installed.rc == 0 and cmake_version.stdout is version_compare(cmakeVersion, operator='lt') )
-    - (ansible_architecture == "x86_64") or (ansible_architecture == "ppc64le") or (ansible_architecture == "s390x") or (ansible_architecture == "aarch64")
+    - ansible_architecture != "armv7l"
     - (ansible_distribution != "openSUSE")
   tags: cmake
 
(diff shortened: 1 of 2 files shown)


Haroon-Khel:
@Willsparker does the vpc job test on aarch64?

sxa:
@Haroon-Khel [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) uses VirtualBox so is only able to do x64 operating systems. We are now doing [tests using qemu](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) on which we've only just recently added the aarch64 support, however I think OpenJ9 have only just enabled cmake which is why this is now an issue.

gdams:
@Haroon-Khel the travis job does run arm64 testing (it builds the dockefile) see https://travis-ci.com/github/AdoptOpenJDK/openjdk-infrastructure/jobs/365313217

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
```diff
@@ -31,7 +31,7 @@
     checksum: sha256:8f864e9f78917de3e1483e256270daabc4a321741592c5b36af028e72bff87f5
   when:
     - (cmake_installed.rc != 0 ) or (cmake_installed.rc == 0 and cmake_version.stdout is version_compare(cmakeVersion, operator='lt') )
-    - (ansible_architecture == "x86_64") or (ansible_architecture == "ppc64le") or (ansible_architecture == "s390x")
+    - (ansible_architecture == "x86_64") or (ansible_architecture == "ppc64le") or (ansible_architecture == "s390x") or (ansible_architecture == "aarch64")
```
sxa, line 34: This covers all linux architectures other than armv7l (arm32) right? So I would suggest modifying this conditional to say all architectures other than armv7l to make it easier to scale this.

However it's probably worth looking at whether we even need an arm32 specific clause here. Perhaps we need to check whether cmake 3.11.4 is already in the package repositories for various distributions and if so add it into the Common role where we install stuff through the package manager. If we do that in a way that installs on armv7l/arm32 then I suspect the clause at the start of this role which checks for cmake being installed would trap that and mean we don't need all these conditionals.

Haroon-Khel:
Ive checked yum, zypper and apt. None have cmake v 3.11.4 in their package repos. Im not sure which package manager our only freebsd machine uses as I do not have access to it, `build-spearhead-freebsd12-x64-1`

sxa:
Which distributions did you test that on?

Haroon-Khel:
@sxa Ive tested this on ubuntu x86, cmake installs as expected. The travis job runs it on aarch64; cmake installs. I dont have an arm32 machine to test this on, perhaps this is something we can add to the qemu playbook check job in the future

sxa:
@Haroon-Khel I think you just volunteered to set one up ;-)

PR #1465, opened 2020-07-27
pbTests: Refactor searchLogFile functionality; Add logging for builds and tests

Willsparker:
Start to log builds and tests in a `VPC` run. This allows those logs to be searched as well, thereby stopping a `VPC` run from going onto testing if a build failed. This also stops false positives in Jenkins.
In additional, some other changes have been made:
- Adding the additional `.\ConfigureRemotingForAnsible.ps1` script calls into the Windows Vagrantfile. Previously, only the one to enable `Credssp` was ran, which did seem to suffice- however I've added these as it is required when setting up machines, and also seems to make connecting to the Windows VM via Ansible more stable.
- Change the `sed` commands that were used to add lines to the `ansible.cfg` file to an `awk` command, as this stops the requirement of `GNU sed`. (Thanks @sxa , for figuring that one out)

PR #1452, opened 2020-07-20
pbTests: Add support for ARM64 in QEMUPlaybookCheck

Willsparker:
Fixes: #1444  (This PR in conjunction with #1447 will fix 1444)

Add support for testing our playbooks on ARM64 Debian Buster through the `QEMUPlaybookCheck` Jenkins job. 
In addition to the changes done here, the `infra-softlayer-ubuntu1804-x64-1` machine has been updated to have the `ARM64.dsk.xz` file in `/qemu_base_images`. Like the other architectures, the disk image is 15GB, and has the user `linux` on it to run the playbook on.
The guide I used to setup a working Debian Buster ARM64 QEMU VM is from here https://wiki.debian.org/Arm64Qemu , using the `debian-10.4.3-20200610-openstack-arm64.qcow2` image found at https://cdimage.debian.org/cdimage/openstack/current/ .
I tested this on the `infra-softlayer-ubuntu1804-x64-1` machine and it ran through, though I can test it by configuring the `QEMUPlaybookCheck` job to pull from my git branch.

PR #1473, opened 2020-07-28
pbTests: Add logging to QEMU builds/tests & Rename ARM64 to AARCH64

Willsparker:
Ref: #1465 #1452 #1444 

Add logging to the build and test branches in `qemuPlaybookCheck.sh` to stop false positives (as it stops running a test if a build fails).  
If a build or test fails, it exits with error code `127` - the Jenkins jobs have been configured to report `unstable` if the build or tests fails, but `failure` if the playbook fails.

Tested at:
https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/60/
https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/61/. (The Jenkins job has been configured to report unstable if `exit 127` since this job ran)

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -165,7 +165,7 @@ if grep 'buster' /etc/*-release >/dev/null 2>&1; then
 	export CXX=/usr/bin/g++-7
 fi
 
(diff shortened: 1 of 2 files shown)


Willsparker:
Here's an example of it working where a build fails so it reports `unstable` : 
https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/62/ARCHITECTURE=aarch64,label=vagrant/console

PR #1474, opened 2020-07-28
Enable EPEL repo for rhel 6

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1470

Added a task into `Redhat.yml` which adds the EPEL repo for rhel 6.

Tested on a rhel 6 x86 machine. EPEL repo and `libdwarf-devel` are both installed without error. Im unable to test this on other architectures as only x86 rhel 6 machines are available to me through fyre. However, according to https://fedoraproject.org/wiki/EPEL#What_packages_and_versions_are_available_in_EPEL.3F, for rhel 6, EPEL is only available on x86 and i386. Hence I have added a condition which only installs the EPEL repo on x86

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
@@ -27,7 +27,7 @@
 
 - name: Enable EPEL release for RHEL6
   yum: name=http://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
-  when: 
+  when:
     - ansible_distribution_major_version == "6"
     - ansible_architecture == "x86_64"
   tags: patch_update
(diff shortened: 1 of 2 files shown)


Haroon-Khel:
Is this also an issue on Centos 6?

Willsparker:
@Haroon-Khel https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/9a8939be81093e4d91784cfe42dc2008204b9fda/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml#L9
Don't believe so

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -25,6 +25,13 @@
     - ansible_distribution_major_version == "8"
   tags: patch_update
 
+- name: Enable EPEL release for RHEL6
```
Willsparker, line 28: This is doing the same thing as the task above it- just modify the conditions, as it already has the `{{ ansible_distribution_major_version }}` variable in it :-)

Haroon-Khel: Good point

Haroon-Khel:
The `ignore_errors: yes` can ignore the cases when installing the epel repo for rhel 6 on unsupported architectures as mentioned above

Issue #1472, opened 2020-07-27
ccache built from source if it is already in /usr/local/gcc/bin


sxa:
In most linux cases we use a gcc-7 extracted from a prebuilt tarball which includes a copy of ccache.
The ccache role only checks if there is a version in `/usr/local/bin` before deciding whether to build it's own. We should check for that location OR /usr/local/gcc/bin (Or perhaps symlink /usr/local/bin/ccache -> /usr/local/gcc/bin/ccache during the `gcc_7` role instead.

Willsparker:
I'll look into this :) 
I think symlink-ing would be the best option- presumably the build scripts are setup to search for `ccache` at set places so they may not be able to find that version, even if `ansible` does.

sxa:
On the appropriate builds we add `/usr/local/gcc/bin` into the PATH https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/platform-specific-configurations/linux.sh#L135 so it's not needed but I'm ok with the symlink

PR #1476, opened 2020-07-28
UnixPB: Symlink ccache to stop building from source

Willsparker:
Fixes: #1472

Stops `ccache` being built from source when it is extracted from the binary retrieved in the `gcc_7` task. 

VPC Run: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/773/
(Only CentOS VMs are ran as they're the only ones that have the `gcc_7` task ran on them)


Willsparker:
Rerun : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/775/

Willsparker:
Okay, I've put in a task that checks if anything is at `/usr/local/bin/ccache`. Also changed the tags to `gcc_7` as opposed to `gcc-7`, to stay consistent with the name of the role.

PR #1477, opened 2020-07-29
UnixPB: Fix path variable for cmake on rhel6

Haroon-Khel:
Despite it being there from the gcc_48 role in the playbook, the system cannot detect gcc-4.8.5 on the machine when building cmake during the `cmake` role, resulting in the error:
```
        "C compiler on this system is: gcc       ",
        "---------------------------------------------",
        "Error when bootstrapping CMake:",
        "Cannot find a C++ compiler supporting C++11 on this system.",
```
Placing the included PATH variable before ./configure seems to have fixed this.

I received this error on two different rhel 6 machines, and this fix worked for both of them.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/cmake/tasks/main.yml
@@ -48,7 +48,7 @@
 
 # NOTE: PATH setting is for RHEL6/CentOS6 since gcc 4.4.7 is not C++11 compliant
 - name: Running ./configure & make for cmake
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
@Willsparker Could you run a vpc job for this pr to ensure that its compliant with other distros

Willsparker:
Here ya go: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/776/

Haroon-Khel:
Passed on all distros aside from Debian8 and Freebsd12, but it looks like those failures are unrelated to this pr

Haroon-Khel:
@sxa It seems to. I just tested it again, on a machine without cmake. I removed the PATH variable entirely, failed as expected. I  then returned it as is and the playbook ran fine

Issue #1469, opened 2020-07-27
Windows ansible script fails at Windows Update step


lumpfish:
Running the ansible playbook on a newly created Windows Server 2012 R2 machine failed at the Windows Update step with:
```
TASK [Windows_Updates : Download and Install Windows Updates] ******************
fatal: [9.20.198.24]: FAILED! => {"changed": false, "filtered_updates": {}, "msg": "Failed to search for updates: Exception calling \"Search\" with \"1\" argument(s): \"Exception from HRESULT: 0x80072EE5\"", "updates": {}}
...ignoring

TASK [Windows_Updates : Reboot machine if necessary] ***************************
fatal: [9.20.198.24]: FAILED! => {"msg": "The conditional check 'update_result.reboot_required' failed. The error was: error while evaluating conditional (update_result.reboot_required): 'dict object' has no attribute 'reboot_required'\n\nThe error appears to be in '/disk2/git/openjdk-infrastructure/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Windows_Updates/tasks/main.yml': line 11, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Reboot machine if necessary\n  ^ here\n"}
```

Willsparker:
I looked into this yesterday and I think this is an issue with the provisioning service that @lumpfish and I use to create the windows machines. It seems to be an intermittent issue as something can be done to make it work, for instance: Manually run windows update, checking for updates from the internet; opening Internet Explorer; Just waiting for awhile - though none of these consistently fix it.

However, the issue is replicated when you manually check for updates that are specifically managed by your system administrator. You're able to update the system from updates found online so if anyone else encounters this problem, I'd suggest doing that, rebooting and skipping the `Windows_Updates` task on the playbook. Closing issue

sxa:
Let's hold off on closing until @lumpfish has verified it works for him :-)

Willsparker:
Closing due to @lumpfish 's confirmation that #1478 works :-)

PR #1478, opened 2020-07-29
winPB: Ignore errors on win_update reboot task

Willsparker:
Ref: #1469 

While the above seems to be an isolated issue for those particular machines, it's shown that there's no point having `ignore_error: yes` on the first task, if it doesn't have it on the second task, as the `update_result` variable is undefined if the first task fails. This also allows for workarounds to be done for the group of machines that the referenced issue affects without the playbook failing or having to add `Windows_Updates` to the `ansible-playbook` `--skip-tags` option.

ping @lumpfish

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Windows_Updates/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Windows_Updates/tasks/main.yml
@@ -11,5 +11,6 @@
 - name: Reboot machine if necessary
   win_reboot:
     reboot_timeout: 3600
+  ignore_errors: yes
   when: update_result.reboot_required
   tags: Windows_Updates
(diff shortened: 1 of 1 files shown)


Willsparker:
@sxa So - Windows updates are divided up into 2 types of updates: Ones managed by the system administrator and ones that can be retrieved by getting the updates via the internet. The playbook fails because there's something wrong with the system administrator ones- which, I presume, are handled by the provisioning service we used.
I was able to replicate the issue on the machine locally, so really, it has nothing to do with our playbooks- however I want to make the change I'm doing in this PR as it's illogical to have `ignore_errors: yes` on the first task but not the second.

Willsparker:
@sxa Commit in :-)

lumpfish:
I'm rerunning this latest code on a new Windows 2012 Server machine.  It is now past the previous failure and still running as I write this.

lumpfish:
Confirming that the script ran to completion on my Windows Server 2012 R2 machine.

PR #1487, opened 2020-08-03
unixPB: Pull gcc48 source from an https URL

sxa:
Because https is always better than http ;-)


sxa:
Testing at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/789 but since the file is identical it shouldn't be a problem :-)

Willsparker:
( Restarted it at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/790/ )

PR #1493, opened 2020-08-05
Fix URL for Docker GPG key

aahlenst:
Furthermore, I changed the URL to the one recommended in the Docker docs (see https://docs.docker.com/engine/install/debian/).

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml
@@ -212,7 +212,7 @@
 
 - name: Add Docker GPG apt key for Raspbian
   apt_key:
-    url: htps://yum.dockerproject.org/gpg
+    url: https://download.docker.com/linux/debian/gpg
     state: present
   when:
     - docker_installed.rc != 0


aahlenst:
Further down, they state:

> Note: This procedure works for Debian on x86_64 / amd64, Debian ARM, or Raspbian.
> (...)
> $ sudo apt-get update
> $ sudo apt-get install docker-ce docker-ce-cli containerd.io

So they seem to contradict themselves. Considering that I've provisioned a Raspbian 10 machine with the playbook and https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/87ad1d3bb2982ab650f347b7de9cbbed295a853d/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml#L297-L305 installs via apt, should I really rework the playbook to use that script?

aahlenst:
I need to correct myself: Because it's Raspbian 10, I actually got `docker.io`, the Debian package. So the relevant section for my change would be https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/87ad1d3bb2982ab650f347b7de9cbbed295a853d/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Docker/tasks/main.yml#L287-L295.

Willsparker:
Hmm okay- If it works, I think this is the best way to do it then, for the sake of consistency among the tasks.

Issue #1485, opened 2020-08-03
WinPB: Stray speech-mark still appears in %PATH% after the playbook is run.


Willsparker:
Ref: 
https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1205#issuecomment-606037676
https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1401
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1408

It appears that on a fresh system, a speech mark still appears after the playbook is run. This can cause issues with the part of the build that applies the Visual Studio variables to the environment: 
```
06:45:10  configure: Rewriting VS_ENV_CMD to "/cygdrive/c/progra~2/mib055~1/2017/commun~1/vc/auxili~1/build/vcvars~2.bat"
06:45:10  configure: Trying to extract Visual Studio environment variables
06:45:12  \Microsoft was unexpected at this time.
06:45:12  configure: Could not succesfully extract the envionment variables needed for the VS setup.
```

Willsparker:
I have been able to recreate the issue. In addition, the stray speech mark causes `OpenSSL` to not be built. I hypothesise it's because the speech mark in the following code block causes the command to finish prematurely: 
```
- name: Install OpenSSL-1.1.1g 64-bit (VS2013)
  win_shell: set PATH=C:\Strawberry\perl\bin;C:\openjdk\nasm-2.13.03;%PATH% && ... etc
```
SO all that does is set the path. 
The alternative is that when it gets to executing the `MSVS` shell script that is used to set the environment variables, it errors (with the same `\Microsoft was unexpected at this time` error message above).

Willsparker:
After running through several iterations (and coming close to madness), I abandoned the idea of using `cmd` and `setX /M`
For some reason, when running through Ansible, it refused to have correct syntax and kept giving the following error message:
```
ERROR: Invalid syntax. Default option is not allowed more than '2' time(s).
```
I think this may be to do with the double quotes in the path that is messing up the syntax of `setX /M ...`. Even going back to how it was before #1408 wouldn't succeed via Ansible - it did locally, but not with Ansible. This sort of makes me question if the task to remove the speech mark ever worked in the first place.

Anyway - I've opted for a solution that uses `powershell` , which I think also looks easier to understand what it's doing: 
```
- name: Remove double quotes from Path & Add C:/cygwin64/bin to front of path
  win_shell: |
[...]

PR #1495, opened 2020-08-06
WinPB: Fix task to remove quotes from path after cygwin install

Willsparker:
Fixes: #1485 
ping @lumpfish 

Uses Powershell to remove the speech mark & put `C:/cygwin64/bin` to the beginning of the path, as the `setX /M` command didn't appear to work. When the speech mark was present in the Path, it also stopped building OpenSSL due to the speech mark causing the `./vcvarsall` script (or equivalent) to not work - however without producing an error message (otherwise this would've been spotted earlier).

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2010/tasks/main.yml
@@ -29,6 +29,18 @@
     - c:\Program Files (x86)\Microsoft Visual Studio 10.0\DIA SDK\bin\amd64\msdia100.dll
   tags: MSVS_2010
 
+# NOTE: This construct is because an SQL Server entry is being added to
+# the path which has an erroneous quote in it, causing OpenSSL not to build
+# See: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1485
+- name: Remove double quote from %PATH%
+  win_shell: |
+    $NewPath=$env:path.Replace('"','')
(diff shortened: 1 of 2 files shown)


lumpfish:
@Willsparker - when is the erroneous double quote added?  Is it there before we start running the playbook or does something we are installing at it.
I don't think cygwin has anything to do with it, so adding it to the cygwin task doesn't seem right.  Rather it would seem to 'belong' to the end of the task(s) which adds the double quote or if it's already there on a new install, right up front in the playbook before we do anything else.

Willsparker:
Looking at newly provisioned machine, the speech mark isn't there. I'll run through the playbook to see what could add it. The playbook comment says it's added with Microsoft SQL Server - so I'm thinking MSVS_2010, but I'll confirm.

Willsparker:
Right, I've had a look - funnily enough, I can't pin down a task that adds the speech mark. However, (rather embarrassingly), the solution I put in with #1401 appears to add a speech mark when there isn't one.
However, before I put that fix in, _something_ was adding a speech mark to the path at some point, as it affected the `softlayer-2012r2` machines which were setup before that was put in ( https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1205#issuecomment-606037676 ), so I still think it should be in the playbook, however I'm not really sure where.
The [original fix](https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/796) said the speech mark was added with an SQL server entry- which lines up with what I've been seeing in my testing too.

sxa:
If it is a result of the SQL Server setup in one of the Visual Studio roles would this not be best put into that role?

PR #1496, opened 2020-08-09
ansible: add alpine build image

gdams:
Allows us to build musl openjdk binaries

See https://github.com/AdoptOpenJDK/TSC/issues/168 for more information


sxa:
As long as this doesn't break anything (checking via https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/813/ but it'll be later today before it's scheduled) I'm happy to put this in

PR #1501, opened 2020-08-11
winPB: convert 'raw' modules to 'win_shell' when making symlinks

Willsparker:
Use `win_shell` instead of the `raw` module when creating symlinks in the Windows Playbook. This was prompted by getting an error message whilst investigating #1485 : 
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: winrm.exceptions.WinRMError: Access is denied.  (extended fault data: {'transport_message': 'Bad HTTP response returned from server. Code 500', 'http_status_code': 500, 'wsmanfault_code': '2147942405', 'fault_code': 's:Receiver', 'fault_subcode': 'w:InternalError'})
fatal: [9.20.194.40]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout":
```
This seems to only be specific to the Windows machines I was provisioning (@lumpfish didn't get this issue when using machines from the same provider), however I found the `win_shell` module didn't cause this error message. In addition, the `raw` module isn't meant to be used when it can be avoided: https://docs.ansible.com/ansible/latest/modules/raw_module.html#synopsis. (It's main use seems to be communicating with a guest that doesn't have `python` installed). Due to this, I thought I'd just change them anyway :-)

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_32bit/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Clang_32bit/tasks/main.yml
@@ -37,7 +37,9 @@
   tags: clang_32bit
 
 - name: Create symlink to C:\openjdk\LLVM32
-  raw: cmd /c mklink /D "C:\openjdk\LLVM32" "C:\Program Files (x86)\LLVM"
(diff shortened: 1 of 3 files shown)


sxa:
Pretty sure the travis errors are unrelated as they're related to CentOS7 docker image creation so I'm OK to merge this.

PR #1503, opened 2020-08-11
WinPB: install Cygwin using a cache

Willsparker:
Fixes: #1210 
Ref: #1231 , #1219 

The cached used to install Cygwin in the #1219 was flawed. The cache was created using 7-Zip which apparently doesn't retain symlinks when creating archives. In addition, extracting the archive using 7-Zip can cause issues with the paths to the symlinks , i.e. `/etc/ssl/certs/` symlinked to `\etc\pki\tls\certs\` would go to `/etc/ssl/certs` to `C:/etc/pki/tls/certs` for some reason.
To get round this, @sxa had the idea to use `tar` and `xz` built by cygwin ( located at https://ci.adoptopenjdk.net/userContent/winansible/cygwin64/bin/ ), which can be extracted by 7-Zip, which can then be used to extract the main archive.
I've been able to prove the playbook works with https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/801/ , however the build failed, however seemingly not due to a cygwin issue. Therefore, I'll keep this PR in draft until I can get an example of a working build.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -17,9 +17,9 @@
 
 - name: Retrieve cygwin cache from ci.adoptopenjdk.net
   win_get_url:
-    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64.tar.xz
-    dest: C:/temp/cygwin64.tar.xz
-    checksum: e89608178b4d99042fd356687c84aa18e220fb0dcd0c9d4bc60983ce538d9efe
+    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64.tar.gz
+    dest: C:/temp/cygwin64.tar.gz
+    checksum: 8f8b1d7d00233e384e725390c4018941fe7125290b94ecc0eeb3cf6684b4ce23
     checksum_algorithm: sha256
   when: (not cygwin_installed.stat.exists)
   tags: cygwin
@@ -28,23 +28,17 @@
   win_get_url:
     url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64/bin/*zip*/bin.zip
     dest: C:/temp/cyg_toolset.zip
-    checksum: 5974b93eefe033bd57cc14c30f7ba67584a5b9249b71cede7d9e05d87113d908
+    checksum: 39a94bb8ac6e5810437d7928381943d9f3a811729ad05d2967a5a7da8bce5b7d
     checksum_algorithm: sha256
   when: (not cygwin_installed.stat.exists)
   tags: cygwin
 
-- name: Extract cygwin_toolset to C:/temp
+- name: Use cygwin 'tar' and 'pigz' to extract cygwin cache
   win_shell: |
-    cd C:\temp
-    C:\7-Zip\7z.exe x C:\temp\cyg_toolset.zip
-  when: (not cygwin_installed.stat.exists)
-  tags: cygwin
-
-- name: Use cygwin 'tar' to extract cygwin cache
-  win_shell: |
-    $env:path="C:\temp\bin;$env:Path"
+    Expand-Archive -LiteralPath C:\temp\cyg_toolset.zip -DestinationPath C:\temp;
+    $env:path="C:\temp\bin;$env:path"
     cd C:\
-    tar --force-local -xvf C:\temp\cygwin64.tar.xz
+    tar --force-local -I pigz -xf C:\temp\cygwin64.tar.gz
   args:
     executable: powershell
   when: (not cygwin_installed.stat.exists)
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -15,6 +15,7 @@
   register: cygwin_installed
   tags: cygwin
 
+# For details about the cache, see: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1210#issuecomment-679076545
 - name: Retrieve cygwin cache from ci.adoptopenjdk.net
   win_get_url:
     url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64.tar.gz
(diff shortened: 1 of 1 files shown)


Willsparker:
Right, finally managed to get a JDK built with a cached cygwin install, however it was locally. Ready to Review

gdams:
I've been using this cache file in https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1513 and the tar command is insanely slow... I'm wondering if we can make it more efficient or opt for a tool that can extract in parallel?

aahlenst:
pigz does that well. It's parallel gzip. If size is not an issue, ISOs should retain symlinks.

Willsparker:
Converting to draft to look into using `pigz` instead, and also replace the use of 7-Zip with:
```
Expand-Archive -LiteralPath C:\temp\cyg_toolset.zip -DestinationPath C:\temp;
```
As suggested by @gdams

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/838/ 
Ready for review - The addition of `pigz` and using `Expand Archive` worked and reduced the cygwin install down to about 18 minutes, as opposed to the 30-35 minutes it takes for a standard install.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -15,88 +15,37 @@
...
   tags: cygwin
 
-- name: Download Cygwin installer
+- name: Retrieve cygwin cache from ci.adoptopenjdk.net
   win_get_url:
-    url: https://cygwin.com/setup-x86_64.exe
-    dest: 'C:\temp\cygwin.exe'
-    force: no
-    checksum: 4dd4d4531e8e63ade849daaaf587ba1c1430368701772c8ee42a27f4e8c373e4
+    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64.tar.gz
```
aahlenst, line 20: Is it obvious where the tarball comes from and how it's created?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -15,88 +15,37 @@
...
-# grep version.
-###############################################################################
-- name: Download grep-2.27-2
+- name: Retrieve cygwin toolset from ci.adoptopenjdk.net
   win_get_url:
-    url: "{{ Cygwin_SITE_URL + '/x86_64/release/grep/grep-2.27-2.tar.xz' }}"
-    dest: "{{ Cygwin_ROOT_INST_DIR + '\\tmp\\grep.tar.xz' }}"
-    force: no
-    checksum: 36ae2483f68b3c5a4bd93814fa4f7e2576dbe6559d4ffa0589c2159e85de91f6
+    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64/bin/*zip*/bin.zip
```
aahlenst, line 29: `*zip*` looks suspicious. Is there some wildcard matching involved?

aahlenst (commented):
My first playbook review. Please ignore the review if not helpful.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
Willsparker: That was just the link given to me when I clicked 'all files in zip' at https://ci.adoptopenjdk.net/userContent/winansible/cygwin64/ . I would assume it's just the way Jenkins generates the link, I don't think it's any wildcard matching :-)

gdams: Yeah that `*zip*` puts all the binaries located in https://ci.adoptopenjdk.net/userContent/winansible/cygwin64/bin/ into a single zip file.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
Willsparker: I've been trying to keep a step by step in the linked issue (#1210) , which includes creating the cache- however, I'll write a comment explaining how it was made and link it above the task to make it more obvious

aahlenst (approved):
For me it's good to go.

sxa:
Interesting - i wouldn't have thought pigz would provide such a significant boost in this scenario compared to the download and disk writing that's done as part of this

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
sxa: I'd prefer us to just create a zip file of the files required to bootstrap this and use that. When I started on this i didn't expect there to be so many files that were going to be required :-)

sxa (commented):
I can't help feel we could save a fair bit of additional time by streaming this i.e. not saving the intermediate file locally although when I tried this on Windows it kept choking part way through the download ...

Willsparker:
@sxa I'm more than happy to try - I'll create a git branch off this PR to do some testing, and post links to `VPC.sh` runs 👍 
How were you 'streaming' the cache? I can start there

sxa:
`curl https://ci.adopt.../cygwin64.tar.gz | gzip -d | tar xf -` (or replace `curl` with `wget -O` if preferred) ideally from a bash shell of some sort but I think I tried in various shells with no success :-(

Willsparker:
Ah okay- the following `.dll`s are missing that stop me from starting the following `.exes`
```
bash - cygreadline7.dll
curl - cygcurl-4.dll
wget - cyggnutls-30.dll
```

sxa (approved):
On the basis this has been shown to be quicker I'll approve it, but I hope we can get it to stream the download/extract more efficiently at some point ...

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -15,88 +15,38 @@
...
```
sxa, line 30:
```suggestion
    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64_bootstrap.zip
```

sxa (changes requested):
Correcetion: If this change goes in to use the zip then I'll approve :-)

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
Willsparker: URL & Checksum updated with most recent commit :+1:

sxa (approved):
Approving, but let's do a final validate through VPC before you merge this :-)

Willsparker:
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/861/ on it :-) This will fail due to the current MSVS_2019 issue ( #1526 ), but that's after the cygwin install, so is still a viable check.

Willsparker:
The above `VPC` run failed - The issue is that the `cygwin_bootstrap` zip is not unpacking into a folder, but I'll confirm this on a VM (Once I've got my laptop setup for VMs ... )

Willsparker:
Rerunning now : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/862/

Willsparker:
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/862/
This went through okay, so I think it's good to merge :-)

PR #1507, opened 2020-08-13
ansible: remove adoptopenjdk tag from riscv task

gdams:
It doesn't seem to be an adoptopenjdk setup specific task and it's currently getting skipped in the dockerfile builds


sxa:
The tag was added so that non-adoptopenjdk users looking to set up a machine for building x64 binaries don't have their machines polluted with a tool they do not need.

PR #1508, opened 2020-08-13
ansible: add adoptopenjdk tag back to riscv role and explicitly run o…

gdams:
…n docker

### DIFF
--- a/ansible/Dockerfile.CentOS6
+++ b/ansible/Dockerfile.CentOS6
@@ -12,7 +12,8 @@ RUN echo "localhost ansible_connection=local" > /ansible/hosts
 
 RUN set -eux; \
  cd /ansible; \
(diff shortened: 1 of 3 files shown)


Willsparker:
Can you use both `--tags` and `--skip-tags`? I haven't tried that before and I can't see it in the documentation: https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html

gdams:
hmm apparently not... that PR job failed

gdams:
hmm yeah @sxa that is not going to work... I'm not sure how else to proceed here?

gdams:
Trying another solution, it's a bit of a hack having to run the `ansible-playbook` command twice but I think it's the only way to make this work...

sxa:
Maybe adjust it to do a second playbook run that only runs that one role? It won't take effect on any systems other than Cent6 so will end up as a no-pop on other distributions

gdams:
okay, @sxa @Willsparker it worked! PTAL

PR #1517, opened 2020-08-20
winPB: Remove MSVS_2019 checksum

Willsparker:
Removing the checksum as it's changed twice in the span of a few days!
ref: 
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/835/OS=Win2012,label=vagrant/console
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1481#discussion_r472075910

Removing the checksum all together should save on PRs that just change the checksum.


sxa (approved):
Could do with a better way to manage this (which may be caching the installer so it doesn't change...) but for now I'm ok with this.

PR #1525, opened 2020-08-28
Fix alternatives configuration on RHEL flavours

aahlenst:
Previously, a new alternative was created for java/javac that did not configure the slave commands (for example, keytool). This is not necessary, because the RHEL packages being used properly declare their alternatives. Futhermore, it made a lot of standard tools like keytool unavailable.

I'm unsure about the state of Debian/Ubuntu: Ubuntu has the same problematic configuration in https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml#L137-L161. But the proper way there is to use `update-java-alternatives`, not `alternatives`. The Debian playbooks don't do this at all. Happy to address that in a separate PR if someone helps me understand the thinking there.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
@@ -182,44 +182,49 @@
 ####################
 # Set Default Java #
 ####################
-- name: Create Java 8 Alternative
-  alternatives:
-    name: java
-    link: /usr/bin/java
-    path: /usr/lib/jvm/java-1.8.0-openjdk/bin/java
-    priority: 1
-  when:
-    - ansible_architecture == "x86_64"
-    - ansible_distribution_major_version != "6"
+
+- name: Find Default JRE
+  stat:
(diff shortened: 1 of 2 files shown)


gdams:
@aahlenst failure on centos 6:

```
TASK [Common : Set Default JRE] ************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Needed to install the alternative, but unable to do so as we are missing the link"}
	to retry, use: --limit @/ansible/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.retry
```

aahlenst:
Should be fixed. Travis is having a bad time. Jobs cannot be started and it happens for other PRs, too.

Willsparker:
I'm able to test CentOS6/7/8 on x64 here: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/864
(The link won't work until some prior runs have finished, just FYI)

aahlenst:
@Willsparker Thanks for triggering that job. Seems to have successfully completed.

Willsparker (approved):
I'm unable to test the PR on any other platforms except x64, but they all worked, so approving :-)

Issue #1526, opened 2020-08-28
winPB: Stop MSVS_2019 erroring on installation when returning error code 3010


Willsparker:
Ref: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/859/console , https://adoptopenjdk.slack.com/archives/C53GHCXL4/p1597937660002600

Currently in the Windows PB, MSVS_2019 is returning error code 3010. As spotted by @jdekonin , this means the installation was successful but a reboot is required (https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2019). This can be worked around by forcing the task to NOT fail on error code 3010 ( https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#controlling-what-defines-failure ). As a task to reboot the machine is nearly directly after the installation, this should be fine.

PR #671, opened 2019-01-11
Ansible: Fix and change MSVS 2017 install flags

HusainYusufali:
* Changed install flag syntax for Visual Studio 2017
* Install package for [Desktop development with C++](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=vs-2017#desktop-development-with-c) that installs DIA SDK

* During testing after successfully installing VS 2017 the playbook would error out with return code 1.

```
fatal: [test-win-12rt31.fyre.ibm.com]: FAILED! => 
{"changed": true, "cmd": "C:\\temp\\vs_community.exe --wait --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended;includeOptional --quiet --norestart", 
"delta": "0:50:05.466202", "end": "2019-01-09 08:59:17.644974", "msg": "non-zero return code", "rc": 1, 
"start": "2019-01-09 08:09:12.178771", "stderr": "", "stderr_lines": [], "stdout": "
Preparing: C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\6c0d9a2f5596d39f21e5f6e999\\vs_bootstrapper_d15\\HelpFile\\2052\\help.html...\r\r\n
Preparing: C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\6c0d9a2f5596d39f21e5f6e999\\vs_bootstrapper_d15\\HelpFile\\1028\\help.html...\r\r\n
Preparing: C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\6c0d9a2f5596d39f21e5f6e999\\vs_bootstrapper_d15\\HelpFile\\1029\\help.html...\r\r\n
Preparing: C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\6c0d9a2f5596d39f21e5f6e999\\vs_bootstrapper_d15\\HelpFile\\1031\\help.html...\r\r\n
Preparing: C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\6c0d9a2…
```

* I tried various methods using other ansible modules like `win_shell` and `raw` which would hang during the installation

* Added an error exception for the installation so that return code 1 is ignored

Signed-off-by: HusainYusufali <husainyusufali7@gmail.com>

PR #1529, opened 2020-09-02
winPB: Stop MSVS_2019 installation failing on rc=3010

Willsparker:
fixes: #1526 

Ensure that the MSVS_2019 task won't fail when it returns error code 3010 as that states that the installation was successful but the machine needs rebooting to fully install. As a reboot task is scheduled afterwards, this should be fine.
I've also removed the exception for error code 1, as according to https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2019 , `1` does indeed mean that an error has occured.

Looking at MSVS_2017, it appears that the [documentation](https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2017#error-codes) says 1 is an error code too, but according to #671 , it should be ignored as well. For the sake of testing, I'll run this through `VPC`, to determine if a working installation of `MSVS_2019` can still report an error code of 1. If it is, I'll put `1` back in.

VPC Run: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/863/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2019/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2019/tasks/main.yml
@@ -23,7 +23,7 @@
     executable: cmd
   when: (not vs2019_installed.stat.exists)
   register: vs2019_error
-  failed_when: vs2019_error.rc != 1 and vs2019_error.rc != 0
+  failed_when: vs2019_error.rc != 0 and vs2019_error.rc != 3010
   tags: MSVS_2019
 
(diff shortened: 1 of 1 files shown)


sxa:
Should this still be in draft @Willsparker ?

Willsparker:
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/865/OS=Win2012,label=vagrant/console
Rerunning VPC - the last one didn't run properly unfortunately :/

Willsparker:
Latest VPC run went through okay, ready for review :+1:

sxa (approved):

PR #1530, opened 2020-09-02
swap travis with github actions

gdams:
The Travis builds have proven to be incredible flaky so I've swapped to using GH Actions and Docker Buildx to emulate the image creation using Qemu


aahlenst:
According to @gdams on Slack, Travis CI fixed their problems which makes this change unnecessary (at least for the time being).

PR #1532, opened 2020-09-07
Vagrant: Add warning in docs about starting Windows VM with too little memory

Willsparker:
This is a problem I came across when setting up a new laptop. I don't think we should reduce the memory of the Windows Vagrantfile, otherwise it can significantly increase the time taken to execute `VPC.sh` on it- but a warning should probably be put in to the docs, in case the user doesn't have enough available memory and is wondering why their VM is going to the aborted state straight away.

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -276,6 +276,8 @@ ln -sf Vagrantfile.Win2012 Vagrantfile
 
 vagrant up
 ```
+Note: If the machine running this only has 8G of memory, the windows VM may go into an aborted state on bootup. This is due to assigning the VM too much memory.
+To stop this, halve the memory to the VM by changing the following line in the Windows vagrantfile: `v.memory = 5120`
 
 Several files will also need to be edited for Windows:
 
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM

sxa (commented):
Can we trap this situation in the script (perhaps when a connection error occurs) and say something like "If your machine has 8GB or less of RAM, try editing this line in Vagrantfile.Win2012"

If it's possible to detect that situation (e.g. from a log file, or even just "testing windows and host has <=8Gb RAM) then perhaps we can improve the message further.

Willsparker:
>  If it's possible to detect that situation (e.g. from a log file, or even just "testing windows and host has <=8Gb RAM) then perhaps we can improve the message further.

I do quite like that idea. It can be put into the `checkVars` function in _vagrantPlaybookCheck.sh_ . However, in terms of warning people if they're just using the Vagrantfile, not the script, I don't think there's any way to warn them except in our docs - I tried using `config.vm.post_up_message` ( https://www.vagrantup.com/docs/vagrantfile/machine_settings#config-vm-post_up_message ), however that doesn't print a message if `vagrant up` fails :/

sxa (approved):

Willsparker:
I know you both have approved beforehand, but I didn't want to add a commit after approval and then merge right away

sxa:
Since it's just that extra comment that's not a problem - we can merge :-)

Willsparker:
Cool, thanks :-)

PR #1534, opened 2020-09-08
pbTests: Add missing '-' char so aarch64 QPC uses correct boot jdk

Willsparker:
The aarch64 QPC runs keep failing the build portion as it keeps trying to use jdk8 as the boot dir. This is due to `JDK_BOOT_DIR` being set to `/usr/lib/jvm/jdk10`, which doesn't exist, so it defaults to setting the boot dir to be what `JAVA_HOME` is, which is jdk8.


karianna:
@sxa Safe to merge this prior to release?

Haroon-Khel (approved):
LGTM

Haroon-Khel:
@karianna its a change to a script which is only used by jenkins jobs that test the infra's playbooks

PR #1535, opened 2020-09-08
unixPb: Add gcc_7 role to arm32

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1528

related https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1533

Added a condition to the `gcc_7` role so that it runs on Debian for arm32

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Debian.yml
@@ -106,6 +106,12 @@
 ##########################
 # Additional build tools #
 ##########################
+- name: Call gcc-7 for all but arm32
+  package: "name={{ item }} state=latest"
+  with_items: "{{ gcc_7 }}"
+  when: ansible_architecture != "armv7l"
+  tags: build_tools
+
 - name: Call Build Packages and Tools Task for OpenJFX
   package: "name={{ item }} state=latest"
   with_items: "{{ OpenJFX_Build_Tool_Packages }}"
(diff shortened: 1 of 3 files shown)


Haroon-Khel:
Running a qemu job on all but arm32 at https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/93/
(Already passed on arm32 at https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/92/ARCHITECTURE=arm32,label=vagrant/console)

Haroon-Khel:
@Willsparker Could you run a vpc job to ensure that this doesnt break the x64 playbooks, thanks.

Willsparker:
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/871/ 
Running on all OSs except Win2012. I can't see it being an issue, but better safe than sorry :-)

sxa (commented):
Before approving is it definitely the case that the same version of Debian on x64 and arm32 behave differently i.e. x64 has gcc-7 and arm32 doesn't? Which version does arm32 have. Seems odd that they wouldn't have the same one in the same Debian release.

Willsparker:
> Before approving is it definitely the case that the same version of Debian on x64 and arm32 behave differently i.e. x64 has gcc-7 and arm32 doesn't? Which version does arm32 have. Seems odd that they wouldn't have the same one in the same Debian release.

Looking at the succeeding QPC and VPC runs, they're both on Debian 8. x64 is able to install via `apt` , arm32 can't but installs it via the `gcc_7` role. They both end up with `gcc-7.5` that appears to work for the builds. 
On the arm32 VM, if you `apt install gcc` , it installs `gcc-4.9`. `apt-cache search gcc*` shows there's no `gcc-7`, just `4.9` and `4.8` available to it.

Willsparker (approved):
Approving as the [VPC run](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/871/) didn't break :-) I'll wait for @sxa 's approval before merging though

sxa (approved):
Seems odd, but in light of the analysis confirming that the gcc-7 package does not exist in Debian's arm32 repositories I'll let this in

PR #1536, opened 2020-09-08
aixPB: Missing tags, install sudo, path to variables file

Haroon-Khel:
I noticed some missing features on the aix playbook when using it earlier. 

The variables file would not load until I added the `./`. 

The task `name: Grant zeus sudo powers` requires the `/etc/sudoers` file to exist which only does if `sudo` is installed.

And some missing tags.


sxa (approved):
LGTM

Issue #452, opened 2018-09-03
Abstract out a lot of the distro specific checks from the Common directory


sxa:

Willsparker:
@sxa Do you mean just removing the lines such as 
https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/b107e720cbf20e860c9a5f1e07db582a2e25215f/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml#L81
or something else?

sxa:
I think if it's needed for some reason it's ok in the `RedHat.yml` for example, but if there are any in the main part of the common role then they shold probably be moved to distro specific files if possible

Willsparker:
A good example of why descriptions are necessary then ;-) 
Okay, thanks for clarification :+1:

Willsparker:
I've had a look, it looks like the only thing that _could_ be put into distributions would be the macOS hostname stuff, however, I sort of assume that this is required for the `update /etc/hosts` tasks below it.
I am going to remove the unnecessary distro checks that are already in the distro specific files however, just to cleanup.

PR #1540, opened 2020-09-11
UnixPB: Remove unnecessary distro checks in distro specific common files

Willsparker:
Fixes: #452 

From what I can see, the linked issue is pretty much done, however this PR removes some of the unnecessary distribution checks that are already in the distribution specific common files.


sxa (approved):
LGTM

PR #1545, opened 2020-09-16
winPB: Remove MSVS_2017 checksum

Willsparker:
Ref: #1517 , https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/876/console

Again, this is to reduce the PRs to update the checksum every so often. To ensure we don't forget about finding a better solution to this (i.e. caching an installer so the checksum doesn't change), I'll raise an issue and link this PR to it.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
@@ -14,8 +14,6 @@
     url: 'https://aka.ms/vs/15/release/vs_community.exe'
     dest: 'C:\TEMP\vs_community.exe'
     force: no
-    checksum: 29aeed55de6158a0968f8cb17944b733f65b0300cd10ffb2a53011f29159f1c1
-    checksum_algorithm: sha256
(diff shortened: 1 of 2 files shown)


karianna:
I'd prefer landing this after that other solution is put in place.  Also, I'd leave a comment saying why we're not checksumming here

Willsparker:
@karianna Comments have been put in. However, if we wait until we have cached an installer, this PR won't be necessary as the installer won't be updated in our cache, so the checksum would remain the same and we can put it back on

karianna:
Heh - @sxa thoughts on this?  I'm wary of not checking our installer files but I get the pain point here

Willsparker:
Okay, I've put the both `vs_community17.exe` and `vs_community19.exe`, in `/home/will/windows_installers` on the `ci.adoptopenjdk.net` machine. If someone could put them into `https://ci.adoptopenjdk.net/userContent/winansible` for me, I can change this PR into getting the installers from there :-)

Willsparker:
I'm going to close this PR, as I don't think it's going to be too much work to get the windows installers from our own cache, and this PR was just meant to be a stop gap. AFAIK, we can install MSVS_201X with slightly older installers too, so it should get be a case of putting them in the right place, and changing where the playbooks retrieve them.

PR #1552, opened 2020-09-17
Simplify AIX filesystem config script

aixtools:
Simplify AIX filesystem config - assuming 40G as total storage available.

Note: script also checks and verifies `maxuproc`. This may be moved later - so FILESYSTEM config does just one thing.

### DIFF
--- a/ansible/playbooks/scripts/AIX_filesystem_config.sh
+++ b/ansible/playbooks/scripts/AIX_filesystem_config.sh
@@ -1,42 +1,40 @@
-#!/bin/bash
+#!/usr/bin/ksh -e
 ##############################################################
 # AdoptOpenJDK - Script to configure filesystem sizes on AIX #
+# Stop script on error (-e option)
 ##############################################################
 #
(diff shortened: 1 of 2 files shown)


ansible/playbooks/scripts/AIX_filesystem_config.sh
```diff
@@ -72,239 +67,50 @@ if [[ $Maximum_Processes -le "512" ]]; then
...
-        if [[ $unall_disk_in_GB -gt "100" ]]; then
-                if [[ $cur_home -lt "100" ]]; then
-                        add_home=`expr 100 - $cur_home`
-                        chfs -a size=+$add_home"G" /home
-                fi
-        else
-                chfs -a size=+$unall_disk_in_GB"G" /home
-        fi
-fi
+[[ $cur_root -ne "2" ]] && chfs -a size=2G /
```
sxa, line 92: Why are these using `-ne` rather than `-ge` ?

PR #1561, opened 2020-09-24
ansible: add Packages (installer generator) to macOS playbook

gdams:
We are switching the installers over to use [Packages](http://s.sudre.free.fr/Software/Packages/about.html) 

See https://github.com/AdoptOpenJDK/openjdk-installer/pull/248


aahlenst (approved):

Willsparker (approved):
LGTM

PR #1571, opened 2020-09-28
ansible: add macos codesign cert role

gdams:
Adds the missing steps to install the macos codesign certs

Also bumps the centos6 build image to use python 3 because the retart module requires a more recent version of ansible

### DIFF
--- a/ansible/Dockerfile.CentOS6
+++ b/ansible/Dockerfile.CentOS6
@@ -2,9 +2,17 @@ FROM centos:6.9
 
 ARG user=jenkins
 
-RUN yum -y update; yum clean all
-RUN rpm -ivh http://dl.fedoraproject.org/pub/epel/6/i386/epel-release-6-8.noarch.rpm; \
- yum -y install ansible sudo; yum clean all
+# Install Python 3
+RUN yum -y update; yum clean all; \
+    yum -y install gcc openssl-devel bzip2-devel sqlite-devel wget; \
+    cd /usr/src; \
+    wget -q https://www.python.org/ftp/python/3.6.10/Python-3.6.10.tgz; \
+    tar xzf Python-3.6.10.tgz; \
+    cd Python-3.6.10; \
+    ./configure --enable-optimizations; \
+    make install; \
+    rm /usr/src/Python-3.6.10.tgz; \
+    pip3 install ansible
 
 COPY . /ansible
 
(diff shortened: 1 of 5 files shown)


Willsparker (approved):
Approving on the basis that the playbooks are written right - But I'm trusting @gdams that the playbooks do what they say they do

ansible/Dockerfile.CentOS6
```diff
@@ -2,20 +2,28 @@ FROM centos:6.9
...
 ARG user=jenkins
 
-RUN yum -y update; yum clean all
-RUN rpm -ivh http://dl.fedoraproject.org/pub/epel/6/i386/epel-release-6-8.noarch.rpm; \
- yum -y install ansible sudo; yum clean all
+# Install Python 3
+RUN yum -y update; yum clean all; \
+    yum -y install gcc openssl-devel bzip2-devel sqlite-devel sudo wget; \
+    cd /usr/src; \
+    wget -q https://www.python.org/ftp/python/3.6.10/Python-3.6.10.tgz; \
```
karianna, line 9: seems a very specific and slightly older version of Python?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/macos_codesign/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+    reboot_timeout: 3600
+  when:
+    - not macos_version | regex_search("10.10")
+    - default_keychain.stdout != "/Users/jenkins/Library/Keychains/login.keychain-db"
+
+- name: Test Application Certificate
+  shell: |
+    security unlock-keychain -p `cat ~/.password` login.keychain-db
+    rm -rf /Users/jenkins/test && touch /Users/jenkins/test
+    codesign --sign "Developer ID Application: London Jamocha Community CIC" /Users/jenkins/test
```
karianna, line 52: I wonder if we need to parameterize this for folks outside Adopt build farm using these scripts

karianna (changes requested):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/macos_codesign/tasks/main.yml
gdams: We have the AdoptOpenJDK tag here so no need.

ansible/Dockerfile.CentOS6
gdams: This is the newest version I could get to compile on centos6

karianna (approved):

Issue #1546, opened 2020-09-16
winPB: Find a better way to manage MSVS checksums


Willsparker:
Ref: #1517  , #1545 

With both MSVS_2017 & 2019, we've removed the checksums as the installers are often updated, requiring a lot of 'checksum update' PRs, that can block `VPC` runs until the PRs are merged.
We really should have checksum checks whenever downloading anything from the internet, so a better solution needs to be found. 
At https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1517#pullrequestreview-475402407 , @sxa suggested caching the installers somewhere so they aren't updated and the checksum stays the same.

Willsparker:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1572#issuecomment-704430037 

@sxa Would something such as
```
if credentials are present
  Retrieve MSVS_201* from the secrets repo / non public place
else
  Retrieve installer from regular place w/o checksum
```
That way other (non-adoptopenjdk) people can still use the playbooks, but when setting up machines, we can ensure the installations okay?

sxa:
I'd say it's not ideal but better than not having a fallback.

PR #1572, opened 2020-09-28
winPB: Retrieve MSVS_2017/19 installers from ci.adoptopenjdk.net

Willsparker:
Ref: #1545 , https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1517#pullrequestreview-475402407
Fixes: #1546 

Retrieve the MSVS installers from `ci.adoptopenjdk.net/userContent/winansible/MSVS_installers/`. This is to stop the installer updates that change the checksum (thereby meaning I have to PR them to change them every time). AFAIK, the latest updates of MSVS can still be installed with old versions of the installer, so this shouldn't affect the installation of MSVS itself (and therefore won't affect builds). This way we're also able to keep the checksum checks :-)


karianna (approved):
LGTM

Willsparker:
Oh, I forgot to mention, I put this through `VPC`, and the playbook seems good.
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/880/OS=Win2012,label=vagrant/

I'm not quite sure why the build failed tho - I'm rerunning just to double check. https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/881/

sxa (commented):
Typo in "retrieve"

gdams:
Are there any legal issues with hosting the visual studio installers on our own web server?

Willsparker:
Files have been removed from the public server, so this PRs been put into draft until we can find a secure place to put them :-)

aahlenst:
One possible solution is to put those into Artifactory into a private repo that can only be accessed by authenticated users. Works great with Ansible.

gdams:
@aahlenst we want to avoid that because it would mean that other people would not be able to use our playbooks anymore.

karianna:
merge conflicts.

Willsparker:
Conflicts resolved - It was due to the workaround PR that removes the checksum.
From what I can see, if we can't legally redistribute the installer, we're going to just deal with the often changing checksum, or remove it, if we want to keep the playbooks working for everyone.

karianna:
> Conflicts resolved - It was due to the workaround PR that removes the checksum.
> From what I can see, if we can't legally redistribute the installer, we're going to just deal with the often changing checksum, or remove it, if we want to keep the playbooks working for everyone.

Opportunity to write a bot or GitHub Action to create PR like dependabot?

sxa:
Closing as we are not going to redistribute this way

PR #1585, opened 2020-10-02
unixPB: add solaris x64 support

gdams:
The main pain with this is that the get_uri module doesn't work on Solaris so we have to use wget


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -36,21 +36,25 @@
...
     - NTP_TIME
     - gcc_48
     - role: gcc_7                 # OpenJ9
       tags: [build_tools, build_tools_openj9]
     - role: cmake                 # OpenJ9 / OpenJFX
+      when: ansible_distribution != "Solaris" # Compile fails on Solaris
       tags: [build_tools, build_tools_openj9, build_tools_openjfx]
-    - ccache
+    - role: ccache
+      when: ansible_distribution != "Solaris" # Not needed on Solaris
```
sxa, line 57: We say "Not needed on Solaris" yet we still seem to have it in the common vars. Is this correct? I'd normally expect the ccache rol not to go through if it detects another version installed, so this guard seems irrelevant

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -61,19 +65,19 @@
       tags: build_tools
     - role: adoptopenjdk_install  # JDK11 Build Bootstrap
       jdk_version: 10
-      when: ansible_distribution != "Alpine"
+      when: ansible_distribution != "Alpine" and ansible_distribution != "Solaris"
```
sxa, line 68: Gut feel is still that these guards would be better inside the role, so that all the logic relating to the available versions is in one place, but I wouldn't block the PR on that basis at the moment.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -0,0 +1,161 @@
...
+  shell: ./SolarisStudio12.3-solaris-x86-pkg/solarisstudio.sh --non-interactive
+  args:
+    chdir: /tmp
+  tags:
+    - build_tools
+    - solaris_studio
+    - adoptopenjdk
+  when: not solaris_studio.stat.exists
+
+- name: Check if cups is installed
```
sxa, line 85: Is cups not available through pkgutil?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -0,0 +1,161 @@
...
+  file:
+    src: /opt/csw/bin/g{{ item }}
+    dest: /opt/csw/bin/{{ item }}
+    owner: root
+    group: root
+    state: link
+  with_items:
+    - tar
+    - make
+    - sha256sum
```
sxa, line 132: Do we really need this for `sha256sum`? That would look rather odd

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -0,0 +1,161 @@
...
+  command: "{{ item }}"
+  with_items:
+    - svcadm disable sendmail
+    - svcadm disable sendmail-client
+  tags: adoptopenjdk
+
+- name: Modify PATH
+  lineinfile:
+    path: /etc/profile
+    line: PATH=/opt/csw/bin/:$PATH
```
sxa, line 150: What fails when this is not at the start of the PATH - it's generally preferable to have non-system-default tools at the end

(I've got a feeling there was a reason for this in the past but we need to document why in here since it's far from ideal)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
```diff
@@ -41,7 +58,7 @@
     - adoptopenjdk
     # TODO: write a condition when NOT to run this step
     - skip_ansible_lint
-  when: (ansible_distribution != "MacOSX")
+  when: ansible_distribution != "MacOSX" and ansible_distribution != "Solaris"
```
sxa, line 61: These conditionals are looking a bit odd. The comment says "for Redhat" but there seems nothing to guard that. If it's linux-specific perhaps we should be explicit about that instead of saying "Not blah" when something new shows up in the playbooks

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
```diff
@@ -19,6 +20,23 @@
     - Superuser_Account == "Enabled"
   tags: [superuser, adoptopenjdk]
 
+- name: Create Superuser user's home folder (Solaris)
```
sxa, line 23: Somewhat disappointing if this isn't the default

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
```diff
@@ -92,6 +101,25 @@
   when:
     - adoptopenjdk_installed.rc != 0
     - ansible_os_family != "Darwin"
+    - ansible_os_family != "Solaris"
+  tags: adoptopenjdk_install
+
+- name: Install latest release if one not already installed (Solaris)
+  command: wget https://api.adoptopenjdk.net/v3/binary/latest/{{ jdk_version }}/ga/solaris/{{ api_architecture }}/jdk/{{ bootjdk }}/normal/adoptopenjdk?project=jdk -O /tmp/jdk-{{ jdk_version }}.tar.gz
```
sxa, line 108: I take it this is because unarchive isn't working due to a cert error too?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/freemarker/tasks/main.yml
```diff
@@ -21,19 +21,34 @@
     dest: /tmp/
     remote_src: yes
     mode: 0755
-  when: not freemarker.stat.exists
+  when: not freemarker.stat.exists and ansible_distribution != "Solaris"
+  tags: [freemarker, adoptopenjdk]
+
+- name: Download and extract freemarker.jar (Solaris)
```
sxa, line 27: Why have we gone for `unarchive` here with `validate_certs: false` when we've used `wget` elsewhere?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/maven/tasks/main.yml
```diff
@@ -16,7 +16,16 @@
    url: https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
    dest: /tmp/
    checksum: sha512:c35a1803a6e70a126e80b2b3ae33eed961f83ed74d18fcd16909b2d44d7dada3203f1ffe726c17ef8dcca2dcaa9fca676987befeadc9b9f759967a8cb77181c0
-  when: maven_installed.rc != 0
+  when: maven_installed.rc != 0 and ansible_distribution != "Solaris"
+  tags: maven
+
+- name: Download Apache Maven v3.6.3 (Solaris)
+  get_url:
+   url: https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
```
sxa, line 24: Same comment as freemarker ... why use a differnet method for these when using `wget` elsewhere?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/maven/tasks/main.yml
```diff
@@ -16,7 +16,16 @@
...
+   dest: /tmp/
+   checksum: sha512:c35a1803a6e70a126e80b2b3ae33eed961f83ed74d18fcd16909b2d44d7dada3203f1ffe726c17ef8dcca2dcaa9fca676987befeadc9b9f759967a8cb77181c0
+   validate_certs: False
```
sxa, line 27: As per previous comments

ansible/plugins/inventory/adoptopenjdk_yaml.py
```diff
@@ -46,7 +46,8 @@
   # providers - validated for consistency
   'provider': ('azure', 'marist', 'osuosl', 'scaleway',
         'macstadium', 'macincloud', 'softlayer', 'spearhead',
-        'packet', 'linaro','digitalocean', 'ibm', 'godaddy', 'aws')
+        'packet', 'linaro','digitalocean', 'ibm', 'godaddy',
+        'aws', 'esxi')
```
sxa, line 50: Should we make this `packet-esxi` to make it clearer what the provider is? (Arguably doesn't need the `-esxi` but I would definitely be happier if it said packet somewhere)

sxa (changes requested):
Sorry - quite a few comments :-) A few will just be questions but there are a few inconsistencies too

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/maven/tasks/main.yml
gdams: I've used get_url wherever possible. It seems to be down to the TLS version that each server accepts

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/freemarker/tasks/main.yml
gdams: I've used get_url wherever possible. It seems to be down to the TLS version that each server accepts

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
```diff
@@ -48,6 +49,16 @@
   when: Domain == "adoptopenjdk.net" and ansible_distribution == "MacOSX"
   tags: hostname,adoptopenjdk
 
+- name: Set hostname to inventory_hostname (Solaris)
+  shell: "{{ item }}"
+  with_items:
+    - echo "{{ inventory_hostname }}" > /etc/nodename
+    - echo "{{ inventory_hostname }}" > /etc/hostname.e1000g0
```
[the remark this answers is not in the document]
gdams: my understanding is that on nearly all Solaris systems the default is `e1000g0`. I'm not very happy with this approach for setting the hostname in general - it seems very hacky

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Solaris.yml
```diff
@@ -0,0 +1,29 @@
...
+###########
+
+# Command Build Tool Packages
+Build_Tool_Packages:
+  - autoconf
+  - binutils
+  - coreutils
+  - ccache
+  - jdk7
+  - jdk8
```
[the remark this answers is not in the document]
gdams: Not technically needed but our other build machine used this jdk8 to connect to jenkins and the adopt one for buildJDK only. I was simply mirroring the setup

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
gdams: updated, the check only looks in `/usr/local/bin/ccache` which isn't where it's located on Solaris. I've updated the text.

ansible/plugins/inventory/adoptopenjdk_yaml.py
gdams: updated

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: yup it's a required symlink. As part of the build scripts we run sha256sum to generate the checkum

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: OpenJDK expects all the gnutools to be present on Solaris. As part of the build scripts we currently `export PATH=/opt/csw/bin/:$PATH`

I was simply modifying the profile so that when people SSH in the environment is the same

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
gdams: unfortunately, the built-in ansible user module doesn't create the home folder so this was a workaround

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
gdams: I'm currently working on the macOS playbook which will use this role. I'll make those changes over there

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml
gdams: Agreed, I didn't want to break anything so the simplest solution was to skip Solaris. We should clarify if this is Rhel or Linux specific

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
```diff
@@ -14,8 +14,6 @@
     url: 'https://aka.ms/vs/15/release/vs_community.exe'
     dest: 'C:\TEMP\vs_community.exe'
     force: no
-    checksum: 29aeed55de6158a0968f8cb17944b733f65b0300cd10ffb2a53011f29159f1c1
```
sxa, line 17: Erm, this is nothing to do with Solaris so should not form part of this PR

sxa, line 17: This is not related to Solaris so should not be in here

sxa (changes requested):
Commit message should read `unixPB` at start for consistency instead of `ansible:` but this can be fixed on squashing. Otherwise happy to approve once the non-Solaris bit that's wormed it's way into this PR is removed ;-)

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
gdams: rebase sorted this out. looks like a git glitch

gdams:
@sxa updated PTAL

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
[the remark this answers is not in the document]
sxa: Certainly on Solaris 10 I've had interface names of `bge0` although I'd suggest adding a comment to this saying that it may not work if the interface name is not `e1000g0`, or even `TODO: Allow other interfaces` above this section

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
sxa: I know `sha256sum` is used but what tries to invoke it as `gsha256sum` which seems to be what this symlink operation is creating unless I'm missing something?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
sxa: I really don't think that's a good idea. We can mask problems that users will potentially see if we change the OS default tools. We've already hit issues on AIX because of this sort of change and I don't want to risk testing being invalid as a result of the tools not being the standard solaris ones

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/freemarker/tasks/main.yml
sxa: Hmmm is it TLS rather than the CA certs? (the latter should be replaceable)

sxa:
Comments added. Let's try and sort this out tomorrow if possible.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: Other way round, it symlinks gsha256sum to sha256sum. Also gmake to make etc

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -0,0 +1,161 @@
...
+  ignore_errors: true
+  register: pkgutil_installed
+  tags:
+    - build_tools
+
+- name: Install pkgutil
+  shell: |
+    set -o pipefail
+    echo action=nocheck > responses.txt
+    yes all | pkgadd -d http://get.opencsw.org/now -a responses.txt
```
sxa, line 22: `http` on here makes me very nervous although I appreciate that it's due to the same reason get_uri doesn't work.
What is `action=nocheck` required for as it seems to install ok without it?
I'd change it to this to explicitly install `CSWpkgutil` from the bundle
```suggestion
    yes | pkgadd -d http://get.opencsw.org/now CSWpkgutil
```

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -0,0 +1,161 @@
...
+    /opt/csw/bin/pkgutil -U
+    /opt/csw/bin/pkgutil -y -i pkgutil
+    /usr/sbin/pkgchk -L CSWpkgutil
```
sxa, line 25: What were these two lines for? Debugging? If we're downloading the latest in the initial `pkgadd` step this seems unnecessary

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -0,0 +1,161 @@
...
+    chdir: /tmp
+  when: not cups_installed.stat.exists
+  tags:
+    - build_tools
+    - cups
+
+- name: Compile cups 1.4.4
+  shell: |
+    cd /tmp/cups-1.4.4
+    ./configure --disable-gssapi --disable-ldap --prefix=/opt/sfw/cups
```
sxa, line 112: Since the version we want isn't in the sfw/csw repos would it not be preferable to put this into `/usr/local/cups` instead of other namespaces?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
```diff
@@ -4,6 +4,7 @@
 #############
 - name: Create Superuser account - zeus
   action: user name=zeus state=present shell=/bin/bash
```
sxa, line 6: I would strongly advise not setting this to `/bin/bash` as it's not guaranteed to be there on a basic Solaris installation

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
```diff
@@ -4,6 +4,7 @@
...
+  ignore_errors: yes
```
sxa, line 7: Why are we ignoring errors here?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
sxa: Eeeow - surprised it creates that one with a g prefix by default. I'm still not sure about the other two though - pretty sure we have stuff in our scripts that compensates for replacing tar and make with the g prefixed ones (search gtar and gmake in the build repo), and as per other comment I don't think it's right to switch the defaults on the system and potentially cause unecessary confusion

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Solaris.yml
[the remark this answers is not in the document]
sxa: OK - I'd rather "use what we sell" but I won't consider it a blocker to this PR :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
sxa: Thinking some more we should possible store our own copy of this in Vendor_Files (or other comparable location) if we can't update the CA certs that pkgadd is using

sxa: Hmmm in fact given that the CSWpkgutil package is only about 220K gzipped I could just about accept having it stored in this repository and then copied acrosss to the machine to be installed (something I would normally NEVER advocate, but it's very small)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
gdams: This was copied from the Jenkins user step. The create user role fails on Solaris and macOS because of home directory issues but does actually create the user so the skip step works around this.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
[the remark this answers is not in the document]
gdams: updated

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: This is the place that OpenJDK expects cups to be located. It's also where we put cups on the other solaris servers

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
sxa: What's the issue on macos? We could set a variable for the home dir location and pass it into this operation with `home=/export/home/zeus` or wherever else it is (defaulting to `/home/zeus`. We run the risk of playbooks completing without anyone realising that it hasn't done the operations if we ignore errors here.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: updated PTAL

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: removed

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Superuser/tasks/main.yml
gdams: updated

gdams:
@sxa updated PTAL

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
sxa: @gdams Can you remove this part - I don't want that as a default on the machine in case it has other side effects or masks other problems. Happy to approve & merge this if you remove it.

gdams:
@sxa updated PTAL

sxa (approved):
Thanks for the changes - happy for this to go in now but we need to keep an eye on the new workflows to ensure they don't cause any problems.

PR #1596, opened 2020-10-06
winPB: Install Cygwin without a cache

Willsparker:
@gdams found a better way to install Cygwin on the OpenJDK-build github actions, so I thought it'd be best to do the same here: 
https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/883/
This ends up installing in 2 minutes as opposed to the ~20-25 minutes to install it before.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
@@ -15,35 +15,21 @@
   register: cygwin_installed
   tags: cygwin
 
-# For details about the cache, see: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1210#issuecomment-679076545
-- name: Retrieve cygwin cache from ci.adoptopenjdk.net
+- name: Retrieve Cygwin setup
   win_get_url:
-    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64.tar.gz
-    dest: C:/temp/cygwin64.tar.gz
-    checksum: 8f8b1d7d00233e384e725390c4018941fe7125290b94ecc0eeb3cf6684b4ce23
+    url: https://cygwin.com/setup-x86_64.exe
+    dest: C:\temp\cygwin.exe
+    checksum: 4dd4d4531e8e63ade849daaaf587ba1c1430368701772c8ee42a27f4e8c373e4
     checksum_algorithm: sha256
-  when: (not cygwin_installed.stat.exists)
+  when: not cygwin_installed.stat.exists
   tags: cygwin
 
-- name: Retrieve cygwin toolset from ci.adoptopenjdk.net
-  win_get_url:
-    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64_bootstrap.zip
-    dest: C:/temp/cyg_toolset.zip
-    checksum: d7950df964a0618ac47561d349276d4644e27f5b8f0a5fba9b10db702d09ecad
-    checksum_algorithm: sha256
-  when: (not cygwin_installed.stat.exists)
-  tags: cygwin
-
-- name: Use cygwin 'tar' and 'pigz' to extract cygwin cache
+- name: Install Cygwin
   win_shell: |
-    New-Item -ItemType directory -Name cygwin_toolset -Path C:\temp\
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -15,35 +15,21 @@
...
-# For details about the cache, see: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1210#issuecomment-679076545
-- name: Retrieve cygwin cache from ci.adoptopenjdk.net
+- name: Retrieve Cygwin setup
   win_get_url:
-    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64.tar.gz
-    dest: C:/temp/cygwin64.tar.gz
-    checksum: 8f8b1d7d00233e384e725390c4018941fe7125290b94ecc0eeb3cf6684b4ce23
+    url: https://cygwin.com/setup-x86_64.exe
+    dest: C:\temp\cygwin.exe
+    checksum: 4dd4d4531e8e63ade849daaaf587ba1c1430368701772c8ee42a27f4e8c373e4
```
gdams, line 22: do we want to checksum this? it's a dynamic link that changes each time they bump the version so will require regular playbook updates

Willsparker: I was somewhat concerned about that - the alternative is to not have the checksum, or to do what we're looking at with the Visual Studio installers

gdams: I would remove it for now.

Willsparker: Fixed :+1:

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
```diff
@@ -15,35 +15,19 @@
...
-
-- name: Retrieve cygwin toolset from ci.adoptopenjdk.net
-  win_get_url:
-    url: https://ci.adoptopenjdk.net/userContent/winansible/cygwin64_bootstrap.zip
-    dest: C:/temp/cyg_toolset.zip
-    checksum: d7950df964a0618ac47561d349276d4644e27f5b8f0a5fba9b10db702d09ecad
-    checksum_algorithm: sha256
-  when: (not cygwin_installed.stat.exists)
+    url: https://cygwin.com/setup-x86_64.exe
+    dest: C:\temp\cygwin.exe
```
gdams, line 21: should we cleanup this file or does that happen somewhere else?

Willsparker:
Ref: Latest commit, I also removed the unnecessary `var` file :-)

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
Willsparker: We've historically always kept it on the machine, possibly to simplify manual updates? I don't think it necessarily needs to be on there though

gdams: We'll leave it for now. Perhaps another PR to cleanup any tmp files

gdams (approved):

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/cygwin/tasks/main.yml
Willsparker: I intend to have one of those fairly soon anyway :-)

Willsparker:
At the request of @sxa , I've compiled 2 lists of the `cygwin` packages that we install, both before this PR and afterwards. This means if we have something go wrong, we have a log of what's we've changed and what we're missing.

This was done in `mintty.exe` by running `cygcheck --check-setup | awk '{print $1 " " $2}' >> ***_cyglist.txt`

[old_cyglist.txt](https://github.com/AdoptOpenJDK/openjdk-infrastructure/files/5355490/old_cyglist.txt)
[new_cyglist.txt](https://github.com/AdoptOpenJDK/openjdk-infrastructure/files/5355515/new_cyglist.txt)

PR #1600, opened 2020-10-06
docker: fix docker login command windows

gdams:
fixes the failing docker login command on windows


Willsparker (approved):
Syntax looks fine, but sort of relying on you that this fixes the issue 👍

PR #1610, opened 2020-10-09
docker: windows add 8dot3 shortnames and jdk10 boot dir

gdams:


sxa (approved):
LGTM ... Although it would be better to use the API, particularly given that 10 isn't going to get any further updates so the version numnber/path name won't be changing

PR #1611, opened 2020-10-09
unixPB: Update cmake location (Downloads now on github)

sxa:
The cmake website seems to have moved where cmake is stored. This is causing playbook executions to fail as it is not at the expected location. This moves it to the new location on github. Checksum is the same so this should be a safe change.

Testing along with another (build) change at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/891


Willsparker (approved):
LGTM- based on the VPC run :-)

PR #1625, opened 2020-10-14
WinPB: Clean-up zip files

Willsparker:
Just somethings I noticed were left over in the `C:\temp` folder, after setting up the `IBMCloud` machines.
FYI, I'm not sure if the `\` character will cause an issue with the yaml files, but it's flagging it as red


karianna:
Linter not happy for some reason.

Willsparker:
Yeahh- it's due to Ansible not liking the `\` characters when they're used in double quotes. I've either removed them, or change the path to using `/` instead, where the double quotes are required.

VPC Run: https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/901/

karianna (approved):
LGTM

PR #1626, opened 2020-10-14
ansible: Create Windows JDK7 link as jdk-7 instead of jdk7

sxa:
I would like to obtain explicit approval from EVERYONE on the reviewers list. Can I ask no-one else other than me to merge please (which is the default policy for the openjdk-infrastructure repo anyway, but let's make sure no-one misses it for this one ;-)

The link in `C:\openjdk` for the Java 7 boot JDK is created as `jdk7` instead of `jdk-7`. This is inconsistent with all other versions. The current production build machines seem to have a mix, so I'm choosing to standardise on the same as the other versions.

To all reviewers: Please let me know if this will impact you and if so when you will have mitigated it (or if you are unable to do so - in which case we would still have the option to put both in place but I'd rather avoid that)

Attempting test via [VPC](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/900/console)

### DIFF
--- a/ansible/pbTestScripts/buildJDKWin.sh
+++ b/ansible/pbTestScripts/buildJDKWin.sh
@@ -67,7 +67,7 @@ checkJDK() {
 		exit 1
 	fi
 	if ((JAVA_TO_BUILD == 8)); then
-		export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk7
+		export JDK_BOOT_DIR=/cygdrive/c/openjdk/jdk-7
 		export JAVA_TO_BUILD="jdk8u"
 	elif ((JAVA_TO_BUILD > JDK_GA)); then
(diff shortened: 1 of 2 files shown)


AdamBrousseau (approved):
This is fine with me, thanks for the heads-up.

Willsparker (approved):
Yep, nothing against this on the condition it works in VPC. Reruns at https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/902/ & https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/903/

karianna (approved):
LGTM, no impact as far as I can see.

sxa:
Happy with the set of reviews I've had - merging as this matches what the two new machines we use at AdoptOpenJDK now have

PR #1640, opened 2020-10-21
actions: add macOS playbook tester

gdams:
linked to https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1638

Also replaces https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1317

### DIFF
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -1,4 +1,4 @@
-name: Ansible macOS
+name: Ansible Playbook
 
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


Willsparker:
Had a look at the macOS 10.15 check - it appears to be skipping the `adoptopenjdk_install` roles, despite the additions you've made. Also, as far as I can see, macOS isn't able to get a JDK7 for bootstrapping JDK8?

gdams:
@Willsparker agreed, we need a jdk7 bootjdk. We can probably create a separate PR for that. The adoptopenjdk_install role is skipped because AdoptOpenJDK is already preinstalled in the environment. I could try and uninstall them from the environment but not sure if it's worth the effort? 

https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#java

Willsparker (approved):
LGTM , in that case :-)

PR #1642, opened 2020-10-21
UnixPB: update docker-ce repo checksum for centos7

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1641

Tested on a centos 7 machine. I was able to reproduce the error, and updating the checksum allowed the docker role to run without error


Willsparker (approved):
LGTM- double checked the checksum :+1:

PR #1643, opened 2020-10-21
winPB: Replace MSVS_2017/9 installer URLs and reinstate checksums

Willsparker:
Fixes: #1546 

@gdams found some alternative links to retrieve the MSVS installers, that should be persistent (i.e. not have their checksum change every X weeks). This means we're able to reinstate the checksums again :-)

VPC Run: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/908/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2017/tasks/main.yml
@@ -11,7 +11,9 @@
 
 - name: Download Visual Studio Community 2017
   win_get_url:
-    url: 'https://aka.ms/vs/15/release/vs_community.exe'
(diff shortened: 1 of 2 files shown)


gdams (approved):
LGTM subject to vagrant check passing

sxa (changes requested):
For clarity in the future and to allow people to update it easily in the future, can you add a comment indicating that this is the target that you are redirected to when you go to https://aka.ms/vs/15/release/vs_community.exe (Same for both changes)

Willsparker:
VPC run seemed to go okay- It was able to run the playbook and the build fine; the reason it's reporting unstable is because it looks like the Python script timed-out during the testing phase.

sxa (approved):
If we refresh the machines at this level it would also be a fix for https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1270

Willsparker:
Good point - I'll need to find a non-critical time to do that :-)

PR #1694, opened 2020-11-19
TRSS: add forever service creation to TRSS playbook

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1679

Ive added the steps to create the forever services for the TRSS frontend and backend to the trss playbook as per https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1679#issuecomment-729766341

Will test on a ubuntu 18 vm shortly

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
@@ -6,6 +6,9 @@
 - hosts: all
   user: root
   become: yes
+  vars:
+    TRSS_working_dir: "/home/jenkins/openjdk-test-tools/TestResultSummaryService"
+    TRSS_config_path: "/data/db/credentials/trssConf.json"
   tasks:
     - block:
         - name: Load AdoptOpenJDKs variable file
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
@llxia 
(as I cant seem to add you as a reviewer)

sxa (approved):
Looks ok to me, but let's give it a day or so before merging in case anyone else objects :-D
Thanks.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/trss.yml
```diff
@@ -6,6 +6,9 @@
 - hosts: all
   user: root
   become: yes
+  vars:
+    TRSS_working_dir: "/home/jenkins/openjdk-test-tools/TestResultSummaryService"
+    TRSS_config_path: "/data/db/credentials/trssConf.json"
```
llxia, line 11: Do we create this file `/data/db/credentials/trssConf.json`? If not, we will get error when starting TRSS.

Haroon-Khel: I would recommend manually creating and populating the trssConf.json file before running the playbook. How do you feel about it?

sxa: I suppose we could create a template with a password of "password" ... Depends if there's a' risk of that making it into production - is that a password that's used during the creation of the database or will it just throw a (reasonably easy to debug) error message if the passwords don't match?

Haroon-Khel:
Runs fine on a ubuntu 18 vm 👍🏻, using a blank file for trssConf.json

llxia:
`trssConf.json` stores credentials for connecting to DB and CI servers. If we do not have DB password set, an empty `trssConf.json` is fine/correct. Personally, I would recommend adding the password into DB for security reasons at a later stage.

PR #1695, opened 2020-11-20
aixPB: Modify order of executition to have requirements first

aixtools:
##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -32,7 +32,7 @@
     # AIX BOS configuration
     - syslogs
     # TBD: additional tasks below that need to be promoted to
-    # or migrated into an AIX setup role - in paritcular - 
+    # or migrated into an AIX setup role - in paritcular -
     # the tasks that setup the legal shells needed for user configuration
     # may need to be performed AFTER yum processing - rather than before
 
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -23,20 +23,22 @@
     ## Additional Software
     ## OJDK Bootstraps
 
-    # AIX BOS install requirements
+    # 1. AIX BOS install requirements
+
     # These packages are installed using installp - which can enlarge
     # filesystem space, as needed, on demand
-    - X11
     - openssl
 
-    # AIX BOS configuration
+    # 2. AIX BOS configuration
+
     - syslogs
     # TBD: additional tasks below that need to be promoted to
     # or migrated into an AIX setup role - in paritcular -
     # the tasks that setup the legal shells needed for user configuration
     # may need to be performed AFTER yum processing - rather than before
 
-    # Core OSS installation
+    # 3. Core OSS installation
+
     ## Note: this may actually already be performed - partially
     ## depending on how python - to use Ansible - has been prepared
     # yum based installation of additional software
@@ -46,14 +48,22 @@
     - aixfs
     - yum
 
-    # AIX User Admin
+    # 4. AIX User Admin
+
     # users needed - wait until here to be sure bash is installed
     # tbd: Need other actions performed first - mainly adding /usr/bin/bash
     # to the list of legal shells.
     - jenkins_user
     - zeus_user
 
-    # Additional Software: both licensed and OSS
+    # 5. Additional Software: both licensed and OSS
+
+    # Note: although X11 is AIX BOS software - there is a chance that the
+    # Ansible module unarchive: may be called - and that requires,
+    # at a minimum, gtar and unzip to be available - regardless of need.
+    # Likewise, the xlc files may also need the Ansible unarchive and
+    # need to be evaluated after the CORE OSS software has been installed.
+    - X11
     ## verify/install licensed IBM compilers
     - xlc_v13
     - xlc_v16
@@ -62,7 +72,8 @@
     - ant
     - ant_contrib
 
-    # Install OJDK Bootstraps
+    # 6. Install OJDK Bootstraps
+
     - bootjdk7
     - bootjdk8
     - bootjdk9
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -15,32 +15,54 @@
     swap_size: 4096
 
   roles:
-    # AIX base requirements
+    # Roles are put into 6 groups
+    ## AIX BOS install requirements
+    ## AIX BOS configuration
+    ## Core OSS (bash, unzip, gtar)
+    ## AIX User Admin
+    ## Additional Software
+    ## OJDK Bootstraps
+
+    # AIX BOS install requirements
+    # These packages are installed using installp - which can enlarge
+    # filesystem space, as needed, on demand
     - X11
     - openssl
 
-    # AIX configuration
-    - aixfs
+    # AIX BOS configuration
     - syslogs
+    # TBD: additional tasks below that need to be promoted to
+    # or migrated into an AIX setup role - in paritcular - 
+    # the tasks that setup the legal shells needed for user configuration
+    # may need to be performed AFTER yum processing - rather than before
 
+    # Core OSS installation
+    ## Note: this may actually already be performed - partially
+    ## depending on how python - to use Ansible - has been prepared
     # yum based installation of additional software
     # installs some gnu packages required by ansible modules
+    # Note: AIX File system configuration must be run now as RPM based software
+    # cannot expand filesystem space on demand
+    - aixfs
     - yum
 
-    # verify/install IBM compilers
-    - xlc_v13
-    - xlc_v16
+    # AIX User Admin
     # users needed - wait until here to be sure bash is installed
     # tbd: Need other actions performed first - mainly adding /usr/bin/bash
     # to the list of legal shells.
     - jenkins_user
     - zeus_user
 
-    # additional OSS packages
+    # Additional Software: both licensed and OSS
+    ## verify/install licensed IBM compilers
+    - xlc_v13
+    - xlc_v16
+
+    ## additional OSS packages
     - ant
     - ant_contrib
 
-    # install/verify boot versions of ojdk
+    # Install OJDK Bootstraps
     - bootjdk7
     - bootjdk8
     - bootjdk9
@@ -49,6 +71,8 @@
     - bootjdk12
 
   tasks:
+  # TBD: as much as possible reorganize the tasks here into a appropriate role
+  # And place it in one of the logical areas described above under roles:
   ######################################
   # Add bash to available login shells #
   ######################################
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -18,17 +18,21 @@
     # AIX base requirements
     - X11
     - openssl
-    # verify/install IBM compilers
-    - xlc_v13
-    - xlc_v16
+
     # AIX configuration
     - aixfs
     - syslogs
 
     # yum based installation of additional software
+    # installs some gnu packages required by ansible modules
     - yum
 
+    # verify/install IBM compilers
+    - xlc_v13
+    - xlc_v16
     # users needed - wait until here to be sure bash is installed
+    # tbd: Need other actions performed first - mainly adding /usr/bin/bash
+    # to the list of legal shells.
     - jenkins_user
     - zeus_user
 
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -15,27 +15,34 @@
     swap_size: 4096
 
   roles:
-    - aixfs
-    - syslogs
+    # AIX base requirements
     - X11
+    - openssl
+    # verify/install IBM compilers
     - xlc_v13
     - xlc_v16
-    - openssl
+    # AIX configuration
+    - aixfs
+    - syslogs
 
+    # yum based installation of additional software
     - yum
 
+    # users needed - wait until here to be sure bash is installed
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
```diff
@@ -15,28 +15,35 @@
...
-    - syslogs
+    # AIX base requirements
     - X11
+    - openssl
+    # verify/install IBM compilers
     - xlc_v13
     - xlc_v16
-    - openssl
+    # AIX configuration
+    - aixfs
```
sxa, line 25: Why is this moved to after the XLC/X11 installations? Wouldn't moving this cause problems if the file systems weren't large enough to install XLC/X11? I think that's why it was at the start previously.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
```diff
@@ -15,28 +15,35 @@
...
+    - syslogs
 
+    # yum based installation of additional software
     - yum
 
+    # users needed - wait until here to be sure bash is installed
+    - jenkins_user
+    - zeus_user
+
+    # additional OSS packages (needing python)
```
sxa, line 35: Do these really need python? any is a java tool so shouldn't have any other prereqs on it

sxa (changes requested):
Some comments as requested - hopefully once we've got this in, adding others into the top level playbook will be easier

Haroon-Khel (changes requested):
See https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1704#issuecomment-734418604

Just to reiterate, the yum role should run before any role which involves decompressing and unzipping packages, as it installs tar and unzip. I thought this change had already gone in some time ago, but it seems im mistaken.

`fatal: [p9-aix1-ojdk06.osuosl.org]: FAILED! => {"changed": false, "msg": "Failed to find handler for \"/root/.ansible/tmp/ansible-tmp-1606411573.995723-14158-56646512312999/source\". Make sure the required command to extract the file is installed. Command \"/usr/bin/tar\" detected as tar type None. GNU tar required. Command \"unzip\" not found."}`

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
aixtools: This is before xlc installation - it WAS after syslogs - which is a configuration change.

What I am trying to do is get AIX BOS installation first, then AIX configuration changes (that does not require any non-BOS software) and then start looking at licensed software.

As the xlc roles use, potentially, the unarchive: module - the latest update has the role: yum before the xlc role calls.

aixtools: I agree, it looks like it is after, but it is not. No idea how the diff comes up with this.

Further - as to large enough - installp enlarges the filesystem if more space is needed.

aixtools:
> 
> 
> See [#1704 (comment)](https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1704#issuecomment-734418604)
> 
> Just to reiterate, the yum role should run before any role which involves decompressing and unzipping packages (such as x11, xlc13 and xlc16), as it installs tar and unzip. I thought this change had already gone in some time ago, but it seems im mistaken. I ran into the error while testing the linked pr, but thought the change would be more appropriate here
> 
> ```
> fatal: [p9-aix1-ojdk06.osuosl.org]: FAILED! => {"changed": false, "msg": "Failed to find handler for \"/root/.ansible/tmp/ansible-tmp-1606411573.995723-14158-56646512312999/source\". Make sure the required command to extract the file is installed. Command \"/usr/bin/tar\" detected as tar type None. GNU tar required. Command \"unzip\" not found."}
> ```

moved per https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1695/commits/104de6ac49b1c98b5e0a9e35c28c0058b6131e34

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
aixtools: This is what was in the original 'aix.yml' - I did not add it.

I will remove the comment test `(needing python)`.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
```diff
@@ -15,28 +15,39 @@
     swap_size: 4096
 
   roles:
+    # AIX base requirements
+    - X11
+    - openssl
+
+    # AIX configuration
```
sxa, line 22:
```suggestion
    # AIX configuration
    # installp will expand filesystems if needed so aixfs does not need to be before X11 and openssl
```

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
sxa: > What I am trying to do is get AIX BOS installation first, then AIX configuration changes (that does not require any non-BOS software) and then start looking at licensed software.

OK That reasoning is sound, but it might be worth explcitly saying that so that others modifying the file in the future adhere to the same policies

> I agree, it looks like it is after, but it is not. No idea how the diff comes up with this.

Looks ok now - it's showing where it should be based on your comments

> Further - as to large enough - installp enlarges the filesystem if more space is needed.

Yep that's fair, but I'd suggest adding that as a comment so it's clear to anyone modifying the playbook in the future

sxa (changes requested):
If we add in some comments to make it clear why the items have been sequenced the way we have this is good to merge. Let's aim to get it in Monday at the latest :-)

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
```diff
@@ -15,28 +15,39 @@
...
     - aixfs
     - syslogs
-    - X11
+
+    # yum based installation of additional software
+    # installs some gnu packages required by ansible modules
+    - yum
```
Haroon-Khel, line 28: Despite your reasoning [here](https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1704#issuecomment-734921703), running the x11 role before yum resulted in 
```
fatal: [p9-aix1-ojdk06.osuosl.org]: FAILED! => {"changed": false, "msg": "Failed to find handler for \"/root/.ansible/tmp/ansible-tmp-1606750087.0092869-25003-46895319352335/source\". Make sure the required command to extract the file is installed. Command \"unzip\" not found. Command \"/usr/bin/tar\" detected as tar type None. GNU tar required."}
```

I then moved the yum role so that it would run first, and then ran the x11 role subsequently. Ansible did not have a problem unzipping and untarring the archive. The openssl role also includes a archive extract, so the yum role should run before this too.

Haroon-Khel (changes requested):

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
Haroon-Khel: Just saw https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1695#discussion_r531706008

Out of curiosity, is there a technical reason for this order of installation? For example, will X11 install differently if it follows the yum and packages installation? And vice versa

Haroon-Khel:
Ive tested these changes, and those of https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1704, on ojdk06 (with the requested changes). The playbook ran clean. A build of jdk11 hotspot ran fine too. Running an openjdk sanity test resulted in one failure:
```
OLD.getZoneInfoOld()[1]=518
OLD.getZoneInfoOld()[2]=386
OLD.getZoneInfoOld()[3]=276
OLD.getAvailableIDs()=998, total=628
OLD.getAliasTable()=486, total=228
OLD.TotalTZ()=66 (ms)
NEW.getTimeZone()[1]=21
NEW.getTimeZone()[2]=1
NEW.getTimeZone()[3]=1
NEW.getAvailableIDs()=17, total=629
NEW.getAliasTable()=86, total=229
NEW.TotalTZ()=0 (ms)
STDERR:
javazic: warning: found last rules for Morocco inconsistent.
javazic: warning: found last rules for Morocco inconsistent.
javazic: warning: found last rules for Eire inconsistent.
java.lang.RuntimeException:   FAILED:  availableIds don't match
        at TestZoneInfo310.main(TestZoneInfo310.java:163)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at com.sun.javatest.regtest.agent.MainActionHelper$AgentVMRunnable.run(MainActionHelper.java:312)
        at java.base/java.lang.Thread.run(Thread.java:836)

JavaTest Message: Test threw exception: java.lang.RuntimeException
JavaTest Message: shutting down test


TEST RESULT: Failed. Execution failed: `main' threw exception: java.lang.RuntimeException: FAILED: availableIds don't match
```

Ive been informed that the test failure is related to a _common_ bug. Need to investigate this more.

Haroon-Khel:
@smlambert

smlambert:
Background:
The TestZoneInfo310 is most likely the result of using test material that does not match the TZ updates of the JVM.  It is the same issue reported here: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1356 (which was resolved by the adding RELEASE_TAG support, in other words telling the test pipeline which openjdk repo tag to use to pull test material from, so that test material matches the JVM).  In Grinders, where test pipeline is not passed a RELEASE_TAG, it will assume pull test material from tip (openjdk master branch). 

As discussed in meeting today, you can presume its non-blocking, but it is interesting that there is a skew.  Are you using nightly or releases builds(value of SDK_RESOURCE parameter) in your Grinder jobs?

Haroon-Khel:
@smlambert Im running the tests directly on the machine itself, since it is not in jenkins. Im using binaries from the adopt website, https://adoptopenjdk.net/releases.html?variant=openjdk11&jvmVariant=openj9

Haroon-Khel:
@aixtools Seeing as this test failure is non blocking, this pr is ready to be merged (after the requested changes have been put in)

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
aixtools: RPM packaging changes over time. Over the years there has been impact re: the order of installation. My expectation - today - is that it should not matter for most things.

Note: for 'yum'sh' to run, besides having to download it first (which I do not know, immediately, will work with just a base AIX system with no additional AIX software such as ftp to fetch the file) I do know that the script `yum.sh` does require the package `expect` which requires `tk` and `tcl` - plus it needs the `ftp` application - which on AIX 7.2 requires an additional fileset.

None of the requirements are visible when the installation is performed using `all`. IMHO - that is both insecure (installing things considered insecure by IBM since 1995 (`nis` is a good example) - but may also introduce unintended dependencies in the OJDK product - libraries get referenced not because they are really needed - but just because they are there and get included in dynamic load symbol lists.

When researching why things are broken - my quickest solution was to install on a minimal system and see what is demanded - and then verify that it is actually needed. This minimalist approach was not something I thought of on my own - but actually a early mentor over 30 years ago showed me the kinds of problems if prevented. The idiom: An ounce of prevention is better than a pound of cure". With this approach I have been able to resolve numerous issues others have looked at for months - with no explanation.

Back to you core question re: X11 and yum.sh - it should not matter - but I would rather not take the risk. Hence: install AIX bos first, and OSS additions later. If there are conflicts the package managers should report the conflicts. (note: should).

More specifically, the reason for the error you mention is the Ansible unarchive: module which assumes unzip. This assumption is not met by base AIX. Perhaps this message would not occur if the `vendor` file was not zipped and/or did not assume that gtar (which supports the `'z'` argument for uncompress versus using (for gzip compressed files) `gzip -dc <input> | tar xf -`

I am not expert enough to say Ansible has a bug. Rather I would say the design is not working on base AIX.

aixtools:
I hope I saw, and resolved everyone's requests. If I missed something - my apology in advance.

sxa (approved):
Thanks for the changes - LGTM other than the linter issue with extra spaces at the end of one of the lines which should be resolved before merging

aixtools:
Trailing space removed. Thx for the review.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
aixtools: Just to be clear - the yum role does run before the xlc ones. As yum, unzip, gtar, etc. are all 'external' to base AIX it really feels like putting that cart in front of the horse.

Again, I have no access to, nor do I want access to proprietary information - but is unzip even used? Or is it being demanded only because the Ansible unarchive module looks for it, regardless of need - just as it perhaps looks for gtar - when even AIX tar would work?

aixtools:
I'll add one more change - although I feel it is wrong to add something that is likely not used - and definitely not needed. But getting Ansible to modify unarchive: to not fail if unzip and gtar are not present is probably impossible.

Haroon-Khel:
@aixtools What if instead of using the `unarchive` module, we use the [copy](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html) module to transfer the package over, and then use the [shell](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html) module to run something like `tar xvf {{ path to x11 package }}` ?

aixtools:
> 
> 
> @aixtools What if instead of using the `unarchive` module, we use the [copy](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html) module to transfer the package over, and then use the [shell](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html) module to run something like `tar xvf {{ path to x11 package }}` ? Assuming tar is already present on the machine before the yum role.

Something like that would be my suggestion - but that can come later, i.e., the role calls can be moved when that is completed - as part of that PR perhaps.

Haroon-Khel (approved):
LGTM

PR #1696, opened 2020-11-20
aixPB: Modify X11 task use stat module rather than command.

aixtools:
Also, updated URL's with information on the X11 packages installed.

Rather than use command: lslpp - use stat on /usr/lpp/{{ directoryName }} to determine if a fileset is installed.

When one fileset is missing AND the Vendor Files exist - attempt an install, other only report something is missing, and continue. 

##### Checklist


Haroon-Khel:
@aixtools Do you intend to add install steps in a later pr? This pr can only be validated if it includes steps to install those packages

sxa (commented):
Hmmm I'm not sure this is a good idea as it will prevent us from being able to automatically install everything. If we are concerned about the use of Vendor_Files in certain circumstances we should be covering this with optional tags (or detecting if the `Vendor_Files` directory exists before trying) rather than removing the automated install which will prevent us from having a complete working playbook

aixtools (commented):
The key issue here is what filesets are actually needed.

ojdk05 is running, but does not have X11.vfb installed. The file I was asked to install - after the playbook had completed was X11.base.lib.

So, for now - only reporting that it is not installed. Further testing (when build test is done) will help determine what filesets are actually needed.

sxa:
Xvfb is the requirement in order to test java GUI functionality
Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1277

Haroon-Khel:
Ojdk05 now has Xvfb

aixtools:
> 
> 
> Ojdk05 now has Xvfb

Point being - the tests are not catching this. Anyway, per @sxa's comment - they are needed, per my earlier comment - the verification build suite used to verify ojdk05 missed that.

Are you @Haroon-Khel ? opening a new PR for that?

aixtools:
> 
> 
> @aixtools Do you intend to add install steps in a later pr? This pr can only be validated if it includes steps to install those packages

Yes, I'll add install steps - now that the requirements are more clear.

Initially, using 'vendor files' as that is what is known - with a test for whether they are available WHEN all the required filesets are not already installed.

karianna:
Linting bugs

aixtools:
The Docker/Centos6 CI failed check does not have anything to do with this. Final review requested.

aixtools:
> 
> 
> > Ojdk05 now has Xvfb
> 
> Point being - the tests are not catching this. Anyway, per @sxa's comment - they are needed, per my earlier comment - the verification build suite used to verify ojdk05 missed that.
> 
> Are you @Haroon-Khel ? opening a new PR for that?

ps. re: ojdk05 - not the correct level!

```
root@p9-aix1-ojdk05:[/root]oslevel -s -l 7100-05-06-2028
Fileset                                 Actual Level       Service Pack Level
-----------------------------------------------------------------------------
X11.vfb                                 7.1.3.0            7.1.5.30
```

I'll update (new PR I guess) - but this is also why having a static install package that gets unwrapped - is tricky to guarantee the correct version is installed.

Also note ojdk03 (an AIX 7.2 system) also has the 7.1.3.0 software:

```
# oslevel -s
7200-02-02-1810
# lslpp -L | grep vfb
  X11.vfb                    7.1.3.0    C     F    Virtual Frame Buffer Software
```

Basically, because of these differences I felt - not installing by default (is there a way to say do-tags, otherwise use an 'extra-var' as a condition for installing from Vendor Files.

Please see this (updated) URL re: vfb packages: https://www.ibm.com/support/pages/fileset-information-x11vfb

sxa (commented):
Unless I'm mistaken this looks like it now requires the `/Vendor_Files` to be already populated on the AIX machine - this is inconsistent with what we've done elsewhere where the `/Vendor_Files` is supposed to be on the client machine (the one which you run ansible on) and then copied across as required to the target system. With this change you have to put extra things on the machine before you start.

Also can we change the `register` variables to `_vfb_installed` instead of just `_vfb` to make the purpose a bit clearer?

aixtools:
Unless i am mistaken, which is possible, the vendor files for AIX have always been required first. Eg, i hear they must be copied from a laptop first, then the playbook is run. In other words, I have been trying to make it so that the playbook can run without the files, but uses them if they are there. 

In short, this is not my doing. I am just trying to follow what has been common practice for AIX til now.

karianna:
@aixtools merge conflict.

Haroon-Khel:
> Unless i am mistaken, which is possible, the vendor files for AIX have always been required first. Eg, i hear they must be copied from a laptop first, then the playbook is run. In other words, I have been trying to make it so that the playbook can run without the files, but uses them if they are there.

Yes this is the case, however we use ansible to copy them over.

sxa:
> > Unless i am mistaken, which is possible, the vendor files for AIX have always been required first. Eg, i hear they must be copied from a laptop first, then the playbook is run. In other words, I have been trying to make it so that the playbook can run without the files, but uses them if they are there.
> 
> Yes this is the case, however we use ansible to copy them over.

So just so 

>  i hear they must be copied from a laptop first

Not quite - they get copied from the machine that ansible is being invoked on. While that can be the AIX machine itself it's frequently a remote one. The playtbooks should copy it across to the target machine before using it, so in the general case there is no requirement for `/Vendor_Files` to exist on the target system.

aixtools:
> 
> 
> > > Unless i am mistaken, which is possible, the vendor files for AIX have always been required first. Eg, i hear they must be copied from a laptop first, then the playbook is run. In other words, I have been trying to make it so that the playbook can run without the files, but uses them if they are there.
> > 
> > 
> > Yes this is the case, however we use ansible to copy them over.
Maybe you have OTHER Ansible playboosk thay copy them over - but this is all I have to go from:
```
        - name: Transfer and extract X11
          unarchive:
            src: /Vendor_Files/aix/OpenGL_X11.tar
            dest: /tmp
            remote_src: no
          when: does_x11_exist.rc != 0 or does_X11_adt_ext_exist.rc != 0
          tags: x11
```
Where am I missing that these are not already on the target system? Love to learn.
> 
> So just so
> 
> > i hear they must be copied from a laptop first
> 
> Not quite - they get copied from the machine that ansible is being invoked on. While that can be the AIX machine itself it's frequently a remote one. The playbooks 
Again - see above: where - in the existing/historical playbook could I ever see this?

> should copy it across to the target machine before using it, so in the general case there is no requirement for `/Vendor_Files` to exist on the target system.

From the task above - how can they NOT exist on the target system? Is there something magical about the name "/Vendor_Files" that means it is coming from the management node? To me it just looks like a directory in the `/` filesystem named `/Vendor_files/`

Haroon-Khel:
The `remote_src: no` tells ansible to look at the machine on which ansible is running, as opposed to the remote node

aixtools:
And I was thinking `remote_src: no` meant it was already copied.

Haroon-Khel:
Yeah it is a little confusing. Here are the docs https://docs.ansible.com/ansible/latest/collections/ansible/builtin/unarchive_module.html#parameter-remote_src

aixtools:
> 
> 
> Yeah it is a little confusing. Here are the docs https://docs.ansible.com/ansible/latest/collections/ansible/builtin/unarchive_module.html#parameter-remote_src

Confusing is putting it lightly. Yet another example of aspects that beginner books miss - in teaching how to see the world through ansible eyes. And completely demoralizes me - what else about Ansible is 'on it's head', i.e., what else do I have absolutely backwards because I look at the playbook execution from the remote's perspective - rather than seeing all from the management node's perspective - in other words, I saw the remote node as the agentless actor - rather than agentless recipient - as it (the remote) must do some of this.

aixtools:
So, is there a way to verify the existence of something `local`? If not, well, - lost for the time being.

Haroon-Khel:
I think by adding `delegate_to: 127.0.0.1` to any task tells ansible to run the task on localhost. This could be used with the `stat` module to check if a file exists. However, in this particular use case I am not sure if it is worth the time since ansible already will check if the vendor files are present on the local machine when running the task, and will throw an error if not present

aixtools:
As the files are restricted I will think rethink this. Either adding the tag never, and require a tag statement to do the installation, or a stat: on something after the copy TO the remote.

sxa:
Hmmm personally I find that entirely logical rather than confusing or backwards. If you didn't have 

> So, is there a way to verify the existence of something `local`? If not, well, - lost for the time being.

In this situation do you need to? Preumably if you can capture the value of the `unarchive` operation that will give you a good indication of whether it exists or not. We could do a separate copy then unarchive locally but I would image/hope (but I don't know for sure!) that ansible optimises those two together in a pipeline when you use unarchive directly.

Bear in mind that ansible's primary purpose is to quickly set up lots of machines together. i.e. you'd normaly fire it at a group of machines to set them up in a large environment (we have over 100 systems just now) so yes by default you need to think about it in the way that makes most sense for that environment i.e. no source files should be required on the machine you're installing too.

From my perspective, I (almost) never run the `ansible-playbook` command on the target machine and part of the reason is precisely what you're seeing - it means I don't have to manually copy `Vendor_Files` across. In fact for the main UNIX playbooks that can be actively destructive, because it tries to set things like hostnmae to match the inventory file, so if you use `localhost` or an IP address in the inventory file then it resets the hostname to be that value, which is not good.

However, I agree that that's a default expectation of running remotely is not necessarily clear from our docs on running ansible and I'll look at clarifying there. Ultimately our playbooks are expected to be run by two different personas - the adoptopenjdk machine maintainers (i.e. us!) who will normally run it remotely to make it easier to scale, and end-users looking to build on their own machines (who will skip the `adoptopenjdk` tag) who are the ones you'd normally expect to be running locally.

aixtools:
> 
> 
> Hmmm personally I find that entirely logical rather than confusing or backwards. If you didn't have
> 
> > So, is there a way to verify the existence of something `local`? If not, well, - lost for the time being.
> 
> In this situation do you need to? Preumably if you can capture the value of the `unarchive` operation that will give you a good indication of whether it exists or not. We could do a separate copy then unarchive locally but I would image/hope (but I don't know for sure!) that ansible optimises those two together in a pipeline when you use unarchive directly.
> 
> Bear in mind that ansible's primary purpose is to quickly set up lots of machines together. i.e. you'd normaly fire it at a group of machines to set them up in a large environment (we have over 100 systems just now) so yes by default you need to think about it in the way that makes most sense for that environment i.e. no source files should be required on the machine you're installing too.
> 
> From my perspective, I (almost) never run the `ansible-playbook` command on the target machine 
Neither am I - I am running it from a "management node"
and part of the reason is precisely what you're seeing - 
What I see - is because I do not have `Vendor` files - so both remote and local failed. And being new to ansible, I was unaware that `remote` a) refers to the node position, not to the files themselves; or files position - relative to the target node.
IMHO: there is a lot that ansible documentation takes for granted - and maybe `ojdk` has some of those assumptions.
In any case: there has been some learning curve and also misunderstanding (on my part) when I heard people were copying from their laptop. I thought it was to the target - rather than a (new) management node.

it means I don't have to manually copy `Vendor_Files` across. In fact for the main UNIX playbooks that can be actively destructive, because it tries to set things like hostnmae to match the inventory file, so if you use `localhost` or an IP address in the inventory file then it resets the hostname to be that value, which is not good.
> 
> However, I agree that that's a default expectation of running remotely is not necessarily clear from our docs on running ansible and I'll look at clarifying there. Ultimately our playbooks are expected to be run by two different personas - the adoptopenjdk machine maintainers (i.e. us!) who will normally run it remotely to make it easier to scale, and end-users looking to build on their own machines (who will skip the `adoptopenjdk` tag) who are the ones you'd normally expect to be running locally.

Again, I am not running locally - so my dilemma is how to think about the situation when `Vendor` is not available on the management node (as well in cases when the management node and target node are the same node).

As to documentation: the contents of the archived Vendor files is not documented - and when I later got a listing I could see that the contents is not applicable for all platforms (re: X11 which is related to BOS; the compilers can be installed on differing BOS levels.)

karianna:
@aixtools FYI - this PR will need a rebase

aixtools:
> 
> 
> @aixtools FYI - this PR will need a rebase

I'll try to do that cleanly. My first rebases (years ago now, but still) went horribly - making it look asif the PR had 100's of changes, when it was only three.

In short, if I do see that I'll start again manually, and then force push. The change history is lost - but not super important I expect.

iirc: my change erases the above change (as I had already, effectively, removed the labels).

karianna:
@aixtools rebase please :-)

aixtools:
working on it.

karianna:
@aixtools needs a rebase sorry

sxa:
@aixtools I've just done the rebase - is this complete from your point of view?

aixtools:
Hmm, thought I had rebased - I'll fetch and test.

aixtools:
One last typo - and good to go.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
```diff
@@ -1,46 +1,88 @@
...
+      register: _vendor_copied
 
-- name: Install IBM X11 Extensions - installp
-  command: installp -aXYgd /tmp/x11/X11.adt X11.adt.ext
-  ignore_errors: yes
-  when: does_X11_adt_ext_exist.rc != 0
-  tags: x11
+    - name: Install IBM X11 Extensions - installp
+      command:
+        installp -e /tmp/x11.log -aXYgd /tmp/x11 X11.adt.ext X11.vfb X11.base.lib
```
Haroon-Khel, line 53: During the install step I get the following error
```
TASK [Install IBM X11 Extensions - installp] ********************************************************************************************************************************************
task path: /Users/hkhel/AdoptOpenJDK/openjdk-infrastructure/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml:51
fatal: [p9-aix1-ojdk06.osuosl.org]: FAILED! => {"changed": true, "cmd": ["installp", "-e", "/tmp/x11.log", "-aXYgd", "/tmp/x11", "X11.adt.ext", "X11.vfb", "X11.base.lib"], "delta": "0:00:00.066162", "end": "2021-03-05
12:55:41.662029", "msg": "non-zero return code", "rc": 220, "start": "2021-03-05 12:55:41.595867", "stderr": "installp: The specified device /tmp/x11\n\tis not a valid device or file.", "stderr_lines": ["installp: The specified
device /tmp/x11", "\tis not a valid device or file."], "stdout": "", "stdout_lines": []}
...ignoring
```

I think installp does not like installing all 3 in one line?

Haroon-Khel (changes requested):

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
Haroon-Khel: Same error when running on the machine directly
```
root@p9-aix1-ojdk06:[/root]installp -e /tmp/x11.log -aXYgd /tmp/x11 X11.adt.ext X11.vfb X11.base.lib
installp: The specified device /tmp/x11
        is not a valid device or file.
```

Haroon-Khel: I suspect each package's install directory/files need to be specified separately. Hence why they are separate in the role

aixtools: No, installp has no problem with multiple package names. Years of experience. This is something different - and I would need access to the Vendor_Files to examine the issue. I suspect something with the .toc file. Or, the tar file is not unpacking in /tmp/x11 - but elsewhere (e.g., /tmp/x11/x11 ?).

FYI: when .toc does not exist installp reads all files in the directory to create a .toc (table of contents)

aixtools:
closing and opening to see if this gets the same yamllint compliant re: the block: statement

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
aixtools: Under the assumption the tar contents are ./x11/FileNames I removed the creation of /tmp/x11 and renamed the destination of the unarchive: to /tmp

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
```diff
@@ -1,46 +1,83 @@
...
-- name: Transfer and extract X11
-  unarchive:
-    src: /Vendor_Files/aix/OpenGL_X11.tar
-    dest: /tmp
-    remote_src: no
-  when: does_x11_exist.rc != 0 or does_X11_adt_ext_exist.rc != 0
-  tags: x11
+    - name: Install IBM X11 Extensions - installp
+      command:
+        installp -e /tmp/x11.log -aXYgd /tmp/x11 X11.adt.ext X11.vfb X11.base.lib
```
Haroon-Khel, line 48: By only specifying /tmp/x11, the system isnt able to find installation media for X11.adt.ext

Ansible error log:
```
FAILURES\n
--------\n
  Filesets listed in this section failed pre-installation verification\n  and will not be installed.\n\n  Missing Filesets\n
  ----------------\n
  The following filesets could not be found on the installation media.\n
  If you feel these filesets really are on the media, check for typographical\n
  errors in the name specified or, if installing from directory, check for\n
  discrepancies between the Table of Contents file (.toc) and the images that\n
  reside in the directory.\n
\n

    X11.adt.ext                              \n
    X11.base.lib
```

And then from running the command directly on the machine

```
root@p9-aix1-ojdk06:[/root]installp -aXYgd x11 X11.adt.ext X11.vfb X11.base.lib
+-----------------------------------------------------------------------------+
                    Pre-installation Verification...
+-----------------------------------------------------------------------------+
Verifying selections...done
Verifying requisites...done
Results...

FAILURES
--------
  Filesets listed in this section failed pre-installation verification
  and will not be installed.

  Missing Filesets
  ----------------
  The following filesets could not be found on the installation media.
  If you feel these filesets really are on the media, check for typographical
  errors in the name specified or, if installing from directory, check for
  discrepancies between the Table of Contents file (.toc) and the images that
  reside in the directory.

    X11.adt.ext                              
    X11.base.lib                             

  << End of Failure Section >>

```

Only by running `installp -aXYgd /x11/X11.adt X11.adt.ext` could I get X11.adt.ext to install.

Also, I dont think the Vendor file for x11 has X11.base.lib on it

```
root@p9-aix1-ojdk06:[/root]ls -la x11
total 61432
drwxr-xr-x 2  501 staff       256 Oct 22  2019 .
drwx------ 6 root system     4096 Mar 18 15:32 ..
-rw-r--r-- 1  501 staff      5933 Aug 20  2019 .toc
-rw-r--r-- 1  501 staff  21670912 Apr 17  2018 OpenGL.OpenGL_X.dev
-rw-r--r-- 1  501 staff  35705856 Apr 17  2018 OpenGL.OpenGL_X.rte
-rw-r--r-- 1 root system  5476352 Oct 22  2019 X11.adt
-rw-r--r-- 1  501 staff     33792 Apr 17  2018 X11.vfb
```

hence why only vfb and adt.ext were installed with vendor file previously

Haroon-Khel (changes requested):

aixtools (commented):
removed the install reference to X11.base.lib, but still check for it's presence, and report if it is missing.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
aixtools: removed from installp command, but still check it is installed - or not.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
```diff
@@ -1,46 +1,86 @@
...
-    dest: /tmp
-    remote_src: no
-  when: does_x11_exist.rc != 0 or does_X11_adt_ext_exist.rc != 0
-  tags: x11
+# X11.base.lib is not included in the so-called vendor files
+# So no attempt is made to install it.
+# Still reports if it is missing
+    - name: Install IBM X11 Extensions - installp
+      command:
+        installp -e /tmp/x11.log -aXYgd /tmp/x11 X11.adt.ext X11.vfb
```
Haroon-Khel, line 51: The single installp command will still only install X11.vfb and not X11.adt.ext

```
root@p9-aix1-ojdk06:[/tmp]installp -e /tmp/x11.log -aXYgd /tmp/x11 X11.adt.ext X11.vfb
+-----------------------------------------------------------------------------+
                    Pre-installation Verification...
+-----------------------------------------------------------------------------+
Verifying selections...done
Verifying requisites...done
Results...

FAILURES
--------
  Filesets listed in this section failed pre-installation verification
  and will not be installed.

  Missing Filesets
  ----------------
  The following filesets could not be found on the installation media.
  If you feel these filesets really are on the media, check for typographical
  errors in the name specified or, if installing from directory, check for
  discrepancies between the Table of Contents file (.toc) and the images that
  reside in the directory.

    X11.adt.ext                              

  << End of Failure Section >>
```
To install X11.adt.ext, the install directory will need to point to /tmp/x11/X11.adt which will require a separate installp command, unless installp has a flag where you can provide two install directories

Haroon-Khel (changes requested):

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/X11/tasks/main.yml
aixtools: I suspect the table of contents in the TAR file are not current - so added steps to remove previous x11.log (if present) and also any /tmp/x11/.toc - so that installp will create a new one - consistent with the directory contents.

Haroon-Khel (approved):
It now installs as expected. LGTM 👍🏻

PR #1698, opened 2020-11-20
aixPB: xlc16 changes

aixtools:
##### Checklist


ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/xlc_v16/tasks/main.yml
```diff
@@ -13,37 +13,54 @@
...
-    src: /Vendor_Files/aix/XLC/XL_C_Cpp_FOR_AIX_V16.1_EMG.tar.Z
-    dest: /tmp
-    remote_src: no
-  when: xlc16.stat.islnk is not defined
-  tags: xlc16
+- name: Vendor File processing
+  block:
+    - name: Check for Vendor file
+      stat:
+        path: /Vendor_Files/aix/XLC/XL_C_Cpp_FOR_AIX_V16.1_EMG.tar.Z
```
Haroon-Khel, line 20: Same as https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1697#discussion_r530986496

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/xlc_v16/tasks/main.yml
```diff
@@ -13,37 +13,54 @@
...
-  command: installp -aXYg -e /tmp/usr/install.log -d /tmp/usr/sys/inst.images all
-  register: result.xlc
-  ignore_errors: yes
-  when: xlc16.stat.islnk is not defined
-  tags: xlc16
+    - name: Install IBM XLC16 - installp
+      command: installp -aXYg -e /tmp/usr/install.log -d /tmp/usr/sys/inst.images all
+      register: _installp
+      ignore_errors: yes
+      when: _images.skipped == false
```
Haroon-Khel, line 35: I think the previous task has trouble registering "_images", despite the task succeeding, as it causes an error when trying to access "_images.skipped":
```
fatal: [p9-aix1-ojdk06.osuosl.org]: FAILED! => {"msg": "The conditional check '_images.skipped == false' failed. The error was: error while evaluating conditional (_images.skipped == false): 'dict object' has no attribute 'skipped'\n\nThe error appears to be in '/Users/hkhel/AdoptOpenJDK/openjdk-infrastructure/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/xlc_v16/tasks/main.yml': line 33, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Install IBM XLC16 - installp\n      ^ here\n"
```

Haroon-Khel: I think because the previous task `name: Transfer and Extract XLC16` wasnt skipped, ansible doesnt add the `.skipped` attribute to `_images`

aixtools: I'll have to fake something - for when the files are there. Will take some time.

sxa (commented):
In the generic UNIX and Windows playbooks we have a tag `vendor_files` which can be skipped for anyone that doesn't have those files available - all roles that require it are tagged as such so they can be excluded.

I propose doing the same here with the AIX playbook as a preference to having to ignore errors.

aixtools:
ok. I'll look at those re: vendor_files. 

I'll still want to change the verification to use the ansible `stat:` command rather than execute `command: lslpp -L ...`

aixtools:
The logic seems to be quite different:
```
    - role: Get_Vendor_Files
      tags: [vendor_files, adoptopenjdk, jenkins_user, nagios_plugins, superuser]
```

I'll assume, for the moment, that the role `Get_Vendor_Files` only copies files over - whereas with AIX the vendor files are often being transferred and then something is done with them.

It can be done - but also requires a rewrite of everything (e.g., create an AIX `Get_Vendor_Files` role - to collect things, and then create when clauses to only do installs etc, if the copy of the vendor files is there for the later role.

I notice the later roles in UNIX and Windows do not have the vendor_files tag - but it looks like like lots of stat.vendor_files.exists conditionals.

I would prefer to:
short-term - make the roles less "hard-stop" if the vendor_files are not available (and also add the tag `locally` to skip over the copy and other actions if a --skip-tags argument is given.

long-term: - move all `get` actions to a single role that get executed early on - as UNIX and Windows playbooks do AND - adjust the roles that depend on vendor files to be able to work around the files are there, or not.

aixtools:
> 
> 
> In the generic UNIX and Windows playbooks we have a tag `vendor_files` which can be skipped for anyone that doesn't have those files available - all roles that require it are tagged as such so they can be excluded.
> 
> I propose doing the same here with the AIX playbook as a preference to having to ignore errors.

So, the current PR has the tag vendor_flags to skip the block - on purpose - but still being graceful about ignore_errors - as the installp command that followed also did the same.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/xlc_v16/tasks/main.yml
aixtools: When not skipped, `variable`.skipped is not defined - so added a check that the variable is defined

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/xlc_v16/tasks/main.yml
```diff
@@ -2,47 +2,59 @@
...
+    path: /usr/lpp/xlCcmp.16.1.0.license
   register: xlc16
   tags: xlc16
 
-- debug:
+- name: Announce skip of download and installation
+  debug:
     msg: "xlc16 installed, skipping download and installation"
-  when: xlc16.stat.islnk is defined
+  when: xlc16.stat.isdir is defined and xlc.stat.isdir
```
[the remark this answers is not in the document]
aixtools: The error above is that one was xlc16.stat and the other a typo (xlc.stat).
I leave it double (isdefined, isTrue) because during my experiments with stat: I frequently ran into something such as xyz.stat.isdir failing because it was not defined - and that is not the same as false. -- correcting the typo

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/xlc_v16/tasks/main.yml
```diff
@@ -2,47 +2,59 @@
...
+
+    - debug: msg='Error messages from the installp command can be ignored.'
+      when: _installp is defined
 
-- debug: msg='Erorrs from the previous installp command normal'
+    - name: Clean XLC16 tmp files
+      file:
+        path: /tmp/usr
+        state: absent
+      when: _images.skipped == false
```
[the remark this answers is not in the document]
aixtools: likewise - this is similar - needs a conditional checking the variable 'is defined' - but I agree, can always be run without harm.

Haroon-Khel (commented):
The new changes run fine, both when xlc16 is and isnt on the machine. I dont think a build and test is necessary, but ill run one on ojdk06 just to be safe

aixtools:
> 
> 
> The new changes run fine, both when xlc16 is and isnt on the machine. I dont think a build and test is necessary, but ill run one on ojdk06 just to be safe

Your change request is satisfied?

sxa (approved):
LGTM :-)

karianna:
@aixtools you can merge this now when you're ready

PR #1710, opened 2020-11-26
delete all perf machines

gdams:
As discussed on the TSC call

### DIFF
--- a/ansible/plugins/inventory/adoptopenjdk_yaml.py
+++ b/ansible/plugins/inventory/adoptopenjdk_yaml.py
@@ -41,7 +41,7 @@ valid = {
   'arch': ('armv7', 'armv8', 'ppc64le', 'ppc64', 'x64', 's390x', 'arm64', 'sparcv9'),
 
   # valid roles - add as necessary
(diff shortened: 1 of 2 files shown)


ansible/plugins/inventory/adoptopenjdk_yaml.py
```diff
@@ -41,7 +41,7 @@
   'arch': ('armv7', 'armv8', 'ppc64le', 'ppc64', 'x64', 's390x', 'arm64', 'sparcv9'),
 
   # valid roles - add as necessary
-  'type': ('build', 'test', 'infrastructure', 'perf', 'docker'),
```
sxa, line 44: Hmmm do we want to remove this while there's a possibility of getting more later? Gut feel would be leave it in for now so that it's easier for someone to add it in again later

gdams: the inventory parser fails if there are no hosts in the specified group so I had to remove it. it will be very easy to add it back

sxa (approved):

PR #1711, opened 2020-11-26
Revert "actions: add macOS playbook tester (#1640)"

sxa:
macos github checks are failing. If we're not going to fix them as a matter of urgency, we should disable them as they provide zero value.

Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1711

This reverts commit 60c0bdc15a6cfbe7d0fffc00e8b45a563e7b1ccd. (not as a direct `git revert` as that would have removed the playbook too which was done in the same commit)

[EDIT: As part of resolving a merge conflict I've streamlined one of the `when` clauses to put it on one line so it doesn't disrupt the look of the file compared to other parts of it]


aahlenst (approved):

gdams:
closing in favour of https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1665

Issue #1722, opened 2020-11-28
Solaris playbook cannot locate CSWpkgutil.pkg


sxa:
Please put the name of the software product (and affectred platforms if relevant) in the title of this issue

- [ ] Missing install
- [x] Bug in ansible playbook
- [ ] Request for new playbook addition

Details:
```
TASK [Common : Fetch CSWpkgutil.pkg] *******************************************

fatal: [127.0.0.1]: FAILED! => {"changed": false, "checksum": "f3ec18e5de2da7f4dc19e9464ad42a5f9b43ead6", "msg": "Source /export/home/vagrant/.ansible/tmp/ansible-tmp-1606569858.04-21931-70421185104931/source not found"}
```
Not immeidately clear why. Modifying the source path to be `../files/CSWpkgutil.pkg` (relative the the playbook that invokes it) doesn't appear to resolve it. Debug output (with the chane to use `src: ../files/CSWpkgutil.pkg` is below:

`TASK [Common : Fetch CSWpkgutil.pkg] ******************************************************************************************************************************************************************************
task path: /home/sxa/ws/adoptopenjdkPBTests/openjdk-infrastructure-master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml:18
<127.0.0.1> ESTABLISH SSH CONNECTION FOR USER: vagrant
<127.0.0.1> SSH: EXEC ssh -vvv -F /dev/null -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o Port=2204 -o 'IdentityFile="/home/sxa/ws/adoptopenjdkPBTests/openjdk-infrastructure-master/ansible/id_rsa"' -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o 'User="vagrant"' -o ConnectTimeout=30 -o ControlPath=/home/sxa/.ansible/cp/79ad8e6b4d 127.0.0.1 '/bin/sh -c '"'"'echo ~vagrant && sleep 0'"'"''
[...]

PR #1736, opened 2020-12-04
pbTests: Set remote_tmp for Unix VPC runs

Willsparker:
Fixes: #1722 

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) Running now: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/953/ (Mainly checking Solaris works, and the fix doesn't break anything else)

### DIFF
--- a/ansible/pbTestScripts/vagrantPlaybookCheck.sh
+++ b/ansible/pbTestScripts/vagrantPlaybookCheck.sh
@@ -238,7 +238,7 @@ startVMPlaybook()
 	ssh-keygen -R $(cat playbooks/AdoptOpenJDK_Unix_Playbook/hosts.unx)
 	
 	sed -i -e "s/.*hosts:.*/- hosts: all/g" playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
-	awk '{print}/^\[defaults\]$/{print "private_key_file = id_rsa"; print "timeout = 30"}' < ansible.cfg > ansible.cfg.tmp && mv ansible.cfg.tmp ansible.cfg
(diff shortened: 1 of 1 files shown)


sxa (commented):
Hmmm why is this needed when https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/14272c35b38fe04a54759ba3425d181b5d975390/ansible/ansible.cfg#L31 is in the file already? Can we do something that will make that activate properly in VPC?

It does worry me that we're forcing the directory like this so would be good to understand why it doesn't work (or whether we can use a neutral directory like `/tmp`) or at least document why we've got this change

Willsparker:
> is in the file already? Can we do something that will make that activate properly in VPC?
VPC doesn't use the inventory, so `hosts:solaris` means nothing to Ansible. We could force it to, by making the temporary host file into something more like the inventory; i.e.

```
hosts:
  $OS-vagant: {ip: [127.0.0.1]:$vagrantPort}
```
> It does worry me that we're forcing the directory like this so would be good to understand why it doesn't work (or whether we can use a neutral directory like /tmp) or at least document why we've got this change

I believe in most cases `$HOME/.ansible/tmp` is already the case, it just isn't for Solaris (apparently). However, yeah, I get the concern. I could try the `/tmp` directory, and then just make sure the `vagrant` user has the perms to execute from there in the Vagrantfile's provisioning script.

sxa (approved):
> VPC doesn't use the inventory, so `hosts:solaris` means nothing to Ansible

Hmmm ok we could override the hostname, but if it's just a pattern match there then that explains it. I am somewhat concerned that we have a setup that requires additional tweaking to be able to work though.

gdams (approved):

PR #1743, opened 2020-12-07
move CRB packages out of common build tool list

sej-jackson:
Signed-off-by: Sej <sarah_jackson@uk.ibm.com>

The `libdwarf-devel` and `libmpc-devel` packages are no longer in the base repos for RedHat 8, and have been moved to the CodeReady Linux Builder (CRB) repository which is restricted access.  

Moving these packages to the `Additional_Build_Tools_NOT_RHEL8` list ensures that they are still installed on earlier versions of RedHat.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
@@ -24,9 +24,7 @@ Build_Tool_Packages:
   - glibc-devel
   - gmp-devel
   - libcurl-devel
-  - libdwarf-devel                # OpenJ9
   - libffi-devel
-  - libmpc-devel
   - libpng-devel
   - libXext-devel
   - libXi-devel                   # JDK12+ compilation
@@ -47,6 +45,8 @@ Build_Tool_Packages:
   - zip
 
 Additional_Build_Tools_NOT_RHEL8:
+  - libdwarf-devel                # now in CodeReady Linux Builder (CRB) repo
+  - libmpc-devel                  # now in CodeReady Linux Builder (CRB) repo
   - ntp                           # in RHEL8, ntp package is replaced by chrony
 
 Additional_Build_Tools_RHEL8:
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
@@ -26,7 +26,6 @@ Build_Tool_Packages:
   - gmp-devel
   - java-1.8.0-openjdk-devel
   - libcurl-devel
-  - libdwarf-devel                # OpenJ9
   - libpng-devel
   - libXext-devel
   - libXi-devel                   # JDK12+ compilation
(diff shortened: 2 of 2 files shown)


Willsparker (commented):
Makes sense; Ideally, we'd want to still have the packages for RHEL8, could we add a task to enable that repo? [This webpage](https://computingforgeeks.com/enable-codeready-builder-repo-on-rhel/) shows how, but I'm not entirely sure if we should be using `subscription-manager` in the playbooks. ping @sxa

sxa:
> Makes sense; Ideally, we'd want to still have the packages for RHEL8, could we add a task to enable that repo? [This webpage](https://computingforgeeks.com/enable-codeready-builder-repo-on-rhel/) shows how, but I'm not entirely sure if we should be using `subscription-manager` in the playbooks. ping @sxa

I'm ok with switching this off on RHEL8 for now (or perhaps we could check if CRB is enabled first?) but what's the option for CentOS8?

Willsparker:
> what's the option for CentOS8?

That would be the PowerTools repository (See #1750 ). There's a link in the issue which suggests that repository can also be enabled for RHEL8, so we may not have to bother with CRB- if PowerTools contains `libmpc-devel` that is

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
```diff
@@ -47,6 +45,8 @@ Build_Tool_Packages:
   - zip
 
 Additional_Build_Tools_NOT_RHEL8:
+  - libdwarf-devel                # now in CodeReady Linux Builder (CRB) repo
```
sxa, line 48: I'm worried about expanding this section. While I accept that this the name wasn't defined in this PR, `NOT_RHEL8` doesn't feel right for this. It was ok when it was just NTP (Since that has an explicit replacement in RHEL8) but these don't. I'd rather it was called somthign that implied `<8` or `<=7` as the same will likely apply to 8 and later.

sej-jackson: Well it's with RHEL8 that the packages have been moved, but I get your point that it doesn't take into account future releases..... However, the RedHat statement about packages ending up in CRB is that they're unsupported, so does that mean they might disappear altogether in a future release?

I think the naming is fairly clear currently, and there's also other sections like `Java_NOT_RHEL6_PPC64` (with the comment "Not RHEL8 either") which have a similar problem, so if we change one, we should review the rest also.  Would you be willing to let this go to a fresh issue where new names can be discussed and agreed?

sej-jackson: Also, I'd like to name the Centos change in this PR too for Will, and that would have similar naming issues.

sxa: `Java_NOT_RHEL6_PPC64` isn't so bad IMHO because we won't supported anything earlier than 6, so it's clear

sej-jackson:
@Willsparker I've now added the equivalent change for the `CentOS.yml` and updated the comment.  No `libmpc-devel` for this one, and `libdwarf-devel` had already been added to "Additional_Build_Tools_NOT_CentOS8", but just not removed from "Build_Tool_Packages" which is why it still failed on CentOS8.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
sej-jackson: OK, I've raised #1757 to do this properly once we've agreed what names should be used.

sxa (approved):
LGTM - The VPC run at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/OS=CentOS8,label=vagrant/963/console shows that the playbook went through but (as expected) the OpenJ9 buid failed when it tried to access the `libdwarf` headers.

PR #1747, opened 2020-12-08
inventory: add linux installer node

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1577

Machine is in [jenkins](https://ci.adoptopenjdk.net/computer/build-godaddy-ubuntu1604-x64-1/). Bastillion will need to be updated.

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -41,6 +41,7 @@ hosts:
 
       - godaddy:
           rhel7-x64-1: {ip: 160.153.248.216, user: adoptopenjdk}
+          ubuntu1604-x64-1: {ip: 160.153.254.64}
 
       - inspira:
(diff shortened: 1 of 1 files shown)


karianna:
@sxa - Do we want to split this out and have a specific installer section or do many of the build machines double as installers?

sxa:
> @sxa - Do we want to split this out and have a specific installer section or do many of the build machines double as installers?

A lot of the machines (e.g. windows) double up I believe although not this one. I'm ok with doing this for now, but the difference between build/install would be if either the tools required couldn't be put on the build machines, or if we had secret stuff (e.g. signing keys) that we wanted to keep separate.

sxa (approved):
System now added to Bastillion, therefore approving and merging :-)

Issue #1750, opened 2020-12-09
Ansible request for RHEL/CentOS 8 libmpc-devel & libdwarf-devel


Willsparker:
Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1743

- [x] Missing install

Details:
The `libmpc-devel` and `libdwarf-devel` packages have moved to the CodeReady Linux Builder (CRB) repo, and the PowerTools repo for RHEL and CentOS respectively (though we may be able to use PowerTools for RHEL..?).
The referred PR above will remove them from the list of packages to install in the meantime, however we need both the packages to build OpenJ9, so we need to find a way of getting them in.

Willsparker:
Checking [this](https://centos.pkgs.org/8/centos-powertools-x86_64/) we can get `libmpc-devel` and `libdwarf-devel` from the repos,so I'll test with CentOS and then try find a RHEL machine to test on, somewhere

Willsparker:
Checking on a CentOS8 Vagrant VM, the Powertools repo is already installed from the start, it's just a case of enabling it. Unfortunately, attempting to enable it either through `yum` or `dnf` commands didn't seem to work, but enabling it manually did (i.e. using `vi`). This would be easy enough to ansiblise, however.

Willsparker:
Having checked the `test-aws-rhel8-x64-1` machine, unfortunately, the PowerTools repository doesn't exist on the machine ( and in a brief Google search, I couldn't find if it to be installable ). That unfortunately leaves us with having to deal with the CRB repo and `subscription-manager` if we want to build OpenJ9 on RHEL8+. Running a few commands on the machine whilst I was testing, I found that that particular machine hasn't been registered yet, and so repos were available via `subscription-manager`. 

```
[root@54 ~]# subscription-manager repos
This system has no repositories available through subscriptions.
[root@54 ~]# subscription-manager list --available
This system is not yet registered. Try 'subscription-manager register --help' for more information.
```

While we can likely ansibilise the `subscription-manager` commands, registering the subscription may not be possible.

Willsparker:
[...]

PR #1759, opened 2020-12-14
UnixPB: Enable PowerTools repo CentOS8, install libdwarf-devel

Willsparker:
Ref: #1750 

We had the task to enable the PowerTools repo in there, it just had the incorrect path. `libdwarf-devel` resides in the PowerTools repo now, hence the need to enable it, otherwise we can't build OpenJ9.

Note: `libmpc-devel` can also be installed from the PowerTools repo, but looking at #1743 , we never had it in the CentOS build package list? Either way, If we need it, should be easy enough to just add it to the package list :-)

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) . See : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/968/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml
@@ -23,9 +23,9 @@
     - ansible_distribution_major_version == "8"
   tags: patch_update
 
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM

Issue #1717, opened 2020-11-27
Nagios: Add service checks to check synchronisation in Jenkins/inventory/Nagios


Willsparker:
ref: #1716, #1229 

To ensure no machines pop up that we're not expecting them to, similarly, to make sure we're monitoring all the machines too. Could also put in checks to make sure the inventory and Jenkins are consistent (i.e. https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/619#issuecomment-520874015 ).

As mentioned in #1716 , I've already written a script that detects the difference between Nagios and Jenkins, so that will just require some changing to work as a Nagios Plugin :-) We also need to determine what we want to be in an exception list, as I know that some machines are added to Jenkins as experimental machines  - maybe we could start prefixing them with `expr` ... ?

Willsparker:
I've written a script that can retrieve the list of machines in `inventory.yml`
```
export inv=$(wget -qO- https://raw.githubusercontent.com/AdoptOpenJDK/openjdk-infrastructure/master/ansible/inventory.yml)
[[ $? != 0 ]] && echo "PANIC" && exit 3

# Put EOF at end of inv file
inv=$inv$'\n'EOF

Inventory_Server_List=

IFS=
mapfile -t category_array < <( echo $inv | grep -P "^\s{2}-\s" | cut -d- -f2 | tr -d ": -" )

for cat_index in ${!category_array[@]}; do
  beginPattern=${category_array[$cat_index]}
  endPattern=${category_array[$cat_index+1]}

  if [[ $endPattern == "" ]]; then
    endPattern="EOF"
  fi

  inv_section=$(echo $inv | awk '/'$beginPattern'/,/'$endPattern'/')
  mapfile -t prov_array < <( echo $inv_section | grep -P "\s{4}-\s*" | tr -d " :-" )

  for prov_index in ${!prov_array[@]}; do
    provBegin=${prov_array[$prov_index]}
    provEnd=${prov_array[$prov_index+1]}
    if [[ ${prov_array[$prov_index+1]} == "" ]]; then provEnd=$endPattern; fi

    prov_section=$(echo $inv_section | awk '/'$provBegin:'/,/'$provEnd:'/')

    mech_list=$(echo $prov_section | awk '/ip/ { print $1 }' | tr -d ":")
    IFS=$'\n'

    for machine in $mech_list; do
       Inventory_Server_List+="$beginPattern-$provBegin-$machine  "
    done
[...]

PR #1762, opened 2020-12-15
unixPB: Put in check_inventory Nagios plugin

Willsparker:
Fixes: #1717  

Committing the plugin that is being used on the Nagios Server that is used to check the disparity between machines in the Inventory / Jenkins / Nagios. 
This plugin is already in Nagios, and working (see the referenced issue), but if there's ways to improve the script, I'll update those on the machine :-) 
I've also separated the plugins that are meant to be put on the Nagios Server from the plugins that are put on the machines that are to be monitored.

##### Checklist


gdams (approved):

karianna (approved):
lgtm

PR #1767, opened 2020-12-16
Docs: include troubleshoot for windows playbook

Haroon-Khel:

Ive included a solution in the docs, regarding running the windows playbook, which fixes a python related error. Ive experienced it when running from a mac, but I dont think it is exclusive to macs.

```
objc[39516]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[39516]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
ERROR! A worker was found in a dead state
```

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -88,7 +88,7 @@ ansible_winrm_read_timeout_sec: 630
 
 Additional information about `winrm` variables can be found [here](https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/user_guide/windows_winrm.rst#inventory-options)
 
-You may encounter a python related error
+If running from a mac, you may encounter a python related error
 ```
 objc[39516]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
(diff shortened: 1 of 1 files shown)


Willsparker:
I believe I had this issue when I was running any playbook from a mac, and put it into `ansible/README.md` (See https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1184/files#diff-fac2c4b2da0d1130d0f929abfff14008a96373c769abad4fdd1c8ee2ffca5226R181 ). However, I prefer the more verbose message you have here :-)

ansible/README.md
```diff
@@ -88,6 +88,15 @@ ansible_winrm_read_timeout_sec: 630
 
 Additional information about `winrm` variables can be found [here](https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/user_guide/windows_winrm.rst#inventory-options)
 
+You may encounter a python related error
```
Willsparker, line 91: We've only seen this when running the playbook from a mac, AFAIK. Until it's an issue with anything else, could you specify that this is only for mac, just in case this does unforeseen things on any other Unix system :-)

Willsparker (changes requested):

Willsparker (approved):
LGTM

PR #1774, opened 2020-12-21
Inventory: add alibaba machines

Haroon-Khel:
Added the machines build-alibaba-win2012r2-x64-1 and 2 to the inventory file. They are already in jenkins

@sxa Bastillion will need to be updated, if it already hasnt been

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -27,6 +27,10 @@ hosts:
 
   - build:
 
+      - alibaba:
+          win2012r2-x64-1: {ip: 8.208.11.212, user: Administrator}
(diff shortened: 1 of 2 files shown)


sxa (approved):
Bastillion only manages the UNIX-based systems, so this is good to land :-)

sxa (changes requested):
Actually I'm going to reject this until the linter runs clean ... since alibaba is a new provider in here `./plugins/inventory/adoptopenjdk_yaml.py` will need to be updated to have it in the `provider` list

Willsparker (approved):
LGTM

sxa (approved):

PR #1778, opened 2020-12-22
github: remove CentOS6 PR check

sxa:
To avoid confusion and not put off potential contributors we should remove this check for now until https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1745 is resolved.

##### Checklist

- [ ] FAQ.md updated if appropriate
- [ ] inventory changes, ensure bastillion is updated accordingly

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,11 +1,11 @@
 name: Docker
 
 on:
-  pull_request:
-    paths:
-    - ansible/**
-    branches:         
-    - master
+#  pull_request:
+#    paths:
+#    - ansible/**
+#    branches:         
+#    - master
   push:
     paths:
     - ansible/**


gdams:
My only concern with this is that it's what is used to update the docker images for the builds.

sxa:
> My only concern with this is that it's what is used to update the docker images for the builds.

There looks to be actions in there associated with PRs and pushes - is the push executed when it's merged? [NOTE: Original PR had the wrong but comented out - have adjusted to leave it on push] Otherwise aren't we pushing and updating the image on every PR creation regardless of whether it's approved?

Either way, presumably at the moment it's unable to update so is fundamentally non-functional right now and only leading to confusion?

sxa:
> wasn't there a proposed fix coming from Will?

I believe you're referring to https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1765/files which has already been merged and does not appear to have fixed the issue in the github checks.

@Willsparker are you still looking at a further fix?

Willsparker:
That only solves part of the problem. See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1745
The same sed change needs to be done on the SCL repo - but due to that being added in the playbook, it needs to be an ansible task too.

karianna:
@Willsparker Is that something which can be resolved soon or should we land this instead?

Willsparker:
I'll have it done by today :-)

Willsparker:
As of #1780, the CentOS6 PR check works. ([VPC](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/977/OS=CentOS6,label=vagrant/console) also shows that the C6 playbook is working on a vagrant VM, too). Closing this as it's unnecessary now :-)

PR #1806, opened 2021-01-06
inv: Correct Docker marist s390x node name

Willsparker:
ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1716#issuecomment-755202080

Just some mistakes I noticed whilst going through the list of machines that need to be added to Nagios. As these are inconsistent with the node names in Jenkins, my Nagios check (#1717) is unable to find them. I opted for changing the inventory entry as this is easier than renaming the node in Jenkins :-)

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -96,7 +96,7 @@ hosts:
           ubuntu1604-x64-1: {ip: 160.153.235.114, user: adoptopenjdk}
 
       - marist:
-          ubuntu1603-s390x-1: {ip: 148.100.113.58, user: ubuntu}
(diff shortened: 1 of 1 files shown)


sxa (changes requested):
this change to add the `l` onto a subset of the arm systems makes the naming inconsistent as many are still left as `armv8` / `armv7`, so I would suggest we need to change the definitions elsewhere.

Willsparker:
I'll take that out - we should probably change it in Jenkins then :-)

karianna (approved):
lgtm

sxa (approved):

Issue #1858, opened 2021-01-21
Vagrant: Windows Vagrantfile can't finish provision script


Willsparker:
Seen both locally and here: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/992/

`wget` is failing in the provisioning script, therefore it can't retrieve `ConfigureRemotingForAnsible.ps1`, and therefore Ansible can't connect. The error message is as shown: 
```
powershell.exe : wget : The request was aborted: Could not create SSL/TLS secure channel.
```
This has been seen before on Windows machines, but wasn't resolved: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/336#issue-312056271

Willsparker:
This is also the case for `https://cygwin.com/setup-x86_64.exe`, so it's the Window VM's issue, not the URL

Willsparker:
Forcing Windows to use TLS1.2 seems to have fixed the issue. Not sure why this has only recently become an issue. PR incoming

Willsparker:
For background: Most websites use `tls1.2` as security. Windows, being Windows, still supports `tls 1.0` and `1.1`, even though it broadly isn't used anymore. For some reason, this isn't the default when running `wget` or `Invoke-WebRequest` commands, and so the error message above occurs. The command `[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12` forces Windows to use `tls 1.2`, therefore fixing the issue- however, this doesn't persist through a new instance of Powershell, so it may be best to make this part of the PS profiles, via Ansible.
i.e. Ansibilising this process: https://www.howtogeek.com/50236/customizing-your-powershell-profile/

PR #1860, opened 2021-01-22
vagrant: Force win2012r2 to use tls1.2 in provision script

Willsparker:
Fixes: #1858 

Also updates the comments to better reflect what's happening in the VF

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access). In progress: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/993/

### DIFF
--- a/ansible/Vagrantfile.Win2012
+++ b/ansible/Vagrantfile.Win2012
@@ -1,14 +1,22 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Runs powershell as an administator and gets/executes an Ansible provided script that configures WinRM to allow Ansible to communicate over it. Then places a file in the shared folder of the VM that contains the IP address of the VM.
+# Runs Powershell as an administator and does the following:
+#  - Gets/executes an Ansible provided script that configures WinRM to allow Ansible to communicate over it.
+#  - Resizes the disk to ~100GB, in line with the 'disksize.size = 100GB' option in the config below
+
 $script = <<SCRIPT
 Start-Process powershell -Verb runAs
+
+# Windows 2012r2 needs to be forced to use TLS1.2, see: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1858
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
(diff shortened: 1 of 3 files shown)


Willsparker:
VPC run was able to boot the VM with the provision script as it's meant to, and started the playbook as usual.

sxa (approved):
Interesting ... Do we know what change caused this to be a problem? Was it after of the ansible reinstallations or has there been a change in the virtualbox image?

And therefore under what circumstances might it be seen outside a vagrant environment (i.e. do we need to document it somewhere?)

Willsparker:
I'm unsure of the cause - the problem occurs when I run a Windows Vagrant VM on my laptop too, so I don't think it's the Ansible re-installation. From the looks of https://app.vagrantup.com/mwrock/boxes/Windows2012R2 , the image hasn't been changed either. 
A quick google doesn't show any other people having this issue in the past month either. 

As for effects outside a Vagrant environment - I just tried `Invoke-WebRequest` on both `test-azure-win2012r2-x64-1` and `build-alibaba-win2012r2-x64-1` and they have the same issue, so yeah, I'll document it somewhere, as part of this PR.

sxa:
This makes me nervous because it doesn't appear to be reproducible from @Haroon-Khel's ansible client (also macos)

Willsparker:
This particular fix has nothing to do with Ansible - this fix is so the provisioning script at the beginning of the Windows Vagrantfile can be run - and the command isn't persistent after the shell it's in has been closed, so it won't affect the playbooks

sxa:
LGTM (other than whatever has caused the CentOS check to fail) - and thanks for including the doc updates too :-)

I feel we still need a story around the `7-Zip` thing. Would be good to know if that is related to whether cygwin is on the box

Willsparker:
Okay, merging. 

This [Windows VPC run](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/996/) succeeded, without any issues, so it may have just been a coincidence that `7-zip` failed? Not sure, but I'll keep an eye on it.

Ref: C6 check, a `tar.xz` has been removed from https://ci.adoptopenjdk.net/userContent/gcc/ ; asked about it [in Slack](https://adoptopenjdk.slack.com/archives/C53GHCXL4/p1611733179055400)

Issue #1850, opened 2021-01-18
Allow AWX to locate private data (Vendor_FIles/secrets) required for setup


sxa:
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [ ] Missing install
- [ ] Bug in ansible playbook
- [x] Request for new playbook addition

Details:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1560 prevented AWX from installing using the `/Vendor_Files` directory and stops anyone who doesn't have access to the secrets repo from being able to set up a machine at adoptopenjdk. This is currently (a) undocumented and (b) Not so easy to automate securely.

My gut feel is that we should allow either the secrets repo or the `/Vendor_Files` location to be used. We could also not have the public keys requiring GPG encryption.

Willsparker:
@gdams , could you weigh in on this? The PR @sxa linked is the one we implemented.

Regarding the automation - a very hacky workaround could be to setup the AWX server with a dummy github account that has admin access. Also, whats the issue of having `dotgpg` on AWX? 
The alternative, I suppose, is set it up so it skips `get_vendor_files` if it detects the Vendor Files.. I'll look at doing that.

Willsparker:
Note: For this, I'm assuming that the vendor files at `/Vendor_Files` is going to be in the same format as the Vendor Files within the secrets repo, just not gpg encrypted. This will allow me to do a similar method to what this (and the task afterwards) does: https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/92bb453988ca77e13250883c35910ba4fcfe236b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Get_Vendor_Files/tasks/main.yml#L16

i.e. https://github.com/AdoptOpenJDK/secrets/blob/master/vendor_files/Jenkins_User_SSHKey.gpg in the local directory would be `/Vendor_Files/Jenkins_User_SSHKey`

PR #1870, opened 2021-01-28
unix/winPB: Allow use of local Vendor_Files folder

Willsparker:
Fixes: #1850 

In cases where `dotgpg` can't be used, or we need to automate running the playbooks, (such as AWX, see the referenced issue), currently, there's no way to run the `adoptopenjdk` tagged roles (or any roles requiring vendor file material). To fix this, I've adapted the `Get_Vendor_Files` role to allow  the use of the local vendor files, or retrieving them from the secrets repo in the case that the machine doesn't have local vendor files.

This assumes that the local vendor files:
1) Follow the same structure as the vendor_files in the secrets repo (i.e. https://github.com/AdoptOpenJDK/secrets/blob/master/vendor_files/Nagios_User_SSHKey.gpg --> /Vendor_Files/Nagios_User_SSHKey)
2) aren't gpg encrypted.

##### Checklist

- [x] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md#commit-messages) (Sort of)
- [ ] FAQ.md updated if appropriate : N/A
- [ ] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) :  Not beneficial as the `adoptopenjdk` is skipped in `VPC`. 
- [ ] inventory changes, ensure bastillion is updated accordingly: N/A

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -21,7 +21,7 @@
   roles:
     - Debug
     - role: Get_Vendor_Files
-      tags: [vendor_files, adoptopenjdk]
+      tags: [vendor_files, adoptopenjdk, jenkins_user, nagios_plugins, superuser]
     - Version
     - Common
     - Providers                   # AdoptOpenJDK Infrastructure
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Get_Vendor_Files/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Get_Vendor_Files/tasks/main.yml
@@ -3,24 +3,54 @@
 # VENDOR_FILES
 ##############
 
-- name: check out AdoptOpenJDK/secrets
+- name: Check if local Vendor_Files exists
+  stat: path=/Vendor_Files
+  register: local_vendor_files
+  delegate_to: localhost
+  run_once: true
+
+- name: Check out AdoptOpenJDK/secrets
   git: repo=git@github.com:AdoptOpenJDK/secrets.git dest=vendor_files force=true
   delegate_to: localhost
+  when:
+    - not local_vendor_files.stat.exists
   run_once: true
 
-- name: check if dotgpg is installed
+- name: Check if dotgpg is installed
   shell: command -v dotgpg
   delegate_to: localhost
+  when:
+    - not local_vendor_files.stat.exists
   run_once: true
 
-- name: generate list of vendor_files
+- name: Generate list of remote vendor_files
   command: dotgpg cat {{ item }}
   delegate_to: localhost
-  register: vendor_files
+  register: remote_vendor_files_list
+  when:
+    - not local_vendor_files.stat.exists
   loop: "{{ lookup('fileglob', 'vendor_files/vendor_files/*', wantlist=True) }}"
   run_once: true
 
-- name: set facts from output of dotgpg
+- name: Generate list of local vendor_files
+  command: cat {{ item }}
+  delegate_to: localhost
+  register: local_vendor_files_list
+  when:
+    - local_vendor_files.stat.exists
+  loop: "{{ lookup('fileglob', '/Vendor_Files/*', wantlist=True) }}"
+  run_once: true
+
+- name: Set facts from output of dotgpg
   set_fact: {"{{ item.item | basename | replace('.gpg','') | replace('.','_') }}":"{{ item.stdout }}"}
-  with_items: "{{ vendor_files.results }}"
+  with_items: "{{ remote_vendor_files_list.results }}"
+  when:
+    - not local_vendor_files.stat.exists
+  run_once: true
+
+- name: Set facts from output of local vendor_files
+  set_fact: {"{{ item.item | basename | replace('.','_') }}":"{{ item.stdout }}"}
+  with_items: "{{ local_vendor_files_list.results }}"
+  when:
+    - local_vendor_files.stat.exists
   run_once: true
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
@@ -35,7 +35,7 @@
   roles:
     - Debug
     - role: Get_Vendor_Files
-      tags: [vendor_files, adoptopenjdk]
+      tags: [vendor_files, adoptopenjdk, jenkins]
     - Version
     - Common
     - Windows_Updates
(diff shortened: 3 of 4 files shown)


Willsparker:
Ref: Latest Commit.

I'm unsure about how we want to proceed with the AIX playbook - I'm happy to change it so it's in line with Windows/Unix, if that's whats wanted. The latest commit just updates where the SSH-keys are expected to be, so they're in line with where the Windows/Unix playbooks expect them to be.

sxa (commented):
Slightly odd that we're getting this
```
[WARNING]: Unable to find 'vendor_files/vendor_files' in expected paths (use
-vvvvv to see paths)
```
in the output along with the `Check if dotgpg is installed` section - that path is referenced in the subsequent `Generate list of remote vendor_files` but that hsould be skipped when `/Vendor_Files` exists ...

I'll do some extra testing today but so far this is at least looking more promising than it was before - thanks :-)

sxa:
> I'm unsure about how we want to proceed with the AIX playboo

I think we need to do that, otherwise things just get messy, but happy for it to be done in a separate PR to this ...

sxa:
This isn't working for me at the moment - it's not correctly picking setting the authorized_keys file. The original version (SHA `f4403a07798cd285de12ac8a19c5861ca78e415e`) works ok

sxa:
This doesn't work as-is because the user roles have e.g.
```
     key: "{{ Jenkins_User_SSHKey }}"
```
instead of
```
      key: "{{ lookup('file', '{{ Jenkins_User_SSHKey }}') }}"
```

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/group_vars/all/main.yml
```diff
@@ -15,12 +15,12 @@ Sudoers_File: /etc/sudoers
 
 # Jenkins User Variables:
 Jenkins_Username: jenkins
-Jenkins_User_SSHKey: /Vendor_Files/keys/id_rsa.pub
+Jenkins_User_SSHKey: /Vendor_Files/Jenkins_User_SSHKey
```
sxa, line 18: I don't think these lines are going to achieve the desired effect (sadly) - the `_SSHKey` values now ned to be the contents of the file, not the filename itself

sxa (changes requested):
In addition to the comment above, the reason things weren't being picked up (as per my previous comment) was because this currently doesn't pick up the values properly if you are only executing a specific role e.g. running the playbook with `-t jenkins_user` so perhaps we need to ensur etha the `Get_Vendor_Files` tasks are all tagged with all the roles that make use of them?

sxa:
Just tried with a modified version of `Get_Vendor_Files` that included the `jenkins_user` and `nagios_plugins` tags on each task, and move my files around from the old locations (I'm not sure I like the new names ad structure though) and it works, so we should identify the list of tags we need to catch these situations and add them to complete this.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/group_vars/all/main.yml
Willsparker: Ah okay - I'll revert those, and I'll make a separate PR for AIX :-)

Willsparker:
With the latest commit, I think the tags I've added cover all the roles that need the Vendor_Files - `jenkins_user, nagios_plugins, superuser` in Unix's case & `jenkins` in Window's case

sxa (approved):

sxa:
> With the latest commit, I think the tags I've added cover all the roles that need the Vendor_Files - `jenkins_user, nagios_plugins, superuser` in Unix's case & `jenkins` in Window's case

This is now lookgin promising and appears to work in myt scenario - I presume you've tested that it still successfully picks things up from the `secrets` repo if required ;-)

Hmmm just had another thought after looking at it - if I was to run it with `--skip-tags nagios_plugins` would that stop it locating the jenkins_user key?

Willsparker:
Sadly, it does :| 
Maybe it'd be better to just keep the `adoptopenjdk` tag on it, and put in the README.md something along the lines of "If you want to run the playbook with any of the following tags: `jenkins, nagios_plugins, superuser`, then the `vendor_files` tag must be included"
It's not the best solution, but I would assume that the people who have access to the Vendor Files would read the README? 

Alternatively, I could put the tasks that set all the variables from the local Vendor Files, into the `Common/main.yml` role, but I'd say that feels like a worst solution ...

aixtools:
On something semi-related:
a) I like hands-free incl - report errors/warnings as you find them - but continue as long as possible
b) while developing a change to jenkins_user for AIX - and not having access to the secret I came up with the following logic

1. register: jenkins_user_rsa (or jenkins_user_pki, or whatever you like - different from the tag in any case).
2. ignore_errors: yes
3. Add new debug task to report key could not be copied when: jenkins_user_rsa.failed == true

What I did not add was a task to test if a key is already installed (after the failed). With an extra condition such as this the message could be changed to something as: key was not not updated, if required; or when nothing is there "jenkins_user" is not functional - key is missing (and perhaps have the playbook stop, although my preference is to continue and see what else may be missing/needed).

The names of the variables don't interest me. However, imho, having a playbook that plays to the end - reporting everything it finds is more efficient that re-running a playbook and have it stop at the first missing bit it runs into. With a list of missing items - I can do the searching, updating, and then re-running the playbook with the expectation all should pass this time.

sxa:
I think those points in the last comment are separate from this PR and should not hold up merging of it - @Willsparker I'm happy for this to go in now if you are (It's actively causing a problem not having it in) and other enhancements to cover the scenaios I described can be done 

Ref our other discussion @aixtools I do think it's worth having a `vendor_files` tag in the AIX ones too in order to avoid your "seven different skip tags" :-)

@Willsparker Do you know if the playbooks run ok if neither secrets or `/Vendor_Files` is accessible and `--skip-tags vendor_files` is used?

Willsparker:
If the `vendor_files` tags are skipped, the roles that use the vars from them will need to be skipped too. Doing `--skip-tags adoptopenjdk,jenkins` will do that :-) 

Yeah, I'm happy for this to go in

Willsparker:
@aixtools While I agree that it's frustrating to have to re-run the playbook whenever it fails (and having to re-run it many times to fix all the issues), I don't think it's a good idea to force them to succeed if they don't work. I think this is just asking for issues to go unnoticed when using the playbooks to setup actual machines, from people not looking through the Ansible logs - I know I don't tend to look through the logs if the Ansible reports the playbook as succeeding.

sxa:
> If the `vendor_files` tags are skipped, the roles that use the vars from them will need to be skipped too. Doing `--skip-tags adoptopenjdk,jenkins` will do that :-)

Hmmmm I've just had a situation where I've tried to do a "basic" machine setup, so I thought I'd just run the `adoptopenjdk` role. Needless to say, this tries to create the jenkins user but then skips the reading of the key, so it falls over. Will need to think about a solution for that.

Willsparker:
Really? Ansible should run that role, If you specify `--tag adoptopenjdk` ... :thinking:

Issue #1803, opened 2021-01-05
build-osuosl-aix71-ppc64-1: tests which specify -Xgc:metronome fail (openj9 option)


lumpfish:
Tests which use the openj9 command line option `-Xgc:metronome` fail when run on `build-osuosl-aix71-ppc64-1` but pass when run on `test-ibm-aix71-ppc64-2`.

The symptoms are the same as those described in https://github.com/eclipse/openj9/issues/10579, where the underlying reason was said to be "AIX machine is not configured for High Resolution Clock to support Metronome".

aixtools:
The issues discussed in the note above have been addressed in the AIX playbook. Apparently the changes were never applied manually to the osuosl systems and/or the playbooks were not re-run to update user settings.

karianna:
@aixtools Safe to close this now in that case?

sxa:
@lumpfish Are you able to verify whether this now works? It's clear from this issue which test showed the problem

aixtools:
Likewise, as I mentioned in #1816 (read asked) the possibility of just rebuilding `build-osuosl-aix71-ppc64-1`, and perhaps later `build-osuosl-aix71-ppc64-2` - rather than trying to dig in a system that has been manually adjusted several times. Especially if `test-osuosl-aix71-ppc64-1` seems to be working correctly - which was recently built using the latest playbook(s).

lumpfish:
The tests still fail when run on build-osuosl-aix71-ppc64-1: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6081/consoleFull

This will rerun on of the tests affected:  https://ci.adoptopenjdk.net/job/Grinder/parambuild/?JDK_VERSION=8&JDK_IMPL=openj9&JDK_VENDOR=adoptopenjdk&BUILD_LIST=functional&PLATFORM=ppc64_aix&TARGET=cmdLineTester_jvmtitests_5

aixtools:
I looked at the changes listed in https://github.com/eclipse/openj9/issues/10579#issuecomment-742627758 - and I had already added the capabilities to jenkins.

What I notice is still different - root was not the administrator of the group staff (should not influence capabilities), but the shell is still `ksh` and not `bash`.

Further, while `/home/jenkins/.profile` does not exist - there is nothing (simple) to find to describe what variables are being set by /etc/environment and/or /etc/profile.

for the record:

```
[...]

PR #1874, opened 2021-01-31
aixPB: Add RBAC configuration to enable RealTime Clock usage by user, e.g., jenkins

aixtools:
##### Checklist

Fixes: #1803 - shows that AIX versions earlier than AIX 7.1 TL5 need RBAC support to enable a non-root account to access the realtime clock needed by `-Xgcpolicy:metronome` These RBAC settings enable user accounts to access the AIX realtime clock using a role named `rtclk`.


ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/jenkins_user/tasks/main.yml
```diff
@@ -8,62 +8,77 @@
...
-  ignore_errors: yes
-  when: jenkins.stat.isdir is not defined
-  tags: jenkins_user
+  block:
+    - name: Create jenkins user
+      command: mkuser home="/home/{{ Jenkins_Username }}" shell="/usr/bin/ksh" {{ Jenkins_Username }}
+      ignore_errors: yes
+      tags: jenkins_user
 
-- name: Create SSH Key folder for {{ Jenkins_Username }}
```
sxa, line 22: Doesn't need to be changed now but we should look at doing this in the same we we do on the other UNIX platforms: https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Jenkins_User/tasks/main.yml#L69

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/rbac/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+#########################################################################
+# Playbook to create authorization to access realtime clock privelidge
+# The role is assigned to the jenkins user
```
sxa, line 3: Given that this is quite unusual, can we add a link in the comments to the issue regarding this problem, so that people can understand why we added it.

sxa (commented):
I trust that you've got the syntax right, although was this PR supposed to have the `jenkins` user changes in it?

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/rbac/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+#########################################################################
+# Playbook to create authorization to access realtime clock privelidge
```
adamfarley, line 2: Typo change: privelidge = privilege

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/jenkins_user/tasks/main.yml
aixtools: I would like to make more changes - good discussion for a different PR.

sxa (approved):
Since these changes have been verified on a test machine I'm happy with this. When deploying to the build machines, let's ensure that we deploy one at a time and ensure there are no side effects before running it on other systems.

Issue #1901, opened 2021-02-05
Ansible request for Ubuntu 20.04/aarch64 OS detection for Nagio_Plugins role


sxa:
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [ ] Missing install
- [x] Bug in ansible playbook
- [ ] Request for new playbook addition

Details: Not quite sure why, but when attempting to deploy the playbook on a Ubuntu 20.04/aarch64 machine the `ansible_distribution_major_version` is set to `***Undefined***` (as is `ansible_os_family`) which means that [this guard](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/2925d55ff93ca8bbbbaf6ada6184b7aa18847a69/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml#L30) in the `Nagios_Plugins` role causes the playbook to fail. Easily reproducible by running with `--tags debug`

Willsparker:
Happy to look at this. But surely `ansible_distribution_major_version` being defined is going to cause a lot more issues than just Nagios, and we should probably figure out why it's doing that. Maybe a platform to add on QPC :thinking:

Willsparker:
Closing due to #1912

PR #1912, opened 2021-02-09
unixPB: Fix U20 condition in Nagios_Plugins

Willsparker:
Ref: #1901 

The condition wasn't correctly skipping the role, which probably wasn't helping the referenced issue. @sxa , could you test this to see if this fixes the issue - If your `ansible_distribution_major_version` variable really was undefined, there would presumably be failures much sooner in the playbook than Nagios.

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
@@ -27,7 +27,7 @@
     state: latest
     update_cache: yes
     pkg: nagios-plugins-common
-  when: (ansible_distribution_major_version != 20)
+  when: not (ansible_distribution_major_version == "20")
 
 ##########
 # Layout #
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
@@ -27,7 +27,7 @@
     state: latest
     update_cache: yes
     pkg: nagios-plugins-common
-  when: not (ansible_distribution_major_version == "20")
+  when: (ansible_distribution_version is version_compare('20', operator='ge'))
 
 ##########
 # Layout #


sxa:
> The condition wasn't correctly skipping the role, which probably wasn't helping the referenced issue. @sxa , could you test this to see if this fixes the issue - If your `ansible_distribution_major_version` variable really was undefined, there would presumably be failures much sooner in the playbook than Nagios.

Nope - that was the first failure :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
```diff
@@ -27,7 +27,7 @@
     state: latest
     update_cache: yes
     pkg: nagios-plugins-common
-  when: (ansible_distribution_major_version != 20)
+  when: not (ansible_distribution_major_version == "20")
```
sxa, line 30: Could we use something like this instead? Presumably the conditional will be valid for future versions >20 too which this won't catch?


```suggestion
  when: (ansible_distribution_version|version_compare('20.04', '>='))
```

sxa (changes requested):

sxa:
Related observation: We should check whether https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/00a3d5ab4b3742f722f5606a5ff43e6ab8ee02d6/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/CentOS.yml#L15 is doing what it should do ...

Willsparker:
I believe that's okay, as it has the speech marks around the "8", but I'll double check :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
Willsparker: Put in the latest commit, your syntax was a little off :-)

karianna (approved):
LGTM

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
sxa: Doesn't surprise me - came off the top of my head and I trusted you to adjust it accordingly ;-) Thanks

sxa (approved):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/nagios_Ubuntu.yml
Willsparker: :joy: You're welcome

Issue #1915, opened 2021-02-09
Ansible request for locales on RHEL77/s390x


sxa:
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [ ] Missing install
- [x] Bug in ansible playbook
- [ ] Request for new playbook addition

Details:
Tried to run playbooks on the bulid*rhel* systems at marist and got this:

```
TASK [Common : Install 'locales' package] **************************************
fatal: [build-marist-rhel77-s390x-1]: FAILED! => {"changed": false, "msg": "No package matching 'locales' found available, installed or updated", "rc": 126, "results": ["No package matching 'locales' found available, installed or updated"]}
fatal: [build-marist-rhel77-s390x-2]: FAILED! => {"changed": false, "msg": "No package matching 'locales' found available, installed or updated", "rc": 126, "results": ["No package matching 'locales' found available, installed or updated"]}
```

Willsparker:
I'm going to test things on `build-marist-rhel77-s390x-1` - for `rhel7`, `localectl list-locales` can be used to get the list of locales, however, I'm unsure if the `locale_gen` task will work. The [ansible module](https://docs.ansible.com/ansible/2.5/modules/locale_gen_module.html) page doesn't provide much information about how it works - however the `/etc/locale.gen` file isn't on the systems, so I assume we'll need to go a different route for RHEL7

PR #1917, opened 2021-02-10
unixPB: Find Locale-list based on RHEL major_version

Willsparker:
Fixes: #1915 

Sets the locale list command based on the `ansible_major_distribution_version`. The `if` statement is required because ansible will register variables even if a task is skipped - otherwise I would have had 2 separate tasks that register to `localeList`, as that'd be easier to read.

- [ ] [FAQ.md](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md) updated if appropriate : N/A
- [ ] other documentation is changed or added (if applicable) : N/A
- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) : Not required as we don't test the RHEL branch in VPC
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly : N/A

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
@@ -199,17 +199,20 @@
     name: locales
     state: present
   when:
-    - not (ansible_distribution_major_version == "7")
+    - (ansible_distribution_major_version is version_compare('7', operator='lt'))
   tags: locales
 
-- name: Set locale list command
-  set_fact:
-    locale_list_command: "{% if ansible_distribution_major_version=='7' %}localectl list-locales{% else %}locale -a{% endif %}"
+- name: Install 'glibc-common' package
+  package:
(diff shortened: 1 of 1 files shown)


sxa (commented):
I'm nervous about the `7` check. Seems odd if 6 and 8 would have the same, but the one in the middle are not.

sxa:
As far as I can see the `locale` command is available on RHEL7 and RHEL8 from the `glibc-common` package which is already in the playbooks - could this be an ordering issue?

Willsparker:
Ah okay, I was under the impression came from the `locale` package, which isn't available on 7. I'll give that a test :-)

Willsparker:
I had a check on `build-marist-rhel77-s390x-1`, it already had `glibc-common` installed and the `locale` command does exist :facepalm: . My mistake

karianna (approved):
LGTM assuming linter will get resolved down the line

sxa (approved):

PR #1925, opened 2021-02-12
Implement dockerhost.yml playbook for systems running static docker containers for test

sxa:
##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access): [VPC](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1025/) [VPC](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1033/)

This adds in a new playbook for creating a "Docker-only" host which holds statically built containers used to run testing (Formalises what was prototyped in issue [1809](https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1809)). It contains:
- New "DockerStatic" role to create and start one container on each of Ubuntu 16.04/18.04/20.04/20.10, Fedora 33 and CentOS 8
- New `adopt_etc` role which contains the items from Common which make changes to the system `hostname` etc. (Since we don't want all of the `Common` things in the `dockerhost.yml`
- Move of the definitions of the `crontab_patching` things from Common to the `Crontab` role so that it doesn't require `Common` to be executed first (makes the role self-sufficient)

To be addressed (potentially later - these would be good "introductory" things for new people to resolve):
- CentOS8 Dockerfile does not yet work so is excluded from the full process
- [Core file creation needs permanent fix](https://github.com/AdoptOpenJDK/openjdk-tests/issues/2138#issuecomment-755396683) - probably in the new `adopt_etc` role. Related: https://github.com/AdoptOpenJDK/openjdk-tests/issues/791  https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1829#issuecomment-762880860
- Locale-related failures need a solution (https://github.com/AdoptOpenJDK/run-aqa/issues/59, https://github.com/eclipse/openj9/issues/11620 closed, but https://github.com/eclipse/openj9/issues/11915 still outstanding)
- Switch download of java (for Jenkins agent) to V3 API and potentially move the JDK to 11 from 8
- Potentially have an optional "force rebuild/restart" option (Not default as it may be in use!)


sxa:
Not quite sure yet what it's complaining about - ansible-lint does not object in the same way when run on my local machine.

sxa:
With updates to the core file settings this fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1817

sxa:
NOTE TO REVIEWERS:
- Does not contain NTP_TIME role because the requisite packages are not installed since we don't run `Common` at present
- Nagios plugins do not work on the hosts I'm using. That can be resolved laster
- Because of the way I've put the jenkins ssh key in, I can't also have the bastillion ssh key along with it as well from Vendor_Files, so this breaks the ability for bastillion to administer up the jenkins user (Can fix separately)

Willsparker (approved):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
```diff
@@ -0,0 +1,26 @@
...
+RUN rm /tmp/jdk8.tar.gz /tmp/ant.zip /tmp/ant-contrib.tgz
+# Set up jenkins user
+RUN useradd -m -d /home/jenkins jenkins
+RUN mkdir /home/jenkins/.ssh
+RUN echo "Jenkins_User_SSHKey" > /home/jenkins/.ssh/authorized_keys
+RUN chown -R jenkins /home/jenkins/.ssh
+RUN chmod -R og-rwx /home/jenkins/.ssh
+RUN service ssh start
+CMD ["/usr/sbin/sshd","-D"]
+RUN apt-get install -qq -y git curl make gcc xvfb libxrender1 libxi6 libxtst6
```
aahlenst, line 23: Not having `apt-get update` here might cause problems if the layer from line 3 `RUN apt-get update && ...` is re-used but this one isn't.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
```diff
@@ -0,0 +1,26 @@
...
+# Install ant
+RUN wget -q -O /tmp/ant.zip 'https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.5-bin.zip'
+RUN wget -q -O /tmp/ant-contrib.tgz  https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.tar.gz
+RUN ln -s /usr/local/apache-ant-1.10.5/bin/ant /usr/bin/ant
+RUN unzip -q -d /usr/local /tmp/ant.zip
+RUN tar xpfz /tmp/ant-contrib.tgz -C /usr/local/apache-ant-1.10.5/lib --strip-components=2 ant-contrib/lib/ant-contrib.jar
+# Clear up space
+RUN rm /tmp/jdk8.tar.gz /tmp/ant.zip /tmp/ant-contrib.tgz
+# Set up jenkins user
+RUN useradd -m -d /home/jenkins jenkins
```
aahlenst, line 16: We should predefine a UID/GID that does ideally not collide with a host UID/GID. One popular recommendation is to use `10000`.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
```diff
@@ -0,0 +1,26 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -qq -y perl openssh-server unzip
+# Get java8
+RUN wget -q 'https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=hotspot&os=linux&arch=x64&release=latest&type=jdk' -O /tmp/jdk8.tar.gz
+RUN mkdir -p /usr/lib/jvm/jdk8 && tar -xpzf /tmp/jdk8.tar.gz -C /usr/lib/jvm/jdk8 --strip-components=1
+# Install ant
+RUN wget -q -O /tmp/ant.zip 'https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.5-bin.zip'
+RUN wget -q -O /tmp/ant-contrib.tgz  https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.tar.gz
```
aahlenst, line 9: What about checksum checks?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
```diff
@@ -0,0 +1,26 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -qq -y perl openssh-server unzip
```
aahlenst, line 3: `ENV DEBIAN_FRONTEND=noninteractive`?

aahlenst (commented):
My comments regarding the Ubuntu Docker file apply to all to Ubuntu/Debian Docker files.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
sxa: Since these containers are fully isolated and not sharing the file system with the host (unlike the build ones) that shouldn't be a concern for these.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
sxa: Agreed ... Fixed.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
sxa: Could do, although it's a bit more of a pain to do that within a dockerfile (also since this is for testing rather than building it's not such a big risk to our build production cycle) - suggest if we want it we do it under a separate PR :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
sxa: Done

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
aahlenst: According to my understanding, you can still escape from the container and end up with the same UID on the host as in the container as long as Docker user namespacing isn't active and configured. If I read the Dockerfile correctly, sshd is run as root, anyway, so it might make sense to tackle this in a separate PR if we really need sshd.

aahlenst (approved):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.u1604
aahlenst: Filed #1954.

PR #1933, opened 2021-02-15
VPC: add MBCS test for vagrant playbook check job

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1927#issuecomment-778289501

I have added the MBCS_Tests_pref_ja_JP_linux_0 test as a quick test to ensure that the required perl modules are on a machine after the playbook runs on them, since installing perl from yum seems to sometimes omit certain perl modules.

I was initially going to add this test as something optional, but ive instead replaced the existing MachineInfo test altogether as the MachineInfo test's only purpose is to see if a test _can_ run

What do you think @Willsparker?

### DIFF
--- a/ansible/pbTestScripts/testJDK.sh
+++ b/ansible/pbTestScripts/testJDK.sh
@@ -13,6 +13,7 @@ mkdir -p $HOME/testLocation
 [ ! -d $HOME/testLocation/openjdk-tests ] && git clone https://github.com/adoptopenjdk/openjdk-tests $HOME/testLocation/openjdk-tests
 $HOME/testLocation/openjdk-tests/get.sh -t $HOME/testLocation/openjdk-tests
 cd $HOME/testLocation/openjdk-tests/TKG || exit 1
-export BUILD_LIST=system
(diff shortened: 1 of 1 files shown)


sxa:
Kicked off in VPC at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1038 (That's going to take a while as it's queued behind several others)

Willsparker (approved):
LGTM, on the basis that VPC passed. It's always nice to have _useful_ tests running :grin:

sxa:
I've got a feeling the system test ones can show up other prereq problems, so MachineInfo_0 is also useful. Perhaps we should leave both in, although I can't remember specifically what prereqs are specific to the system tests. @lumpfish Do you know what prereqs on the machines are specific to the system test suites (maybe just in the compile phase)?

lumpfish:
The system test build should get everything the tests need except for ant (to start the build), wget / curl (to get the prereqs), gnu make (to run the build).

Issue #1945, opened 2021-02-17
Ansible request for Windows shortnames


sxa:
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [ ] Missing install
- [x] Bug in ansible playbook
- [ ] Request for new playbook addition

Details: This is showing up during the `[Common : Ensure 'Program Files (x86)' has a shortname]` task in the Windows playbook on at least some machines, preventing successful completion:

`fatal: [test-ibmcloud-win2012r2-x64-2]: FAILED! => {"changed": true, "cmd": "fsutil file setshortname \"C:\\Program Files (x86)\" \"PROGRA~2\"", "delta": "0:00:00.119001", "end": "2021-02-17 01:23:29.342042", "msg": "non-zero return code", "rc": 1, "start": "2021-02-17 01:23:29.223041", "stderr": "", "stderr_lines": [], "stdout": "Error:  Access is denied.\r\n\r\n", "stdout_lines": ["Error:  Access is denied.", ""]}`

Willsparker:
This can't be run after we have installed the rest of the playbook, because services start running in the `Program Files (x86)` folder. 
I'll put the `Ignore_Errors: yes` var into it, as this should allow the playbook to be re-run on already setup machines.

PR #1953, opened 2021-02-19
winPB: Ignore Program Files (x86) shortname errors

Willsparker:
Fixes: #1945 

Ignores the errors when creating shortnames. Reasoning for this is that, if the playbook is run on a machine that's already setup, the task in question will fail as there will be processes running in the Program Files (x86) folder. This will occur, even if the folder already has it's shortname. 
If, on an initial playbook run, the task fails and is ignored, the problem will be evident when the machine is tested, and manual intervention would be required anyway. 

##### Checklist


sxa (approved):
I feel it would be better to check this and only attempt this if necessary, as it could otherwise cause follow-on failures, but since this is blocking deployments at the moment I'll let it thorugh

PR #1957, opened 2021-02-19
UnixPB: symlink /usr/bin/java to /usr/java8_64/jre/bin/java

Haroon-Khel:
The symlink should be pointing at /usr/java8_64/jre/bin/java not /usr/java8_64/, causing aix machines to use java6 as a default


karianna (approved):
lgtm

sxa (approved):

PR #1958, opened 2021-02-19
WindowsPB: Include cmake in cygwin package list

Haroon-Khel:
Ive removed the cmake role from the main.yml of the windows playbook since we use cygwin's cmake for openj9 on windows. Ive also included cmake as the list of packages that will get installed with cygwin.

Will test on one of the alibaba machines shortly


Willsparker (commented):
Given we're removing `cmake` from `main.yml`, should we remove the whole role? ping @sxa

sxa:
> Given we're removing `cmake` from `main.yml`, should we remove the whole role? ping @sxa

Since I don't think we call it from any other playbooks I'd be ok with removing it completely since it seems unlikely we would put it back in - I'm not aware of anything else that would still need it (Second opinion from @AdamBrousseau might be useful.

sxa:
> Will test on one of the alibaba machines shortly

How did the testing go?

Haroon-Khel:
Seemed to do the trick. On alibaba-2:
```
$ ls -la cygwin64/bin/ | grep cmake
-rwxr-xr-x    1 Administrator None  5211155 Aug 10  2020 ccmake.exe
-rwxr-xr-x    1 Administrator None  5622291 Aug 10  2020 cmake.exe

```
Which were definitely not there before. Running a jdk11 openj9 job on the machine
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-windows-x64-openj9/

Haroon-Khel:
Fails before getting to the cmake part, with this error
```
13:44:48  checking for OPENSSL... configure: error: Cannot locate the the path of OPENSSL_DIR
13:44:48  configure: The path of OPENSSL_DIR, which resolves as "/cygdrive/c/openjdk/OpenSSL-1.1.1j-x86_64-VS2017", is invalid.
```
Its never hit this before. Investigating

Haroon-Khel:
I see that the openssl version was recently changed from i to j https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1947
Ill rerun the openssl role on alibaba -2

Haroon-Khel:
Running again, https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-windows-x64-openj9/929/console

sxa:
> Running again, https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-windows-x64-openj9/929/console

Looks ok :-)

AdamBrousseau (approved):
We do call the roll when we run the playbooks internally but upon investigation we don't use and we shouldn't be installing it so I'm fine with removing it completely.

Willsparker (changes requested):
As per the above conversation, just remove the standalone cmake roles and I'll approve :+1:

Haroon-Khel:
@Willsparker Cmake role removed 👍🏻

Willsparker (approved):
Thanks!

Issue #1877, opened 2021-02-01
Ansible request for CentOS6 adoptopenjdk_install role


Willsparker:
Affecting both `VPC` and the CentOS6 GitHub Check: 
- https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/997/OS=CentOS6,label=vagrant/console
- https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1875/checks?check_run_id=1804893220

- [ ] Missing install
- [x] Bug in ansible playbook
- [ ] Request for new playbook addition

Details:

The issue is with `adoptopenjdk_install` : 
```
fatal: [localhost]: FAILED! => {"attempts": 3, "changed": false, "msg": "Failed to validate the SSL certificate for github-releases.githubusercontent.com:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine  (the python executable used (/usr/bin/python) is version: 2.6.6 (r266:84292, Jun 20 2019, 14:14:55) [GCC 4.4.7 20120313 (Red Hat 4.4.7-23)]) or you can install the `urllib3`, `pyOpenSSL`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible. The exception msg was: hostname 'github-releases.githubusercontent.com' doesn't match either of '*.githubassets.com', 'githubassets.com'.", "status": -1, "url": "https://api.adoptopenjdk.net/v3/binary/latest/8/ga/linux/x64/jdk/hotspot/normal/adoptopenjdk?project=jdk"}
```
Apparently it's failing to validate the SSL Certificate for github-releases.githubusercontent.com. If all else fails, we can use `validate_certs=False`, but this is a last resort.

Willsparker:
Replicated problem on local CentOS6 VM. Looking at doing as the Issue suggests, and updating Python to 2.7.9, however, we'll still have to ensure the GH check and the VPC machine has Python 2.7.9

Willsparker:
Hmm, installing `2.7.9` ruined my Ansible install.... Not something I want to do on the VPC machine

[...]

PR #1970, opened 2021-02-23
unixPB: Install Python2.7 on C6; Use it to install JDKs

Willsparker:
Fixes: #1877 

IT WORKS! (in VPC at least: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1067/ )

Installs Python 2.7 on CentOS6 and then uses that Python to run the `adoptopenjdk_install` tasks. See the referenced issue for what other things were tried or suggested. This solution also adds some future proofing somewhat, in that we can just use the Python2.7 whenever we have another `unarchive` / `get_url` cert issue because CentOS6 is old.

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) : See above

### DIFF
--- a/ansible/Dockerfile.CentOS6
+++ b/ansible/Dockerfile.CentOS6
@@ -14,6 +14,7 @@ RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo; \
     ./configure --enable-optimizations; \
     make install; \
     rm /usr/src/Python-3.6.10.tgz; \
+    pip3 install --upgrade pip; \
     pip3 install ansible
 
 COPY . /ansible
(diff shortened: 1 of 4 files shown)


sxa (commented):
How long does this role to execute? Could be preferable to use the pre-built one I put in place last week instead which would also avoid having to install the extra prereq packages on the machine.

karianna (changes requested):
Fix YAMLLint please :-)

Willsparker:
> How long does this role to execute? Could be preferable to use the pre-built one I put in place last week instead which would also avoid having to install the extra prereq packages on the machine.

About 10 minutes in VPC , not sure on the GHCheck as it doesn't have timestamps. But yeah, okay, I'll try prebuilt :-)

karianna (approved):

Willsparker:
Well it works! :tada:

gdams (approved):

PR #1993, opened 2021-03-02
inventory: add nine test macs

gdams:
##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)


sxa (commented):
Hmmm two things;
- Since these are primarily to reduce costs at other providers AIUI they shouldn't be targeted exclusively as performance machines, so we should call them `test-` otherwise we're not going to be using them to best effectiveness for the desired cost saving purpose
- I'm not keen on the provider type of `adoprium` ... While that tells you about the hardware owner it does not give an indication of who to contact/what the admin team needs to do if there is a failure of connectivity etc. (And also we may have multiple locations at some point and I don't want to add to confusion ... We haven't used it own name for the SPARC server hosting) So for that reasonI'd prefer e.g. `nine`

gdams:
@sxa agreed, updated PTAL

karianna (approved):
LGTM

PR #1988, opened 2021-03-01
AixPb: create hosts_file role for aix

Haroon-Khel:
Currently the aix playbook does not have a role/task which enters the correct entry in `/etc/hosts`, leading to improper entries added to this file; the correct entry being something that looks like

```
$public_ip $hostname.$domain $hostname

// For localhost
127.0.0.1   localhost  localhost.localdomain
```

So far this role is a direct copy of the `adopt_etc` role from the Unix playbook and is a draft. I will be making it suitable for aix, I just wanted to get the bulk of it down for now

PR #1996, opened 2021-03-03
UnixPb: change ansible_default_ipv4.address variable to ansible_host

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1994

On some machines (build-scaleway-ubuntu1604-armv7-1 and -2) ansible_default_ipv4.address isnt able to detect the public ip, but instead detects the private ip.

This causes problems with the `hosts_file` tagged tasks which enter lines in `/etc/hosts` based on the ansible_default_ipv4.address variable. Therefore I have changed this variable to ansible_ssh_host, which I have found better detects the public ipv4 of a machine. 

The line that ive added to the `Debug` role will be removed after I have adequately tested this pr

Tests list:

- [x] Suitable for scaleway machine test case
- [x] Does `ansible_host` detect ip or hostname?
- [x] Localhost test case
- [x] `inventory_hostname != ansible_default_ipv4.address` redundant?

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
@@ -3,15 +3,6 @@
 # Main #
 ########
 
-###########################
-# Overide amd64 to x86_64 #
-###########################
-- name: If ansible_architecture is amd64 set it to x86_64
-  set_fact:
-    ansible_architecture: "x86_64"
-  when:
-    - ansible_architecture == "amd64"
-
 ###############################
 # Set ansible_processor_vcpus #
 ###############################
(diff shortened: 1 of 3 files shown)


sxa (changes requested):
I believe `ansible_ssh_host` has been deprecated in favour of `ansible_host` - can you test whether that works ok too?
I suppose my only concern is whether that will be set to a hostname instead of an IP address, as the `/etc/hosts` entry must have an IP address at the start

sxa (commented):
What's the reason for the `inventory_hostname != ansible_ssh_host` conditions?

aahlenst:
Don't want to be the party crasher, but we have machines without a public IPv4 address. They have a public IPv6 address, however.

sxa:
> Don't want to be the party crasher, but we have machines without a public IPv4 address. They have a public IPv6 address, however.

That's something we'll have to fix separately and not directly related to the problem this PR is solving though, unless you've spotted something I haven't. In fact, this PR removes the direct reference to the `ipv4` variable.

Haroon-Khel:
> What's the reason for the inventory_hostname != ansible_ssh_host conditions?
@sxa In cases where a user is running the playbooks from their local machine onto a remote machine, ansible will use the remote machine's public ip address as the inventory_hostname (unless the remote machine can be accessed by its fqdn, but this isnt the case for our machines I think). In this case the condition `inventory_hostname != ansible_default_ipv4.address` will correctly prevent an improper entry in the hosts file (because those two variables equal the remote machine's public ip), something like 
```
$ipv4Address $ipv4Address.adoptopenjdk.net $ipv4Address
```
But in the case of these scaleway machines, the condition `inventory_hostname != ansible_default_ipv4.address` will not prevent this, since ansible_default_ipv4.address picks up the private ip of these machines

Haroon-Khel:
build-alibaba-ubuntu1804-armv8-1 is experiencing the same issue. ansible_default_ipv4.address picks up the machine's private ip instead of public. Hence this machine has a line for its private ip in /etc/hosts. Have fixed this

Haroon-Khel:
>I suppose my only concern is whether that will be set to a hostname instead of an IP address, as the /etc/hosts entry must have an IP address at the start

After testing on a handful of test and build machines, the variable `ansible_host` gives the remote machine's ip only, not hostname thankfully. Though this is when running the playbooks from my laptop. Im not sure if there will be a difference when running from awx

Haroon-Khel:
I noticed that if you keep the `hosts` parameter in the top level `main.yml` to be
`hosts: "{{ Groups | default('localhost:build:test:perf:!*zos*:!*win*:!*aix*') }}"`
And if you dont supply a hostname on the command line, then ansible will run the playbook on the local machine (that is to make changes to localhost). In this case, `ansible_host` becomes `127.0.0.1`, and the adopt_etc role will make an improper entry in /etc/hosts. Therefor the condition 
`ansible_host != "127.0.0.1"` 
Needs to be added

aixtools:
This should not be limited to unixPB: - imho - we are speaking convention and it should be applied to all playbook families, e.g., in the PR currently open #1988

Haroon-Khel:
I think that should be all the changes needed. Ive kept the ansible_host variable in the Debug role. It may be redundant, but it wouldnt hurt to have it there

sxa (approved):

sxa:
This is probably ok now with the change in variable, but I want to review properly tomorrow when I feel more awake, therefore dismissing my previous review for now :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
```diff
@@ -3,15 +3,6 @@
 # Main #
 ########
 
-###########################
-# Overide amd64 to x86_64 #
-###########################
-- name: If ansible_architecture is amd64 set it to x86_64
```
Willsparker, line 9: Why is this task being removed? Our playbooks are set to detect the `x86_64` architecture, not `amd64`.

Willsparker (changes requested):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/main.yml
Haroon-Khel: Much like the `hostname` and `hosts_file` tasks that ive removed from the Common role, this same task appears in the `adopt_etc` role

Willsparker: Ah, alright :+1:

Willsparker (approved):

Haroon-Khel:
>This is probably ok now with the change in variable, but I want to review properly tomorrow when I feel more awake, therefore dismissing my previous review for now :-)

@sxa care to rereview? Or are you happy with the changes?

sxa (approved):
Looks ok but the commit message should be changed to indicate it's now `ansible_host` instead of `ansible_ssh_host` (Can either be done before merging or during the squash/merge procedure)

PR #1998, opened 2021-03-04
unixPB: Fix Swapfile detection

Willsparker:
Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1990#issuecomment-789549739

This _should_ fix the Swap File issue we're having, see the referenced issue. The issue for the affected machine(s) (`build-scaleway-ubuntu1604-armv7-*`) was that when running `$ cat /etc/fstab`, this was being returned: `# UNCONFIGURED FSTAB FOR BASE SYSTEM` - so the `grep` check failed, so the tasks following attempted to create the swapfile whilst it's in use.

I also changed some of the tasks to fit the format we have for the rest of the playbooks.   

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1092/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Swap_File/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Swap_File/tasks/main.yml
@@ -9,14 +9,6 @@
 # Not installed: When swap_root.rc or swap_tmp.rc does NOT equal 0
 # Assume swap path of /swapfile override where applicable
 
-- name: Test if swapfile exists
-  command: grep swap /etc/fstab
-  ignore_errors: yes
-  register: swap_exists
-  tags:
-    - swap_file
-    - skip_ansible_lint
-
 - name: Set Swapfile path - /swapfile
   set_fact:
     swap_path: /swapfile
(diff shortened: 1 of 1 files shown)


gdams (approved):
LGTM once linter issues fixed

Willsparker:
Thanks for the commit :sweat_smile:

Willsparker:
Hang on, the VPC run failed

Willsparker:
I'll rerun it now, just to check, but it should be good after that :-) https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1093/

Willsparker:
1093 failed, due to `mount` module needing a `path` variable to be specified, however, I believe because `state: present`, it's not actually used. The `path` variable replaced the `name` variable in this module quite some time ago, so I've just made it 'none', like it was before. VPC Run: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1094/

karianna:
CentOS6 GH Action failed?

sxa (approved):
. o O (So the logic around detecting the existing swapfile was just plain broken? Wonder how this ever worked ...)

I'm also sceptical of the special cases that set it to `/tmp/swapfile` in a couple of situations. Suspect we can remove that, but that can be done separately (And I'd want to look back on the history to understand why it was added first - maybe a file system size issue, but `/tmp` isn't really a sensible place for a swapfile IMHO)

Willsparker:
The GHCheck failing is seemingly due to issues with installing Python3 - it looks like there's both a broken URL, and `wget` isn't installed
```
https://vault.centos.org/6.10//os/x86_64/repodata/repomd.xml: [Errno 14] PYCURL ERROR 7 - "Failed to connect to 2600:1f16:c1:5e03:a992:8ee1:99d2:6f8: Cannot assign requested address"
Trying other mirror.
Error: Cannot retrieve repository metadata (repomd.xml) for repository: base. Please verify its path and try again
/bin/sh: wget: command not found
tar (child): Python-3.6.10.tgz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
```

PR #1697, opened 2020-11-20
aixPB: xlc13 changes

aixtools:
##### Checklist

- [x] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md#commit-messages)

PR #2007, opened 2021-03-09
aixPB: bypass the need for yum.sh to download and install AIXToolbox yum package

aixtools:
##### Checklist


aixtools:
The Linter/Yamllint comes up with an error - but compared with #1696, #1697 and #1698 where a similar syntax of:
```
- name: Label the Block
  block:
  - name: Description of first task in block # this is the line yamllint complains about
```
* This is the yamllint explanation:
```
Check failure on line 28 in ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
GitHub Actions
/ Yamllint

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml#L28

[indentation] wrong indentation: expected 4 but found 2
```
* note: yamllint is giving the same complaint twice when viewed via https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/2007/files
image

sxa (commented):
> The last months using the script (manually) yum.sh - has failed more times than it has succeeded.

This still needs to download `rpm.rte` though, which is on the same server so I don't think this will solve that particular problem from being an annoyance.

aixtools:
It was downloading yum.sh from the server without an issue. I do not care to speculate why the yum.sh script fails when called from Oregon - but works fine from NL. During testing (in Oregon) I noticed the downloads are MUCH faster, and seem to work consistently.

sxa:
> It was downloading yum.sh from the server without an issue. I do not care to speculate why the yum.sh script fails when called from Oregon - but works fine from NL. During testing (in Oregon) I noticed the downloads are MUCH faster, and seem to work consistently.

I'm a bit confused here - you sat the the script fails when called from Oregon, but also that during testing in Oregon it works consistently,

Also there are two initial downloads - the `yum.sh` and the `rpm.rte` - are you suggesting in this PR that `yum.sh` is problematic and `rpm.rte` is ok? We should add that information into this PR for clarity if that's the case (But it seems very odd if true given that they're both on the same FTP server...) If there are such failures we should look at reporting them via https://community.ibm.com/community/user/power/communities/community-home?CommunityKey=10c1d831-47ee-4d92-a138-b03f7896f7c9 or another forum.

aixtools:
> 
> 
> > It was downloading yum.sh from the server without an issue. I do not care to speculate why the yum.sh script fails when called from Oregon - but works fine from NL. During testing (in Oregon) I noticed the downloads are MUCH faster, and seem to work consistently.
> 
> I'm a bit confused here - you sat the the script fails when called from Oregon, but also that during testing in Oregon it works consistently,
> 
> Also there are two initial downloads - the `yum.sh` and the `rpm.rte` - are you suggesting in this PR that `yum.sh` is problematic and `rpm.rte` is ok?
In December (both mine and @Haroon-Khel experience) (and yesterday (and maybe last Friday) @Haroon-Khel requested my assistance because yum.sh was not running correctly. I do not care, nor do I want to speculate why yum.sh has trouble `fetching` rpm.rte. The fact is: when it fails it uninstalls `rpm.rte` and the system is worse off. Maybe the issue is AIX ftp client? As I said, I don't care.

@Haroon-Khel suggested using `Vendor_Files` and for something new I can create that setup and test it. However, I recall reading (between the lines in other PRs) that you (@sxa) would prefer to not add to Vendor_Files - especially when (semi-)licensed material is concerned.
I thought it worthwhile to show that the role can operate using get_url: in case something other than Vendor_Files is preferred.

We should add that information into this PR for clarity if that's the case (But it seems very odd if true given that they're both on the same FTP server...)
From three different servers (Oregon, Netherlands, UK) - three different IP addresses are the result of `ping ftp.dhe.ibm.com`:
- 170.225.15.112
- 129.35.224.112
- 92.242.132.24
 If there are such failures we should look at reporting them via https://community.ibm.c92.242.132.24om/community/user/power/communities/community-home?CommunityKey=10c1d831-47ee-4d92-a138-b03f7896f7c9 or another forum.
* I tried reporting this last December, they could not find anything wrong - even as I tested and it failed.
* In January further testing showed no more issues - so I thought it was resolved until @Haroon-Khel came back with a request for help.

* Footnote: the way yum.sh works - the transfer tends to be multiple times slower that an ansible get_url: - just an observation, not a complaint.

Haroon-Khel:
@aixtools after some convincing from @sxa, it would be more ideal to place working (non corrupted) copies of `rpm.rte` and `yum_bundle.tar` at http://ci.adoptopenjdk.net/userContent

sxa:
>  From three different servers (Oregon, Netherlands, UK) - three different IP addresses are the result of ping ftp.dhe.ibm.com:

That's not inherently a problem (although from where I am in the UK and the OSUOSL boxes it resolves as `170.225.15.112` right now) but what was the failure? Was it a DNS lookup failure, a ping failre, a failure to route to the server, a failure to connect to the FTP port specifically? If they weren't able to identify a fault in the server we should at least consider the possibility it's not a problem at that end.

karianna:
@aixtools - YAMLLint (2 v 4 spaces)

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
```diff
@@ -6,39 +6,61 @@
...
+  when: _yum.stat.islnk is not defined
   tags:
+    - yum
     - rpm_remove
     # TODO: rpm used in place of yum or rpm_key module
     - skip_ansible_lint
-
 ####################################
 # Install yum and update to latest #
+#   url: ftp://public.dhe.ibm.com/aix/freeSoftware/aixtoolbox/ezinstall/ppc/yum.sh
```
sxa, line 24:
```suggestion
# Previous version of this downloaded and executed
#   url: ftp://public.dhe.ibm.com/aix/freeSoftware/aixtoolbox/ezinstall/ppc/yum.sh
# These steps perform the same atasks without requiring python (?) to be installed first
```

sxa (changes requested):
Suggest this doc update just to make it clear where these operations came from and why we aren't using the "normal" method

aixtools:
From ojdk06, on AIX 7.2 TL2 SP5 - the download, while not what I would consider `fast` (232.3Kbytes) it seems to be accurate: AND inaccurate was the bigger issue.
```
ftp> mget rpm.rte
227 Entering Passive Mode (170,225,15,26,126,241)
150 Opening BINARY mode data connection for rpm.rte (40987648 bytes).
226 Transfer complete.
40987648 bytes received in 172.3 seconds (232.3 Kbytes/s)
local: rpm.rte remote: rpm.rte
ftp> dir rpm.rte
227 Entering Passive Mode (170,225,15,26,129,203)
150 Opening ASCII mode data connection for /bin/ls.
-rw-rw-r--    1 35092    200        40987648 Feb 15 01:03 rpm.rte
226 Transfer complete.
ftp> !ls -l rpm.rte
-rw-r--r--    1 root     system     40987648 Mar 23 08:45 rpm.rte
```

From my Power6 (in the Netherlands versus Oregon) - also connecting to service.boulder.ibm.com: much faster (factor 10x)
```
ftp> mget rpm.rte
227 Entering Passive Mode (170,225,15,26,235,161)
150 Opening BINARY mode data connection for rpm.rte (40987648 bytes).
226 Transfer complete.
40987648 bytes received in 14.1 seconds (2838 Kbytes/s)
local: rpm.rte remote: rpm.rte
```

Further study of my NL partitions:

* While getting, comsistently, 2800+ Kbytes from one partition I was getting, consistently, 450+ Kbytes from other partitions. Looking for differences I saw that I had increased the default tcp_buffers (ISNO settings) for the ethernet connection. When I did the same on the `slow` partitions - speed increased.
* Testing on ojdk06 - even with the ISNO change - speed from IBM ftp site remains slow - but DL speed from my NL server increased to an amount that makes me believe my server was the bottleneck - not ojdk06.
* Summary: from my NL server ojdk06 has input between 1800 and 2800 - consistently; from the IBM site - for the file rpm.rte it continues to be 150k (and less) - sometimes with error.

* Getting back to a question from @sxa - what is the problem. IMHO - not rpm.rte, but yum.sh. And the key issue is that yum.sh is not able to adequately respond to things gone wrong.

* My confidence is higher in ansible using get_url to get the two files yum.sh picks up, and then using command: statements to install the downloads.

sxa (approved):
LGTM - Let's use this to set up the replacements machines as discussed in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/2115

PR #2015, opened 2021-03-09
AixPb: add debug role for aix

Haroon-Khel:
Short: Windows and Unix playbooks have it, so why not Aix

Long: I find that debug messages which print out variable values, such as `inventory_hostname`, `ansible_hostname`, `ansible_host`, etc, are helpful when adding features to the playbooks which require conditions based on these variables. Its laborious to have to add these debug messages manually, only to have to remove them before pushing changes. Its identical to the [Unix debug role](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Debug/tasks/main.yml), except for the `Nagios_Monitoring` and `Nagios_Master_IP` variables, those have been removed since they are not used in the aix playbook.

Do let me know if you feel others should be removed


Willsparker (approved):
LGTM - Useful for debugging !

PR #2020, opened 2021-03-10
UnixPb: add stdout of registered variable

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1910#issuecomment-795907758

The condition should compare the stdout of the variable, not the whole variable. Else it fails with this error

```
fatal: [test-macstadium-macos1014-x64-1]: FAILED! => {"msg": "The conditional check 'not macos_version | regex_search(\"10.12\")' failed. 
The error was: Unexpected templating type error occurred on ({% if not macos_version | 
regex_search(\"10.12\") %} True {% else %} False {% endif %}): expected string or bytes-like object\n\n
The error appears to be in '/tmp/awx_763_m5aop2n8/project/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml': 
line 93, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\n
The offending line appears to be:\n\n\n- name: Install Build Tool Packages NOT macOS 10.12\n  ^ here\n"}
```

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
@@ -96,7 +96,7 @@
   homebrew: "name={{ item }} state=present"
   with_items: "{{ Build_Tool_Packages_NOT_10_12 }}"
   when:
-    - not macos_version.stdout | regex_search("10.12")
+    - not (macos_version.stdout | regex_search("10.12"))
   tags: build_tools
 
 - name: Install Build Tool Casks
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
@@ -96,7 +96,7 @@
   homebrew: "name={{ item }} state=present"
   with_items: "{{ Build_Tool_Packages_NOT_10_12 }}"
   when:
-    - not macos_version | regex_search("10.12")
+    - not macos_version.stdout | regex_search("10.12")
   tags: build_tools
 
 - name: Install Build Tool Casks


sxa:
@gdams Were the playbooks used for setting up the new machines at nine? I would have thought you'd have hit this, unless you were using an alternate version of ansible that didn't fail this check.

Haroon-Khel:
Which version of ansible does the AWX server use?

My own mac hit this error too, I use:
```
ansible 2.9.13
  config file = None
  configured module search path = ['/Users/hkhel/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.9.13_1/libexec/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.0 (default, Oct 27 2020, 14:15:17) [Clang 12.0.0 (clang-1200.0.32.21)]
```

Haroon-Khel:
@sxa 
test-macincloud-macos1010-x64-2 and test-macincloud-macos1010-x64-1 are still unreachable. It would be ideal if I could test the playbook on either of these, since they are not currently being used by a jenkins job, hence I could bring them temporarily offline

sxa:
It owul dhave been whichever one was latest and supplied with the verison of AWX we use, but I don't think it matters too much if (a) it's broken and (b) you get the same problem :-)

ping @gdams again - were the playbooks used to set up the macos machines at nine?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
```diff
@@ -96,7 +96,7 @@
   homebrew: "name={{ item }} state=present"
   with_items: "{{ Build_Tool_Packages_NOT_10_12 }}"
   when:
-    - not macos_version | regex_search("10.12")
+    - not macos_version.stdout | regex_search("10.12")
```
Willsparker, line 99:
```suggestion
    - not (macos_version.stdout | regex_search("10.12"))
```
This reads better to me, personally :-)

Haroon-Khel: Just for you 😉

Willsparker (approved):
You're a saint

gdams (approved):
Weird, just tried this on my machine and the previous version fails in the same way as reported. I did bump my ansible version recently though?

sxa:
@gdams To be clear were the nine machine set up with the full playbooks from openjdk-infrastructure?

gdams:
Yes they had the full playbook run on them, Not sure how it skipped over that error though

gdams:
Just checked the nine machines:

```
admin@test-nine-macos1015-x64-1 JavaVirtualMachines % brew install ccache
Warning: ccache 4.2 is already installed and up-to-date.
To reinstall 4.2, run:
  brew reinstall ccache
```

They both have `ccache` installed so the logic must have worked when I ran the playbook

Issue #1980, opened 2021-02-26
VPC: Debian10 apt-get upgrade GRUB issue


Willsparker:
Log: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Debian10,label=vagrant/1086/console

I hit this whilst running locally too - the issue appears to be that `grub-pc` is being upgraded, which causes a pop-up, that Ansible can't handle. I googled around , and found this: https://unix.stackexchange.com/questions/422282/apt-get-y-upgrade-keep-asking-me-to-confirm-keep-tmp-grub-xc3mmkp0zx-file-and , and supposedly setting `DEBIAN_FRONTEND=noninteractive` is meant to fix it, however I was unable to get this to work- having modified the ansible task:

```
- name: Run apt-get upgrade
  apt:
    upgrade: safe
    update_cache: yes
    force: yes
    dpkg_options: 'force-confdef'
  environment:
    DEBIAN_FRONTEND: noninteractive
  tags: patch_update
```
This gave a different error message (cut out the not needed bits):
```
"Setting up grub-pc (2.02+dfsg1-20+deb10u3) ...", "/dev/vda does not exist, so cannot grub-install to it!", "You must correct your GRUB install devices before proceeding:", "", "  dpkg-reconfigure grub-pc", "dpkg: error processing package grub-pc (--configure):", " installed grub-pc package post-installation script subprocess returned error exit status 1", "Errors were encountered while processing:", " grub-pc"]}
```

PR #2022, opened 2021-03-11
vagrant: apt-mark hold grub-pc on Debian10

Willsparker:
Fixes: #1980 

As mentioned in the issue, the Debian10 Vagrant VM wants to upgrade `grub-pc` when running the upgrade task from the playbook, however, that causes a popup, which Ansible can't handle. The reason I have specifically put in a Vagrantfile fix as opposed to a playbook fix is because if this were a _real_ machine, the `grub-pc` package could be installed manually - and I don't know of the complications that not upgrading `grub-pc` can cause - however, that doesn't matter on a VPC run.

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) : https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1103/

### DIFF
--- a/ansible/Vagrantfile.Debian10
+++ b/ansible/Vagrantfile.Debian10
@@ -4,7 +4,11 @@
 $script = <<SCRIPT
 sudo apt-get install tree -y
 sudo apt-get install software-properties-common -y
-sudo apt-get install gpg -y	
+sudo apt-get install gpg -y
+
+# The Debian10 VM needs to upgrade grub-pc, which causes a pop-up, breaking VPC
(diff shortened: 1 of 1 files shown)


sxa (approved):
Done .. Not an ideal long term solution but it's documented adequately if we're able to revisit it in the future

Willsparker:
Ideally, I'd like to rebuild _all_ of the vagrant boxes at some point so we don't need hacky things like this, and it's well documented how we can update them

Issue #1918, opened 2021-02-10
build-osuosl-aix71-ppc64-2: tests which create core files fail


lumpfish:
https://ci.adoptopenjdk.net/job/Test_openjdk16_j9_extended.functional_ppc64_aix/21/consoleFull
Tests which analyse the contents of a core file taken during the test are failing on build-osuosl-aix71-ppc64-2.  The error messages suggest the core file is not actually being written successfully:
```
21:29:10  Testing: Run !allclasses
21:29:10  Test start time: 2021/02/08 13:29:09 Pacific Standard Time
21:29:10  Running command: /home/jenkins/workspace/Test_openjdk16_j9_extended.functional_ppc64_aix/openjdkbinary/j2sdk-image/bin/jdmpview -core j9core.dmp
21:29:10  Time spent starting: 29 milliseconds
21:29:11  Time spent executing: 1688 milliseconds
21:29:11  Test result: FAILED
21:29:11   [OUT] DTFJView version 4.29.5, using DTFJ version 1.12.29003
21:29:11   [OUT] Loading image from DTFJ...
21:29:11   [OUT] 
21:29:11   [OUT] Could not load dump file and/or could not load XML file: Image file '/home/jenkins/workspace/Test_openjdk16_j9_extended.functional_ppc64_aix/openjdk-tests/TKG/output_16128142891542/cmdLineTester_classesdbgddrext_1/j9core.dmp' not found.
21:29:11   [OUT] For a list of commands, type "help"; for how to use "help", type "help help"
21:29:11   [OUT] > DDR is not enabled for this core file, '!' commands are disabled
21:29:11   [OUT] > 
```
To rerun one of the failing tests: https://ci.adoptopenjdk.net/job/Grinder/parambuild/?JDK_VERSION=16&JDK_IMPL=openj9&JDK_VENDOR=adoptopenjdk&BUILD_LIST=functional&PLATFORM=ppc64_aix&TARGET=testDDRExt_General_0

lumpfish:
Still seeing this on the jdk16 milestone builds: https://ci.adoptopenjdk.net/job/Test_openjdk8_j9_sanity.functional_ppc64_aix/136/

lumpfish:
[...]

PR #2052, opened 2021-03-18
aixPB: update RBAC role to include priv needed for core dumps

aixtools:
Fix: #1918 
##### Checklist


Haroon-Khel:
Testing on ojdk06

Haroon-Khel (commented):
Despite this change, I still get a dump error when running the testDDRExt_General_0 test

```
     [exec] addTestJitHint adding hint to java/lang/Thread$State.values
     [exec] JVMDUMP034I User requested System dump using '/home/jenkins/openjdk-tests/TKG/../TKG/output_16165189377817/testDDRExt_General_0/DDREXT.J9CORE.DMP' through com.ibm.jvm.Dump.SystemDump
     [exec] JVMDUMP012E Error in System dump: cannot find core file: "A file or directory in the path name does not exist.". check "ulimit -Hc" is set high enough
```

Disabling the rtclk role for the jenkins user did allow this test to pass

aixtools:
Probably yet another PRIV needed. There is a reason people have found RBAC tough to implement. Lack of documentation re: what PV unlocks what (or locks by side-effect) is itterating.

If you can help out with a test I can run stand-alone (eg, with a save workspace) I can run tracepriv to see what are being used.

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1918#issuecomment-802985284

@sxa's testing of the pr

aixtools:
re: #1918 - the AIX 7.2 they are testing on is AIX 7.2 TL5 SP1 - which is the 'latest' version of AIX - and not likely (yet) to be the version most businesses are running.

Are we still committed to supporting AIX 7.1 TL4 SP4 (or earlier) and later AIX systems - OR - do we need to update the AIX systems to a later 'base' level.

sxa:
> Are we still committed to supporting AIX 7.1 TL4 SP4 (or earlier) and later AIX systems - OR - do we need to update the AIX systems to a later 'base' level.

That seems like something separate to this issue, unless upgrading the SP is likely to make a difference to the solution. Do you think it would?

sxa:
What is the current situation on this - do we need to make further changes or are we confident that the set of tests which generate core dumps are ok with what's here?

aixtools:
We never really tested this. Instead we have hacked away at jenkins user settings.

Depending on the `capabilities` setting will always be dependent on AIX os-level, and there is no documentation (easily available).

sxa:
> Depending on the `capabilities` setting will always be dependent on AIX os-level, and there is no documentation (easily available).

Sounds like an opportunity to be trailblazers and put some docs in place (or even feedback to IBM if they want a blog post from us on our findings!)

aixtools:
As this PR has conflicts - closing it and deleting this branch. I'll open a new PR with a new current patch.

Issue #1544, opened 2020-09-15
AIX playbook should not put `/opt/freeware/bin` at start of /etc/environment


sxa:
This can break certain default AIX startup scripts which require the AIX `grep` command.

We need to determine the impact to the adoptopenjdk scripts of changing this.

@Haroon-Khel Possibly somethign we can do while switching to the invididual Role-based playbooks?

aixtools:
The tactic I have seen most, over the years, is to either add PATH at the beginning of a script (as opposed to making something a system default), OR having variables defined to execute a specific command, e.g. GREP=/opt/freeware/bin/grep and then calling it using ${GREP}.

And after glancing at the playbook - I see you are 'installing', i.e., attempting to install, a dated version of openSSL (1.0.2.1601) while the version 1.0.2.2100 is already installed.

FYI: 
![image](https://user-images.githubusercontent.com/10852597/93395735-b84b6680-f876-11ea-9cba-f0434a6b7f1d.png)

aixtools:
Hmm. Guess my message from yesterday did not take.

Short bit: openssl is also not current. playbook says 1601, but 2100 (1.0.2u) is current version.

sxa:
@aixtools I think we need to be clear on why that PATH statement was there in the first case really (Doesn't appear to be commented in the playbook...) I suspect we should undo it on one of the machines and see what breaks :-)

Haroon-Khel:
@sxa Do we know which AIX start up scripts fail as a result of this? And is the issue here that the PATH variable is simply at the **start** of `/etc/environment`, or that its there at all?

aixtools:
One thing it breaks is rc.nfs. I wanted to do some AIX maintenaince
(using NFS) - and fell over this. It takes too much time to find and
document all the things that break because the syntax of 'coreutils'
and/or 'binutils' is different than the AIX versions.

So, it is not that it is in the PATH, but that it is first. Note: if a
nodejs command is requiring a coreutils/binutils command syntax - and it
is not providing the PATH - then that will break.

In short, the 'bad' is that the systems are unmanageable. Updates cannot
be applied (at least not on the system I first saw this, and it had not
changed the PATH - it had overwritten /usr/bin/grep.

Hope this helps,
Michael (aka aixtools)

aixtools:
Smarter now: I see where it happens in aix.yml
```

      ######################################
      # Add bash to available login shells #
      ######################################
[...]

PR #2057, opened 2021-03-19
AixPb: Put /usr/bin to the front of the PATH

Haroon-Khel:
Following on from a discussion with @sxa in slack, /usr/bin should be at the beginning of the PATH variable, ahead of /opt/freeware/bin

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -97,7 +97,7 @@
         path: /etc/security/login.cfg
         regexp: 'shells = '
         replace: 'shells = /bin/bash,'
-      tags: 
+      tags:
         - login_shell
         - adoptopenjdk
 
@@ -118,7 +118,7 @@
           AIXTHREAD_HRT=true
           PKG_CONFIG_PATH=/opt/freeware/lib64/pkgconfig:/opt/freeware/lib/pkgconfig
           PERL5LIB=/opt/freemarker/lib/perl5
-      tags: 
+      tags:
         - login_shell
         - adoptopenjdk
 
@@ -127,7 +127,7 @@
         path: /etc/environment
         regexp: 'PATH=/usr/bin:/etc'
         replace: 'PATH=/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:/etc'
-      tags: 
+      tags:
         - login_shell
         - adoptopenjdk
 
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -97,14 +97,18 @@
         path: /etc/security/login.cfg
         regexp: 'shells = '
         replace: 'shells = /bin/bash,'
-      tags: login_shell
+      tags: 
+        - login_shell
+        - adoptopenjdk
 
     - name: Add bash to available login shells
       blockinfile:
         dest: /etc/shells
         block: |
           /bin/bash
-      tags: login_shell
+      tags: 
+        - login_shell
+        - adoptopenjdk
 
     # move to role later
     - name: Set variables for global environment
@@ -114,14 +118,18 @@
           AIXTHREAD_HRT=true
           PKG_CONFIG_PATH=/opt/freeware/lib64/pkgconfig:/opt/freeware/lib/pkgconfig
           PERL5LIB=/opt/freemarker/lib/perl5
-      tags: login_shell
+      tags: 
+        - login_shell
+        - adoptopenjdk
 
     - name: Add freeware and xlc to PATH for global environment
       replace:
         path: /etc/environment
         regexp: 'PATH=/usr/bin:/etc'
         replace: 'PATH=/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:/etc'
-      tags: login_shell
+      tags: 
+        - login_shell
+        - adoptopenjdk
 
   ##############
   # freemarker #
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -120,7 +120,7 @@
       replace:
         path: /etc/environment
         regexp: 'PATH=/usr/bin'
-        replace: 'PATH=usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin/'
+        replace: 'PATH=/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin/'
       tags: login_shell
 
   ##############
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -92,9 +92,12 @@
   ######################################
   # Add bash to available login shells #
   ######################################
-    - name: Add bash, ksh and /bin/false to available login shells
-      command: chsec -f /etc/security/login.cfg -s usw -a shells=/usr/bin/bash,/usr/bin/ksh,/bin/false
-      tags:
+    - name: Add bash to available login shells
+      replace:
+        path: /etc/security/login.cfg
+        regexp: 'shells = '
+        replace: 'shells = /bin/bash,'
+      tags: 
         - login_shell
         - adoptopenjdk
 
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -120,7 +120,7 @@
       replace:
         path: /etc/environment
         regexp: 'PATH=/usr/bin'
-        replace: 'PATH=/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin'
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
ping @aixtools

aixtools (changes requested):
Perhaps a typo - missing '/' in string PATH=usr/bin:...

Also, why not use something like the PATH: statement on line 11:
 
 ```
    +10    environment:
   +11      PATH: "/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:{{ ansible_env.PATH }}"
```
And, I wonder if it should not ALSO be changed on line 11?

**Comment by Haroon-Khel** [comment]
I searched through the aix playbook. I dont think that PATH environment variable is used at all. Will need to make a more thorough search

**Review by aixtools — APPROVED** [review]

**Comment by Haroon-Khel** [comment]
Before merging, I will have to test a build and test with this new PATH variable

**Comment by Haroon-Khel** [comment]
jdk11 hotspot build on test-ibm-aix71-ppc64-1 with the new PATH variable to test
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-aix-ppc64-hotspot/890/console

**Comment by Haroon-Khel** [comment]
>jdk11 hotspot build on test-ibm-aix71-ppc64-1 with the new PATH variable to test
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-aix-ppc64-hotspot/890/console

That passed. Running a jdk11 openjdk sanity test
https://ci.adoptopenjdk.net/job/Grinder/7960/console

**Comment by Haroon-Khel** [comment]
That too passed. A nightly openj9 jdk8 job is running on the machine. Just for safe measure it can be used as a final indicator as to whether this new PATH variable breaks anything
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-aix-ppc64-openj9/999/consoleFull

**Comment by Haroon-Khel** [comment]
The above job passed

**Review by aixtools — CHANGES_REQUESTED** [review]
While this works - have you tested it to be sure it works a second time?
i.e., when PATH=/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin/:...

does not regexp to:

PATH=/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin/:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin/:...

How about including /etc in the regexp but replacing it - in a way that the regexp will not match a second time around?

**Comment by Haroon-Khel** [comment]
@aixtools Youre right it does.

The /etc idea is good, but the question is will all new machines have a `/usr/bin:/etc` block in their PATH?

I was thinking the new PATH could be `/opt/IBM/xlC/13.1.3/bin:/usr/bin:/opt/freeware/bin`? That would make the change idempotent as there wouldnt be a `PATH=/usr/bin` to for ansible to find

**Comment by aixtools** [comment]
I can look at few more systems for a default. If needed - in the future 
this could use an with: items: on the regexp. For now, 
=/usr/bin:/etc:... seems to have been default PATH for many levels.

**Comment by Haroon-Khel** [comment]
That works. This regexp is now idempotent

**Review by sxa — CHANGES_REQUESTED** [review]
Can we ensure that this does not run if the `adoptopenjdk` tag is skipped - we shouldn't be modfiying system defaults like this if someone else is running them and skipping the adoptopenjdk ones, especially since it could remove things that they already have in there.

**Comment by Haroon-Khel** [comment]
@sxa Just that task? Or the other `login_shell` tagged tasks too?

```
    - name: Add bash to available login shells
      replace:
        path: /etc/security/login.cfg
        regexp: 'shells = '
        replace: 'shells = /bin/bash,'
      tags: login_shell

    - name: Add bash to available login shells
      blockinfile:
        dest: /etc/shells
        block: |
          /bin/bash
      tags: login_shell

    # move to role later
    - name: Set variables for global environment
      blockinfile:
        dest: /etc/environment
        block: |
          AIXTHREAD_HRT=true
          PKG_CONFIG_PATH=/opt/freeware/lib64/pkgconfig:/opt/freeware/lib/pkgconfig
          PERL5LIB=/opt/freemarker/lib/perl5
      tags: login_shell
```

sxa:
> Can we ensure that this does not run if the `adoptopenjdk` tag is skipped - we shouldn't be modfiying system defaults like this if someone else is running them and skipping the adoptopenjdk ones, especially since it could remove things that they already have in there.

Most of those are additions rather than potentially breaking things, so while it's not as critical, I'd say that yes they should all have it, although I'm not sure why we need to add `bash` to the list. It shouldn't be critical for jenkins to work.

aixtools (changes requested):
Couple of things:
* Considering /bin/bash is a symbolic link outside of / I do not consider it a fantastic argument for a shell (for AIX admin accounts - /bin/sh and /bin/ksh used to be as login shells AND the AIX install makes sure 'shells' are installed as binary programs - not links - these may be in the default shells - BETTER - would be (even as a symbolic link) as more inline with AIX user administration to have the bash shell be /usr/bin/bash.
* Also, the regexp used to add `bash` as a shell leaves it open for recursive addition
* Suggestion: replace `regexp` with 'shells = /bin' and the new string to be 'shells = /usr/bin/bash,/usr'

aixtools (commented):
* replace `/bin/bash` with `/usr/bin/bash`

aixtools (changes requested):
Urgh - my two - now comments - are meant as request changes - so this `comment` is actually two `requests for change`

Haroon-Khel:
@aixtools instead of a regexp, why not something like

`chsec -f /etc/security/login.cfg -s usw -a shells=/usr/bin/bash` ?

On ojdk06, the shells in /etc/security/login.cfg were 

`/usr/bin/ksh,/usr/bin/ksh93,/usr/bin/rksh,/usr/bin/rksh93,/bin/false`

So how about something like

`chsec -f /etc/security/login.cfg -s usw -a shells=/usr/bin/bash,/usr/bin/ksh` ?

Would that suffice? This command is also idempotent

aixtools:
* That the shells include /bin/false is part of my standard hardening (and I have already shortened the default list of shells)
* Using command: chsec seems like a good idea - but to be completely idempotent still use something to check if bash is already among the shells (e.g, not the first one).
* Deleting shells is not a major issue (it mainly affects `chsh` iirc, but please leave /bin/false as a security measure (to block accounts that are not meant to have a shell from ever getting a shell, even via su).

Haroon-Khel:
@aixtools regarding your idempotency comment, would it not be more consistent for all new systems to simply have `/usr/bin/bash,/usr/bin/ksh,/bin/false` in their shells line?

Also, would changing this list of shells affect the way jenkins behaves on the machine?

aixtools:
> 
> 
> @aixtools regarding your idempotency comment, would it not be more consistent for all new systems to simply have `/usr/bin/bash,/usr/bin/ksh,/bin/false` in their shells line?
> 
> Also, would changing this list of shells affect the way jenkins behaves on the machine?

* No, won't change anything for existing accounts. It changes the behavior of `mkuser`, `chuser`, and `chsh` - all from memory.
* I am fine with only three shells in the setting - and the regexp can be to test for that string, and change it if it is different. Should something additional ever be needed - there will be an issue and a PR to fix it. I don't expect anything. Although - would be nice to leave ksh93. :)

aixtools:
> 
> 
> > Can we ensure that this does not run if the `adoptopenjdk` tag is skipped - we shouldn't be modfiying system defaults like this if someone else is running them and skipping the adoptopenjdk ones, especially since it could remove things that they already have in there.
> 
> Most of those are additions rather than potentially breaking things, so while it's not as critical, I'd say that yes they should all have it, although I'm not sure why we need to add `bash` to the list. It shouldn't be critical for jenkins to work.

FYI: jenkins userid does not need bash as shell. iirc - gmake is calling bash directly.

aixtools:
Fixes: #1544

sxa:
> FYI: jenkins userid does not need bash as shell. iirc - gmake is calling bash directly.
Correct - in fact the playbooks create the jenkins user with `/usr/bin/ksh` as the shell so ... to be honest I don't know why we even make that change - presumably somewhere in the past liked creating other user IDs with bash as the shell ;-)

sxa (commented):
For reference, the existing `/etc/security/login.cfg` on aix72-2 (ojdk04) looks like this:
`        shells = /bin/bash,/bin/bash,/bin/bash,/bin/bash,/bin/bash,/bin/bash,/bin/bash,/bin/bash,/bin/sh,/bin/bsh,/bin/csh,/bin/ksh,/bin/tsh,/bin/ksh93,/usr/bin/sh,/usr/bin/bsh,/usr/bin/csh,/usr/bin/ksh,/usr/bin/tsh,/usr/bin/ksh93,/usr/bin/rksh,/usr/bin/rksh93,/usr/sbin/uucp/uucico,/usr/sbin/sliplogin,/usr/sbin/snappd`
And the chsec command will wipe most of those out, which seems somewat undersirable. I feel this role needs to just add entries, not change the ones already on the system.

However I'm not sur what the relationship between these entries and `/etc/shells` is - @aixtools do you know?

aixtools:
@sxa - not sure why or when `/etc/shells` showed up in AIX.

* it exists in AIX 5.3 (so as early as 2004)
* it may have come with AIX 5.1 as part of Linux affinity (since Linux `chsh` uses /etc/shells the same way AIX `chsh` uses `shells` attribute of stanza `usw` in `/etc/login.cfg`.
* As far as I know the `shells` attribute is the only bit that affects `chsh`, and perhaps `[mk|ch]user` commands.

Haroon-Khel:
Closing due to pr not passing eclipse auth check. Reopened at https://github.com/adoptium/infrastructure/pull/2288

PR #2026, opened 2021-03-11
Add musl-dev to Alpine static docker containers

sxa:
Fixes the issue with missing `<stdio.h>` on the machines which prevents compilation of the system test suites (Ref [slack conversation](https://adoptopenjdk.slack.com/archives/CLCFNV2JG/p1615469905017800?thread_ts=1615456719.007700&cid=CLCFNV2JG) where it was raised and verified by @Mesbah-Alam - thanks!)

##### Checklist

- [x] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md#commit-messages)
- [ ] [FAQ.md](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md) updated if appropriate
- [ ] other documentation is changed or added (if applicable)
- [ ] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

PR #2058, opened 2021-03-19
unixPB: Add missing deps to Alpine for testing

aahlenst:
Adds missing dependencies to static Alpine containers for testing. Should resolve https://github.com/AdoptOpenJDK/openjdk-tests/issues/2223#issuecomment-801995589 and probably supersedes #2026.

##### Checklist

- [X] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md#commit-messages)

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.alp311
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles/Dockerfile.alp311
@@ -1,7 +1,8 @@
 FROM alpine:3.11
 
 # Install OpenJDK 8 from Alpine because AdoptOpenJDK is 16+, only (February, 2021).
-RUN apk --update add bash shadow openssh-server openssh-client unzip wget openjdk8 git curl make gcc perl xvfb libxrender libxi libxtst procps
+RUN apk --update add bash shadow openssh-server openssh-client unzip wget openjdk8 git curl make gcc perl xvfb \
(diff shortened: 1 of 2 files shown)


aahlenst:
I didn't have the time yet to run through the entire `_sanity.system`, but so far it looks like all errors caused by missing dependencies are gone.

sxa (commented):
> I didn't have the time yet to run through the entire `_sanity.system`, but so far it looks like all errors caused by missing dependencies are gone.

Are all these additional ones specic to the system test suite? Might be nice to split them out into a separate `apk` line to make it clear what the "base" testing require and what the "extras" for the system tests are. (I appreciate we havent' done this anywhere else, but this may be the start of a new trend of understanding exactly what our dependencies are :-) )

aahlenst:
musl-dev and perl-doc are the only ones for system tests. The rest are regular Java deps. I'm not sure how to handle this, because we need additional things that are not necessarily obvious (scp/sftp, locale/language settings). For example, I missed scp/sftp initially because other distributions automatically install it together with SSH. I already thought about a checklist. What do you think, @sxa?

sxa (approved):
> musl-dev and perl-doc are the only ones for system tests. The rest are regular Java deps. I'm not sure how to handle this, because we need additional things that are not necessarily obvious (scp/sftp, locale/language settings). For example, I missed scp/sftp initially because other distributions automatically install it together with SSH. I already thought about a checklist. What do you think, @sxa?

I think what we're doing is definitely the basis of such a checklist :-) Particularly with basing this on minimal docker containers some of these "extras" that might be in a default OS install show up. If those are the only two that were needed for the system tests I'm pretty much ok with this as-is - I was more nervous that there might have been loads more specific to those suites.

PR #2063, opened 2021-03-20
Install full Xcode

aahlenst:
With JEP 382, macOS gets a new rendering pipeline based on Metal. This needs tools like metal (`xcrun metal`) that are only included in the full Xcode.

##### Checklist


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
```diff
@@ -20,16 +20,12 @@
   tags:
     - skip_ansible_lint
 
-- name: Install xcode-tools
-  script: scripts/install-xcode.sh
-  when: xcode.rc > 1
```
aahlenst, line 25: Homebrew automatically installs the command line tools if required. Reduces code to maintain.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Xcode/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Install Xcode for macOS 10.15 and macOS 11.0
+  shell: "xcversion install {{item }} --retry-download-count=10"
+  when: ansible_distribution_version is version('10.15', operator='>=')
+  loop: # List of available versions: $ xcversion list
+    - 12.4
+    - 11.7
+    - 10.3
+  environment:
+    XCODE_INSTALL_USER: "{{ Apple_ID_User }}"
+    XCODE_INSTALL_PASSWORD: "{{ Apple_ID_Password }}"
```
aahlenst, line 26: Requires new secrets (a non 2FA enabled Apple ID).

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Xcode/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+    SPACESHIP_SKIP_2FA_UPGRADE: 1
+    FASTLANE_DONT_STORE_PASSWORD: 1
+
+- name: Make Xcode 10.3 the default
+  become: yes
+  shell: xcversion select 10.3
```
aahlenst, line 44: Should 10.3 be the default?

aahlenst:
@gdams @sxa I have some questions: 

* Previously, Xcode was only installed on test machines and only if they had 10.14 or 10.15 installed on them. Why? I would have thought that Xcode is only required on build machines.
* Why was Xcode downloaded from Azure, using a private copy?
* Xcode versions and OS versions are closely tied to each other. For example, the upcoming Xcode 12.5 won't be available for 10.15, only for 11 and newer. What versions do we really need?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Xcode/tasks/main.yml
gdams: yeah we use 10.3 as default right now

gdams:
> Previously, Xcode was only installed on test machines and only if they had 10.14 or 10.15 installed on them. Why? I would have thought that Xcode is only required on build machines.

@aahlenst I limited the install to a small subset of machines because it takes a long time to install from blob storage. For most builds (11-15) we use the command-line tools version. We switch between them as part of the platform-specific build scripts.

> Why was Xcode downloaded from Azure, using a private copy?

I couldn't automate the apple ID auth so I chose to stick it in Azure (privately for license reasons). Your PR removes the need for this.

> Xcode versions and OS versions are closely tied to each other. For example, the upcoming Xcode 12.5 won't be available for 10.15, only for 11 and newer. What versions do we really need?

We actually only need 10.3 so far. Can you update the PR to only install 10.3 and we can add other versions where needed. I'm trying to avoid putting several versions on machines because they take up a lot of space.

aahlenst:
> We actually only need 10.3 so far. Can you update the PR to only install 10.3 and we can add other versions where needed. I'm trying to avoid putting several versions on machines because they take up a lot of space.

Proposal: Limit to 10.3 on Intel, latest 12 on ARM machines. We'll only install the full Xcode on the build machines (the test machines should work without). Would that be fine?

gdams:
> Proposal: Limit to 10.3 on Intel, latest 12 on ARM machines. We'll only install the full Xcode on the build machines (the test machines should work without). Would that be fine?

That all sounds fine to me, We will need 10.3 on the test machines as well. We use them as part of the pool to notarize which requires `xcrun`

aahlenst:
Ready for review.

gdams (approved):

PR #2076, opened 2021-03-23
AixPb: add missing yum tags

Haroon-Khel:
Two aix yum tasks were missing `yum` tags. They are no longer missing `yum` tags


karianna (approved):
lgtm

PR #2081, opened 2021-03-25
AixPB: increase jenkins user's data and stack limits to unlimited

Haroon-Khel:
Often I find when running tests or builds on machines directly, im hit with a memory error such as
```
    [javac] # There is insufficient memory for the Java Runtime Environment to continue.
    [javac] # Native memory allocation (malloc) failed to allocate 744016 bytes for Chunk::new
    [javac] # An error report file with more information is saved as:
    [javac] # /home/jenkins/openjdk-tests/TKG/scripts/hs_err_pid12583092.log
```
Thanks to @aixtools I found that adjusting the limit settings for the jenkins user helped to resolve these errors

@sxa What do you think?

The machines on which I have recently hit this error are ojdk06 and build-osuosl-aix71-ppc64-2. However in the past I do recall hitting errors similar to https://stackoverflow.com/questions/14038074/git-pull-fatal-out-of-memory-malloc-failed on some of our aix machines when running builds/tests

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/jenkins_user/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/jenkins_user/tasks/main.yml
@@ -75,9 +75,9 @@
         - "fsize=-1"
         - "core=-1"
         - "cpu=-1"
-        - "data=262144"
+        - "data=-1"
         - "rss=65536"
-        - "stack=65536"
+        - "stack=-1"
         - "nofiles=-1"
       changed_when: False
   tags: jenkins_user


sxa:
Do you have the `LDR_MAXCNTRL` variable set?

Haroon-Khel:
@sxa I dont think so. I cant find anything on a `LDR_MAXCNTRL` variable but I can find docs on a `LDR_CNTRL`. Neither variable is set in the shell session, and I cant find either variable in /etc/environment

aixtools (approved):

karianna (changes requested):
Can you add a comment here as to why?

Haroon-Khel:
@karianna I found that increasing the stack and data limits of the jenkins user to unlimited helped to resolve some of the memory errors (in the first comment) I keep getting when running builds and tests on machines directly.

If there are other ways around this error, Id be open to implementing them instead

sxa:
> @sxa I dont think so. I cant find anything on a `LDR_MAXCNTRL` variable but I can find docs on a `LDR_CNTRL`. Neither variable is set in the shell session, and I cant find either variable in /etc/environment

That'll be it - can you verify whether `export LDR_CNTRL=MAXDATA=0x80000000` resolves those problems instead? If there's a behaviour difference between how your running them locally and under jenkins it would be better to understand why than just set thing to unlimited.

Haroon-Khel:
@sxa I was able to recreate the error on build-osuosl-aix71-ppc64-2. Exporting `LDR_CNTRL=MAXDATA=0x80000000` resolved the issue. I see that its gets exported during build jobs https://github.com/AdoptOpenJDK/openjdk-build/blob/a3b79fbc67b3232e0c22f744293cf1c6e7caf163/build-farm/platform-specific-configurations/aix.sh#L57

I should consult that file more often

PR #2084, opened 2021-03-26
unixPB: Remove OpenSSL 1.0.2r playbook

AdamBrousseau:
OpenJ9 no longer requires this as we fetch
the latest version during the builds.

##### Checklist


karianna (commented):
Wouldn't we need to keep this to be able to run older builds though?

AdamBrousseau:
I suppose. We (OpenJ9) have been cloning the latest openssl code in the build for a long time. I guess we would have to see if there were any releases built without having the source pulled and what the likelihood is we would ever need to rebuild that version.

@pshipton may have a better idea or an opinion.
Also cc @sxa since he is reviewing the sibling PR.

Looking at the Git history, the playbook was first committed on [Apr 18, 2019](https://github.com/AdoptOpenJDK/openjdk-infrastructure/commit/67a7c1ee9eb5d2ca0be21510fc73ba2c423e5fb5#) and the OpenSSL clone script was first committed to the jdk11 extensions repo on [Dec 6, 2018](https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/69393ac714ac6c863a428a591806a4a46f28cad2#diff-df29e5a94c89cdc14361587706198ef9051dc2f1aceb3ba1530c0f127b8cd748). Also, the openjdk-build linux.sh added the check for 1.0.2r on [Apr 9, 2019](https://github.com/AdoptOpenJDK/openjdk-build/commit/f994822e706f13dd5af6e61780932e9f6483ac85). Given the extensions change was the first one to go in, I would say that this will never be needed.

pshipton:
I think the likelihood of having to rebuild such an old release is extremely low. The OpenJ9 project has no expectations that older builds can be re-built after the following release comes out. I don't know that anything keeps track of the test SHAs matching each release, maybe that has been tracked at AdoptOpenJDK? Otherwise we can't test a rebuild. If there is desire to rebuilt old releases, it makes more sense to me that tags or SHAs are recorded for the playbooks, rather than trying to keep the head stream in shape to build any release.

karianna (approved):
LGTM

Willsparker (approved):
Approving based on @pshipton 's comment. And if all else fails, it's not hard to revert this :-)

PR #2098, opened 2021-03-29
doc: miscellaneous updates to CONTRIBUTING guide

sxa:
Changes split out from https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1974

##### Checklist

- [x] [FAQ.md](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md) updated if appropriate

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,21 +9,19 @@ reproducing issues and more.
 ## Mission Statement
 
 To provide **secure**, **consistent**, **repeatable**, and **auditable**
-infrastructure for the AdoptOpenJDK farm. See our full [Mission Statement] for more details.
-
-## Infrastructure Manifesto
-
-* We prefer using prebuilt, Galaxy provided Ansible playbooks over NIHing our own
-* We prefer using binaries from official repositories over building our own
-* We prefer explicit comments within the code to explain our reasoning over implicit assumptions
-* We embrace the Chaos Monkey
+infrastructure for the AdoptOpenJDK farm.
 
 ## Infrastructure As Code
 
 The infrastructure project contains:
 
-1. The [Ansible Playbooks](ansible/playbooks) for bootstrapping the build and test hosts (including a way to test Ansible)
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -9,21 +9,19 @@ reproducing issues and more.
...
+infrastructure for the AdoptOpenJDK farm.
 
 ## Infrastructure As Code
 
 The infrastructure project contains:
 
-1. The [Ansible Playbooks](ansible/playbooks) for bootstrapping the build and test hosts (including a way to test Ansible)
-1. The overriding [Documentation](docs) for the build farm
+1. The [Ansible Playbooks](ansible/playbooks) for bootstrapping the build and test hosts (including a way to test Ansible).
+1. The [Vagrant files](ansible/) for running our full suite of tests.
```
Willsparker, line 19: `Vagrantfiles`, without the space.
Do we use our VFs for anything other than VPC / GitHub Actions? If not, we only use them for testing the playbooks, really.

Willsparker (changes requested):

CONTRIBUTING.md
sxa: That is the point - they run the full set of infrastructure/playbook tests (i.e. the ones we have in this repository. I can add an extra word in. I would have said that "Vagrant files" was correct, as it's the files used for testing vagrant, which includes the VPC/QPC scripts, and the build* scripts etc. and not just the `Vagrantfile` ones

sxa: I think I'll update to point to the pbTestScripts directory which is possibly more useful.

Willsparker: Yeah, I think the extra word would help, just given that there's a lot of testing going on, it can get ambiguous
The reason I say Vagrantfiles is because that's how Vagrant refers to them :-)

Willsparker (approved):
LGTM

PR #2164, opened 2021-04-29
UnixPB: change conditions to include Ubuntu versions greater and equal to 20

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/2143

The tasks https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/a4d8543253f2cf3edbe9b606b809a3ad5fd6beab/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml#L84 and https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/a4d8543253f2cf3edbe9b606b809a3ad5fd6beab/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NTP_TIME/tasks/main.yml#L5 fail on ubuntu versions 20.04 and 21.04, and so need to be skipped on these versions

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
@@ -86,7 +86,7 @@
   with_items: "{{ gcc_compiler }}"
   when:
     - (ansible_distribution_major_version != "9" and ansible_architecture != "armv7l")
-    - ansible_distribution_major_version >= "20"
+    - not (ansible_distribution_major_version >= "20")
   tags: build_tools
 
 - name: Install additional build tools for x86
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
@@ -86,7 +86,7 @@
   with_items: "{{ gcc_compiler }}"
   when:
     - (ansible_distribution_major_version != "9" and ansible_architecture != "armv7l")
-    - ansible_distribution_major_version != "20"
+    - ansible_distribution_major_version >= "20"
   tags: build_tools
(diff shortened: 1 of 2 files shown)


Willsparker (approved):
LGTM - I'm okay with no VPC run, given that it's a simple condition change :)

sxa (commented):
Hmmm elsewhere we are now using e.g. `ansible_distribution_major_version|int >= 10` - Haroon have you verified that this change works on 21.04 and doesn't produce any deprecation warnings with that syntax?
I'm less sure about the lack of VPC run on the basis of the above so I've kicked off https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1141/

Haroon-Khel:
>Hmmm elsewhere we are now using e.g. ansible_distribution_major_version|int >= 10 - Haroon have you verified that this change works on 21.04 and doesn't produce any deprecation warnings with that syntax?
I'm less sure about the lack of VPC run on the basis of the above so I've kicked off https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1141/

Ive verified that the tasks do get skipped without a deprecation warning. I used ansible 2.10.5

sxa:
> Good point - though I restarted your VPC run, as `test` was specified, without `build` :-) https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1142/

I'm tempted to suggest ignoring test if buid isn't set ... Keeps catching me out (Doesn't help that `RUN_BUILD` and `RUN_TEST` aren't next to each other in the parameter list)

Willsparker:
@Haroon-Khel VPC is all good! 

> I'm tempted to suggest ignoring test if buid isn't set ... Keeps catching me out (Doesn't help that RUN_BUILD and RUN_TEST aren't next to each other in the parameter list)

I'll fix both of those

Willsparker:
Given both @Haroon-Khel and @sxa are both away for the next 1-2 week(s), I'm going to squash and merge this so I can get to adding Ubuntu 21.04 to VPC.

Issue #2176, opened 2021-05-07
pbTests: Add logic to buildJDK.sh to set JDK_BOOT_DIR


Willsparker:
A fairly prevalent issue with both VPC and QPC is that when build JDK-8, the build scripts are unable to find JDK-7 in some cases, such as [CentOS6/x64](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1158/OS=CentOS6,label=vagrant/console), or [Ubuntu1804/ppc64le](https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/ARCHITECTURE=ppc64le,OS=ubuntu18,label=qemu/242/console). 

This bought about either due to JDK-7 having a different name to what is expected (i.e. `/usr/lib/jvm/zulu7` for platforms that use the [Zulu-7 task](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/zulu7/tasks/main.yml), `/usr/lib/jvm/java-1.7.0`, for CentOS6 that installs it via `yum`), or if JDK-7 isn't installed at all (i.e. The PPC64le / S390x QPC platforms, that can't use `Zulu-7`).

In both of these instances, the `JDK_BOOT_DIR` needs to be set to either properly point at the JDK-7 directory; or to point at the JDK-8 directory in the case that JDK-7 can't be installed (as the build scripts no longer default to using the same JDK version to build, if it can't find 'Version - 1'). 

Ref: 
https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/2121#issuecomment-829030079
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/2169#issuecomment-834168161

Willsparker:
The changes made can be seen at: https://github.com/Willsparker/openjdk-infrastructure/commit/4ec1de27fd8a1d6ee5b3ce49a0d2eb6689eaa33d

QPC Test: https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/255/
VPC Test: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1161/

This should also help fix QPC's BOOTJDK issues

Willsparker:
[...]

PR #2180, opened 2021-05-12
pbTests: Autodetect JDK-7 or default to JDK8 when building JDK8

Willsparker:
Fixes: #2176 

Adds logic to detect JDK7 in circumstances of building JDK8. Not all platforms share the same process of getting a JDK7 (i.e. it may be to install it from the package manager, install zulu-7, or it may be that JDK-7 isn't available), so they often have different names within `/usr/lib/jvm`, so  `buildJDK.sh` needs to set `JDK7_BOOT_DIR` depending on the situation.

In the case of some platforms (i.e. Ubuntu1804 on S390x and PPC64LE), there is no JDK7, so JDK8 is set as a last resort.

[VPC Run](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1163/)
[QPC Run](https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/258/) (See Ubuntu1804/s390x to see the build start)

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)

### DIFF
--- a/ansible/pbTestScripts/buildJDK.sh
+++ b/ansible/pbTestScripts/buildJDK.sh
@@ -75,6 +75,27 @@ checkJDK() {
 	fi
 }
 
+# This method is required as there isn't a standardised JDK-7 for all platforms.
+# Some may be "zulu7", some may be 'java-1.7.0', or 'openjdk-7-jdk`
+findJDK7() {
+	local jdkPath=""
+	local pathSuffix="java-1.7.0 java-7-openjdk-amd64 zulu7 jdk-7"
+	for path in $pathSuffix
+	do
+		jdkPath=$(find /usr/lib/jvm/ -name $path)
+		if [[ $jdkPath != "" ]]; then
+			break
+		fi
+	done
+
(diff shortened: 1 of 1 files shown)


sxa (approved):
I'll approve this regardless since it'll work but I'd possibly have gone with something like this (untested, off the top of my head):
```
# Autodetect potential locations for bootstrapping JDK8 builds
for JDK7DIR in jdk8 java-1.7.0 java-7-openjdk-amd64 zulu7 jdk-7; do
   if [ -x "/usr/lib/jvm/$JDK7DIR/bin/javac" ]; then
      export JDK7_BOOT_DIR=/usr/lib/jvm/$JDK7DIR
   fi
done
```
(Adding jdk8 in there is a bit of "artistic license" but it will default back to that if it doesn't find it anywhere else)

Issue #2194, opened 2021-05-25
pbTests: Check if QPC machine is fully booted properly


Willsparker:
ref: https://github.com/adoptium/infrastructure/pull/2190#pullrequestreview-664298418 

Currently, in QPC, the machine is booted and then `sleep 180` is run, assuming the machines are booted correctly after 3 minutes. This is not a particularly elegant solution. 
Now that the VM boot output is being logged, a check of that log file every 10-15 seconds could be made that will carry on with the script once the `login:` message has come up, indicating that it has booted. A timeout should probably be put in as well, to stop the Jenkins Jobs hanging if the VM boot fails.

PR #2196, opened 2021-05-26
pbTests: Better logic for checking Qemu VM has booted

Willsparker:
Fixes: #2194 

Checks every 10 seconds whether the QEMU VM has booted, by querying the log file. Timeout is currently set for 180s, as that appears to be the shortest time it takes for all VMs in QPC to boot - so if anything goes wrong with booting, it'd timeout and fail quickly. 

Testing on the `infra-ibmcloud-vagrant-x64-1` machine, though only on Ubuntu18/PPC64LE, I will put it through QPC just to make sure all the platforms work, and that the timeout isn't too restrictive.

##### Checklist

### DIFF
--- a/ansible/pbTestScripts/qemuPlaybookCheck.sh
+++ b/ansible/pbTestScripts/qemuPlaybookCheck.sh
@@ -237,11 +237,29 @@ done
      	  $EXTRA_ARGS \
 	  -nographic) > "$workFolder/${OS}.${ARCHITECTURE}.startlog" 2>&1 &
 
-	echo "Machine is booting; logging console to $workFolder/${OS}.${ARCHITECTURE}.startlog Please be patient"
-	sleep 180
-	tail "$workFolder/${OS}.${ARCHITECTURE}.startlog" | sed 's/^/CONSOLE > /g'
-	echo "Machine has started, unless the above log shows otherwise ..."
+	echo "Machine is booting; logging console to $workFolder/${OS}.${ARCHITECTURE}.startlog"
+	echo "Please be patient - this can take up to 300 seconds"
 
+	SECONDS=0
+	while [ true ];
+	do
+		if tail "$workFolder/${OS}.${ARCHITECTURE}.startlog" | grep -q login; then
+			echo "VM Booted after $SECONDS seconds"
+			break;
+		fi
+		if [ $SECONDS -gt 300 ]; then
+			echo -e "Timeout Reached. See log below:\n"
+			tail "$workFolder/${OS}.${ARCHITECTURE}.startlog" | sed 's/^/CONSOLE > /g'
+			echo
+			if [[ "$retainVM" == false ]]; then
+				destroyVM
+				echo "Removing disk image"
+				rm -f ${workFolder}/${OS}.${ARCHITECTURE}.dsk
+			fi
+			exit 127
+		fi
+		sleep 10
+	done
 	# Remove old ssh key and create a new one
 	rm -f "$workFolder"/id_rsa*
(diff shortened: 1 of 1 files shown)


ansible/pbTestScripts/qemuPlaybookCheck.sh
```diff
@@ -237,11 +237,29 @@ done
...
-	sleep 180
-	tail "$workFolder/${OS}.${ARCHITECTURE}.startlog" | sed 's/^/CONSOLE > /g'
-	echo "Machine has started, unless the above log shows otherwise ..."
+	echo "Machine is booting; logging console to $workFolder/${OS}.${ARCHITECTURE}.startlog"
+	echo "Please be patient"
 
+	SECONDS=0
+	while [ true ];
+	do
+		if tail "$workFolder/${OS}.${ARCHITECTURE}.startlog" | grep -q login; then
```
sxa, line 246: Why `tail` instead of a straight grep of the log - does `login` appear earlier on?
It's always possible  (though unlikely!) that another ten lines of stuff will appear in the 10 seconds since the line we want. Would be good to add a comment if there's a good reason for this additional complexity

ansible/pbTestScripts/qemuPlaybookCheck.sh
```diff
@@ -237,11 +237,29 @@ done
...
+			echo "VM Booted after $SECONDS seconds"
+			break;
+		fi
+		if [ $SECONDS -gt 180 ]; then
```
sxa, line 250: Since we're able to trap it as quick as possible and we're not waiting for the full amount of time I'd be tempted to increase this
```suggestion
		if [ $SECONDS -gt 300 ]; then
```

ansible/pbTestScripts/qemuPlaybookCheck.sh
```diff
@@ -237,11 +237,29 @@ done
      	  $EXTRA_ARGS \
 	  -nographic) > "$workFolder/${OS}.${ARCHITECTURE}.startlog" 2>&1 &
 
-	echo "Machine is booting; logging console to $workFolder/${OS}.${ARCHITECTURE}.startlog Please be patient"
-	sleep 180
-	tail "$workFolder/${OS}.${ARCHITECTURE}.startlog" | sed 's/^/CONSOLE > /g'
-	echo "Machine has started, unless the above log shows otherwise ..."
+	echo "Machine is booting; logging console to $workFolder/${OS}.${ARCHITECTURE}.startlog"
+	echo "Please be patient"
```
sxa, line 241: Make it clear how long the timeout is (300 assumes you also take my suggestion below, otherwise make it 180!)

```suggestion
	echo "Please be patient - this can take up to 300 seconds"
```

sxa (changes requested):

ansible/pbTestScripts/qemuPlaybookCheck.sh
Willsparker: I couldn't say for sure that `login` doesn't appear in the rest of the log - I just used tail as I knew it was definitely at the end. In my experience, additional stuff doesn't extend past 10 lines, however again, I couldn't say for sure. 
I could up it to say 15 or 20 lines using the `-n` argument?

ansible/pbTestScripts/qemuPlaybookCheck.sh
Willsparker: Fixed with the latest force push :-)

ansible/pbTestScripts/qemuPlaybookCheck.sh
Willsparker: Fixed with latest push :-)

sxa (approved):

PR #2201, opened 2021-06-02
Add Ubuntu16.04 build dockerfile

sxa:
Create and Ubuntu16.04 dockerfile similar to the CentOS6/7 ones so that we can continue to run these on a more secure host for the arm32 JDK8/11 builds which we currently build on Ubuntu 16.04

Approvals welcome but PLEASE LEAVE FOR ME TO MERGE ONCE i'M FULLY HAPPY WITH IT.

##### Checklist

NOTES:
- It looks like `openjdk-7-jdk` can no longer be located  so I have removed that and set `JDK7_BOOT_DIR` to point to openjdk-8. It may have been removed from http://ppa.launchpad.net/openjdk-r/ppa/ubuntu/pool/main/o/openjdk-7/ for `armhd` at some point. I have a tarred up version of the `openjdk-7-jdk` directory from one of the current build machines so potentially it could be deployed if we decide that using JDK8 to bootstrap JDK8 is not desirable going forward.
- If we're going to put this into prodution we'll need a way to generate and distribute the image to the build systems, but putting this dockerfile in place is a first step.
- A static copy of an image created form these changes has been created with SHA256 checksum `31ddb40b03588fe36b08202bc41beaa2d4086094871d48eb641e9835e170529a` (recording that for my own benefit as I have not published it anywhere)


sxa:
Ansible Lint failure appears unrelated to this PR but will need to be investigated

gdams:
@sxa I pushed up the jenkinsfile changes in order to build and push this dockerfile

gdams (approved):

sxa:
> @sxa I pushed up the jenkinsfile changes in order to build and push this dockerfile

If I'm reading that right you've done it for aarch64. This is being done as a requirement for armv7l which we are currently building on Ubuntu 16.04 - we should not require this build for architectures other than armv7l unless there's a different use case that you're thinking of.

gdams:
I added a build for x64 just because it would be strange to only build an arm64 image but didn't add ppc64le

sxa:
> I added a build for x64 just because it would be strange to only build an arm64 image but didn't add ppc64le

But we only need it for arm32, not arm64 :-)

gdams:
Haha my bad, I'll update the jenkinsfile later

gdams:
okay @sxa that should be good now

sxa:
Thanks. I'm going to hold off on merging for now due to another issue with the compiler that we need to think about ... (I may just think about it, procrastinate for a bit, merge it anyway, then find a better solution later)
Ref: https://github.com/adoptium/adoptium-support/issues/319

sxa:
Solaris failure does not seem related to this PR therefore merging. Have raised https://github.com/adoptium/infrastructure/issues/2226 to investigate

PR #2203, opened 2021-06-03
Remove LJC codesign certs from all machines

gdams:
We no longer need these codesign certs. The safest option is to fully remove them from the machines to prevent us signing an Adoptium build with the wrong cert

##### Checklist

- [ ] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/CONTRIBUTING.md#commit-messages)

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/macos_codesign/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/macos_codesign/tasks/main.yml
@@ -3,130 +3,20 @@
 # macOS Codesign #
 ##################
 
-- name: Set Jenkins user password
-  user:
-    name: jenkins
-    password: "{{ macOS_Jenkins_Password }}"
+# Since moving to the Eclipse Foundation, we no longer need these steps and so we've changed this role to ensure that no LJC certs exist on the machines.
+# See https://github.com/adoptium/infrastructure/blob/ccdc434d5f13861a03b598e9e34674d978c85d34/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/macos_codesign/tasks/main.yml for the original content.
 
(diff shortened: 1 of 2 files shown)


sxa (commented):
Happy with this in principle but it might be worth waiting until we're a little further down the Adoptium route before merging and deploying. We should also verify whether this will impact other teams using our stuff as we wouldn't want to risk destabilising others that might be relying on those certificates. Tagging @AdamBrousseau (I'm not sure if you're using any of the signing mechanisms that will be affected by this) for a review before accepting this change

AdamBrousseau (approved):
Thanks for the heads up Stewart. After review I think we should be fine with this change. We use a different mechanism now to sign our mac & win builds.

karianna:
Vagrant run is erroring on:

/opt/vagrant/embedded/gems/2.2.16/gems/vagrant-2.2.16/lib/vagrant/patches/net-ssh.rb:10:in `rsa_compat_build_request': undefined method `[]' for nil:NilClass (NoMethodError)

Haroon-Khel:
The vagrant failure is on the Solaris check. I wouldnt think it would be associated with this pr. Either way, ive kicked off another job

sxa:
> The vagrant failure is on the Solaris check. I wouldnt think it would be associated with this pr. Either way, ive kicked off another job

@Haroon-Khel What was the other job? Was it a re-run of the Solaris check or something else?

Haroon-Khel:
>@Haroon-Khel What was the other job? Was it a re-run of the Solaris check or something else?

A re run of the Solaris check. It still fails with the same error. As discussed in Slack, we agreed that the failure does not concern this pr. Is it safe to merge then?

gdams:
I'd say this is safe to merge. @sxa pleas approve and merge

sxa (approved):

PR #2204, opened 2021-06-08
inventory: add new solaris sparc machines

gdams:
Adds support for Solaris Sparc in our playbooks

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
@@ -15,25 +15,59 @@
   tags:
     - build_tools
 
-- name: Fetch CSWpkgutil.pkg
+- name: Fetch CSWpkgutil.pkg (x64)
   copy:
     src: files/CSWpkgutil.pkg
     dest: /tmp/CSWpkgutil.pkg
     force: yes
-  when: pkgutil_installed.rc != 0
+  when:
+    - pkgutil_installed.rc != 0
+    - ansible_architecture == "i386"
   tags: build_tools
 
-- name: Install pkgutil
+- name: Install pkgutil (x64)
   shell: |
     yes | pkgadd -d /tmp/CSWpkgutil.pkg CSWpkgutil
     /opt/csw/bin/pkgutil -U
-  when: pkgutil_installed.rc != 0
+  when:
+    - pkgutil_installed.rc != 0
+    - ansible_architecture == "i386"
+  tags: build_tools
+
+# The ansible copy command does not work at this point because it can't find source in the default shell
+- name: Install pkgutil (sparcv9)
+  shell: |
+    yes | pkgadd -d http://get.opencsw.org/now CSWpkgutil
+    /opt/csw/bin/pkgutil -U
+  when: 
+    - pkgutil_installed.rc != 0
+    - ansible_architecture == "sun4v"
   tags: build_tools
 
 - name: Install Build Tool Packages
   pkgutil: "name={{ item }} state=present"
   with_items: "{{ Build_Tool_Packages }}"
   tags: build_tools
+
+- name: Setting Solaris Studio variables (x64)
+  set_fact: 
+    SolariStudioArch: "x86"
+    DownloadURL: "https://ansiblestorageadopt.blob.core.windows.net/solarisstudiox64/SolarisStudio12.3-solaris-x86-pkg.tar.bz2?{{ SolarisStudio12_solaris_x86_SAS_TOKEN }}"
+  when: ansible_architecture == "i386"
+  tags:
+  - build_tools
+  - solaris_studio
+  - adoptopenjdk
+
+- name: Setting Solaris Studio variables (sparcv9)
+  set_fact: 
+    SolariStudioArch: "sparc"
+    DownloadURL: "https://ansiblestorageadopt.blob.core.windows.net/solarisstudiosparc/SolarisStudio12.3-solaris-sparc-pkg.tar.bz2?{{ SolarisStudio12_solaris_sparc_SAS_TOKEN }}"
+  when: ansible_architecture == "sun4v"
+  tags:
+  - build_tools
+  - solaris_studio
+  - adoptopenjdk
 
 - name: Check if Solaris Studio 12.3 is installed
   stat:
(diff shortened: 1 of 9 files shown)


gdams:
Also added the TCK machine to the JCK Jenkins in https://github.com/eclipse-cbi/jiro/pull/166

gdams:
@sxa could you please check this one :)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -15,26 +15,67 @@
...
   shell: |
     yes | pkgadd -d /tmp/CSWpkgutil.pkg CSWpkgutil
     /opt/csw/bin/pkgutil -U
-  when: pkgutil_installed.rc != 0
+  when:
+    - pkgutil_installed.rc != 0
+    - ansible_architecture == "i386"
+  tags: build_tools
+
+# The ansible copy command does not work at this point because it can't find source in the default shell
```
sxa, line 37: Can you clarify this? What is it that's trying to use `source` instead of `.`?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -86,6 +126,33 @@
     - adoptopenjdk
   when: not solaris_studio.stat.exists
 
+- name: Check if freetype is installed (sparcv9)
```
sxa, line 129: Can you add a comment about why this is only for SPARC? Seems odd that it wouldn't be in pkguntil the repositories if it is for x64, so we should make it clear what the reason is.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -123,6 +190,36 @@
     - build_tools
     - cups
 
+- name: Check if cups-1.5.4-source is present (sparcv9)
```
sxa, line 193: We seem to download, build and install 1.4.4 above. Why do we need both versions? If it's required, please add an explanatory comment for the future.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/maven/tasks/main.yml
```diff
@@ -20,11 +20,7 @@
   tags: maven
 
 - name: Download Apache Maven v3.6.3 (Solaris)
-  get_url:
-   url: https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
-   dest: /tmp/
-   checksum: sha512:c35a1803a6e70a126e80b2b3ae33eed961f83ed74d18fcd16909b2d44d7dada3203f1ffe726c17ef8dcca2dcaa9fca676987befeadc9b9f759967a8cb77181c0
-   validate_certs: False
+  command: wget https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz -O /tmp/apache-maven-3.6.3-bin.tar.gz
```
sxa, line 23: Can we do a SHA check afterwards? Otherwise it looks a bit odd to be removing the checksum verification here.

sxa (changes requested):
Comments above, plus despite the checkbox suggesting that it's been added to bastillion i can't seem to connect with my normal key - can you check and fix that please? :-) I'll also tag @Willsparker so he can double check some of the nagios stuff.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: So we use the ansible copy module on x64 (as seen [here](https://github.com/adoptium/infrastructure/pull/2204/files#diff-5a3d2646edcdff57d8dcd180e791e61aa1c56e9a66bf853127032828f635b30cR18-R26)), but on Sparc the default shell that's installed doesn't seem to have `source` as a built-in command)

e.g

```bash
which source
no source in /usr/sbin /usr/bin /opt/csw/bin/
```

Whereas on my system:

```bash
which source
source: shell built-in command
```

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: done

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
gdams: done

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/maven/tasks/main.yml
gdams: I switched to wget because of the usual SSL issues that we see on Solaris. I'm happy to add checksum validation back but there's a whole set of roles doing the same thing so this should likely be a follow up PR

gdams:
@sxa your key is on the machines, please try again

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
sxa: Yeah you'd normally use `.` in place of source which is why I'm wondering where `source` is being invoked from. It would be disappointing if that was sometihng deep within ansible that we can't control.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml
```diff
@@ -190,6 +191,7 @@
     - build_tools
     - cups
 
+# On Sparc we (for an unknown reason) have to provide the source as an additional configure arguement. I only ever managed to get configure to successfully work with the combination of both CUPS versions
```
sxa, line 194:
```suggestion
# On Sparc we (for an unknown reason) have to provide the source as an additional configure argument. I only ever managed to get configure to successfully work with the combination of both CUPS versions
```

sxa (changes requested):

sxa (approved):
I thought I'd made another comment during my last review but I can't find it now and I can't remember what it was. We do need to get the solaris/x64 tests working cleanly again though as those will make it easier to determine if solaris changes generally will be causing problems.

Willsparker:
I know this is after-the-fact, but yep! The Nagios stuff looks good. I remember there was an issue that I can't currently find, that made reference to the fact that the Nagios home was different for Solaris, which has been reflected here, so all good! :+1:

PR #2223, opened 2021-06-17
unixPB: set /etc/hosts to use IPV4 address rather than ansible_host

gdams:
On Solaris where we use a jumpbox ansible_host is set to cloud.siteox.com rather than the IPv4 address so the machine is unable to resolve its hostname

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adopt_etc/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adopt_etc/tasks/main.yml
@@ -60,7 +60,7 @@
   lineinfile:
     dest: /etc/hosts
     regexp: "^(.*){{ ansible_hostname }}(.*)$"
-    line: "{{ ansible_host }} {{ inventory_hostname }}.{{ Domain }} {{ inventory_hostname }}"
+    line: "{{ ansible_default_ipv4.address }} {{ inventory_hostname }}.{{ Domain }} {{ inventory_hostname }}"
     state: present
   when:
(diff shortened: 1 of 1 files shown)


karianna:
Is that the right thing to do _globally_?

gdams:
> Is that the right thing to do globally?

I'll delegate to @sxa but I don't think there should be any issues here

Haroon-Khel:
The [change](https://github.com/adoptium/infrastructure/pull/1996) to use `ansible_host` instead of `ansible_default_ipv4.address` was because 2 scaleway machines, build-scaleway-ubuntu1604-armv7-1 and -2, had a similar problem; they couldn't detect their public ip addresses using `ansible_host`. Perhaps a separate task could be put in specifically for Solaris?

gdams:
@Haroon-Khel thanks for the reply, I'm wondering if we could perhaps have both lines?

@sxa do you know if we could do that in the `/etc/hosts` file or would that cause issues?

sxa:
We'd need to test it ...

gdams:
I'll do some investigation

gdams:
@sxaI have confirmed that having both lines in `/etc/hosts` doesn't cause any issues. This seems like the most comprehensive solution to cover both options

Haroon-Khel:
Ill test this a few times before merging

Haroon-Khel (approved):
> @sxai have confirmed that having both lines in /etc/hosts doesn't cause any issues. This seems like the most comprehensive solution to cover both options

Works as expected. If having both lines does not cause any hostname or ip resolve errors, then it shouldn't be a problem. I'll approve this

PR #2232, opened 2021-06-22
pbTests add win playbook testing using WSL

gdams:
This will allow everyone to test the playbook using WSL (the Linux CL built into modern windows).  This also allows simpler testing of the Playbooks as we can then use WSL support in GH Actions. This allows Ansible to run 'natively'.

This PR also adds the `reboot` tag to tasks that require a windows restart to be run as well as a modification to the shortnames script to both add a registry key (required for win 2022) and also skip creating a shortname if the original directory doesn't exist.

##### Checklist

### DIFF
--- a/.github/workflows/build_wsl.yml
+++ b/.github/workflows/build_wsl.yml
@@ -56,4 +56,5 @@ jobs:
       run: |
         cd ansible
         export ANSIBLE_CONFIG=./ansible.cfg
-        ansible-playbook -i playbooks/AdoptOpenJDK_Windows_Playbook/hosts.win -b --skip-tags adoptopenjdk,jenkins,reboot playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
+        # Skip MSVS_2013 until https://github.com/adoptium/infrastructure/issues/2178 is fixed
+        ansible-playbook -i playbooks/AdoptOpenJDK_Windows_Playbook/hosts.win -b --skip-tags adoptopenjdk,jenkins,MSVS_2013,reboot playbooks/AdoptOpenJDK_Windows_Playbook/main.yml
--- a/.github/workflows/build_wsl.yml
+++ b/.github/workflows/build_wsl.yml
@@ -1,0 +1,59 @@
+name: Windows Playbook Checker
+
+defaults:
+  run:
+    shell: wsl-bash {0}
+
+on:
+  pull_request:
+    paths:
+    - .github/workflows/build_wsl.yml
+    - ansible/playbooks/AdoptOpenJDK_Windows_Playbook/**
+    branches:         
+    - master
+
+jobs:
+  build-wsl:
+    name: Windows 2019
+    runs-on: windows-latest
+    steps:
+
+    - name: Setup WinRM and Password
+      shell: powershell
+      run: |
+        Set-LocalUser -Name "runneradmin" -Password (ConvertTo-SecureString -AsPlainText "Ansible_password123!" -Force)
+        New-NetFirewallRule -DisplayName "ALLOW TCP PORT 5986" -Direction inbound -Profile Any -Action Allow -LocalPort 5986 -Protocol TCP
+        Invoke-WebRequest https://raw.githubusercontent.com/ansible/ansible/devel/examples/scripts/ConfigureRemotingForAnsible.ps1 -OutFile .\ConfigureRemotingForAnsible.ps1
+        .\ConfigureRemotingForAnsible.ps1 -CertValidityDays 9999
+        .\ConfigureRemotingForAnsible.ps1 -EnableCredSSP
+        .\ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert
+        .\ConfigureRemotingForAnsible.ps1 -SkipNetworkProfileCheck
+
+    - uses: actions/checkout@v2
+
+    - uses: Vampire/setup-wsl@v1
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt install -y ansible
+
+    - name: Setup Hosts file
+      run: |
+        cd ansible
+        rm -f playbooks/AdoptOpenJDK_Windows_Playbook/hosts.win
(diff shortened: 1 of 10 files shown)


sxa (commented):
I'm not in favour of adding more tests when we're failing to maintain the existing ones. The Solaris automated checks have been broken for a while and we've broken the VPC runs and not put the time into resolving them. We need to stabilise what we have before adding even more stuff for whic we won't put in the time/effort to maintain.

gdams:
> I'm not in favour of adding more tests when we're failing to maintain the existing ones. The Solaris automated checks have been broken for a while and we've broken the VPC runs and not put the time into resolving them. We need to stabilise what we have before adding even more stuff for which we won't put in the time/effort to maintain.

@sxa to beg to differ? Looking at the recent runs the Solaris builds are all passing, as are the linter fixes that I merged yesterday. I don't think there is such a thing as over-testing, especially considering the VPC runs can only be triggered by maintainers

karianna:
I think as a principle we should make sure the baseline is green before extending.  Sounds like in this case there's the one VPC PR to fix (which @gdams will look at today) or revert and then this should be OK to go again.

gdams:
@sxa is this good to be merged now?

sxa:
While I'm not going to block it (@karianna you are welcome to subject to my comments below if you wish) I'm not personally going to approve this as I don't want to take any responsibility for maintaining them once this is thrown over the wall. This really needs some of the following before it can even be considered IMHO:

1. An issue with a proposal as to why we want to do this with WSL
2. In the absence of a separate (1) some form of reasoning in the PR description
3. Whether we are suggesting that we gate a PR if it fails due to a specific issue with running under WSL (My feeling is no, since we do not currently support it anywhere in our processes, and most of our team does not use Windows so may struggle to debug a problem that occurs in that environment)
4. In light of (3), whether there will be people available to debug problems in a timely manner if and when this fails and issues are raised due to a non-obvious reason (perhaps whoever approves this will volunteer to monitor and triage failures?)

My personal view: Given the issues we have with VS2013 and VS2019 installs in the playbook at the moment, getting those to green would seem a more higher priority then adding this in (which currently excludes those installs and therefore actively masks underlying problems giving a false sense of success for a critical part of the build infrastructure)

karianna:
I think we should merge this, it's adding more test coverage of the 'state of what we have today' which is always a good thing.

Definitely don't disagree that we need to clean up the remaining VS items.  I'll have a think about how we can best facilitate that (I get that its naturally in MSFT's wheelhouse).

karianna (approved):
LGTM

karianna (changes requested):
Add some docs on how to test on WSL locally or on a remote machine to reproduce any issues that we might see in the GH Action.

sxa:
> when we're using the WSL support in GH Actions. 

Did I miss the issue/plan etc. for utilising this at Adoptium? We should ideally be putting new proposals for things that will affect people in issues and relating PRs like this to them wherever possible.

> We'll maintain the GH Action and work with any failing PRs

I'll hold you to that ;-)

sxa:
@AdamBrousseau I've just spotted that this adds `Community` to some of the path checks - is this likely to impact you if you're not installing the community versions on any of your systems?

AdamBrousseau:
Thanks for the heads up @sxa . I think we're safe on this one since we already have a diverged win playbook for Visual Studio that has the `Professional` paths.

gdams:
@sxa can we get this merged then please?

karianna:
https://github.com/adoptium/infrastructure/pull/2232/files for me is showing a bunch of ansible lint warnings and I think we're still missing docs for this PR.

SInce we're in a release week, does this need to land for that?

sxa:
Still awaiting docs.

aixtools (changes requested):
I see two tasks - getting tags added - where perhaps both tags could be added to one task, and the other task removed.

e.g., keep the roles/MVCS_2017 (but rename to roles/MVCS) and tag as both MVCS_2017 and MVCS_2019 (or one tag: MVCS)

- atm: I read it as unneeded duplicity

zdtsw:
isnt it much expensive to run windows (even WSL) in GH action than linux....

gdams:
> Are there any docs specific to running under WSL that we need?

I wouldn't have said so? It's just a Linux environment?

> Also I assume that since this is adding in a gating GitHub check on each PR that MSFT are taking responsibility for any problems that occur in this environment and will respond to any problems specific to WSL in that check?

This PR check is run against Windows playbook changes only so won't gate Linux playbooks etc. This change has come from myself (an Adoptium contributor). I'm happy to monitor changes etc but am not suggesting that MSFT should in any way take responsibility for this pull request or subsequent breaking changes.

gdams:
> isnt it much expensive to run windows (even WSL) in GH action than linux....

@zdtsw this action is only run against Windows changes (not Linux). There is no associated cost with running GitHub actions other than the fact that you are burning through Actions minutes (of which we have extra provided by MSFT and have never come close to running out of). Windows uses a 2x multiplier as opposed to Linux but this is no way near the 10x multiplier for macOS (see https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers)

sxa:
> > Are there any docs specific to running under WSL that we need?
> I wouldn't have said so? It's just a Linux environment?

If there's no difference running in WSL I'm happy with that answer.

> This PR check is run against Windows playbook changes only so won't gate Linux playbooks etc. This change has come from myself (an Adoptium contributor). I'm happy to monitor changes etc but am not suggesting that MSFT should in any way take responsibility for this pull request or subsequent breaking changes.

I assume the intention of putting it in GHA is to make it run on each PR and therefore gate (block) merging if a failure occurs. I don't think any other github checks are on things we'd definitely block on so the question is - if a change fails only on WSL will the onus be on the PR submitter to fix any such failures, or will you respond to them?

gdams:
@sxa I'm happy to help maintain the jobs, if the WSL element of things break then I'll work to fix it

sxa (approved):
> @sxa I'm happy to help maintain the jobs, if the WSL element of things break then I'll work to fix it

Thanks - SGTM on that basis

Issue #2091, opened 2021-03-28
Ansible request for git and cmake


sxa:
In the GIT_Source role we build and install git from source if the version on the default version of `git` on the machine is either too old or not present.

On some distributions (Such as Debian11) if the distribution's own `git` is new enough but not already present, we don't try to install it and still fall back to compiling it yourself. We should look at this and ensure we use the distribution's `git` where possible.

(same appears to be true for cmake)

Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [x] Missing install
- [ ] Bug in ansible playbook
- [ ] Request for new playbook addition

Details:

sxa:
@Haroon-Khel Are you able to take a look at this one. We seem to be wasting a lot of time during the ansible playbooks on the newer platforms which shouldn't need to build `git` `cmake` and possibly `ccache` as well which suggests the version checks aren't adequate (i.e. it's not properly detecting the right version that comes from the package respositories, so it's causing the playbooks to take longer than they need to on a brand new machine - `cmake` in particular seems to take ages to build from source).

Haroon-Khel:
@sxa Currently the Unix playbook installs the packages at these versions

```
Git 2.15
cmake 3.11.4
ccache 3.4.2
```
Do we still want to maintain these versions? Or are we comfortable with installing later versions of the packages. These specific versions may not be available on all package repositories

sxa:
Later versions are fine if they're available in the default repositories

Haroon-Khel:
Latest package versions offered by repos

Ubuntu

```
Git 2.30.2
Cmake 3.18.4
Ccache 4.2
```

Centos 6 repo

```
Git 1.7.1
Cmake 2.8.12
Ccache 3.1.6
```

Rhel 8 repo

```
Git 2.27.0
Cmake 3.18.2
Ccache 3.7.7
```

Rhel 7 

```
Git 1.8.3
Cmake 2.8.12
Ccache 3.7.7
```

Haroon-Khel:
For the OS's whose repositories offer a version of git, cmake or ccache lower than the version acquired via compiling, we can stick with compiling our own. Otherwise, it will be quicker to download from the repos. Still have to check Debian

karianna:
@Haroon-Khel can this now be closed?

sxa:
Much better - previously for Ubuntu 20.04:
```
13:49:01 TASK [Extract cmake] ***********************************************************
[...]

PR #2241, opened 2021-06-28
UnixPb: Install packages from repositories instead of compiling when necessary

Haroon-Khel:

ref https://github.com/adoptium/infrastructure/issues/2091

The unix playbook compiles Git, Cmake and Ccache to versions 2.15, 3.11.4, 3.4.2, respectively. On OS's whose default repos offer these packages at versions equal to or greater than the compiled versions, I have made it so that these packages are downloaded from there instead. See https://github.com/adoptium/infrastructure/issues/2091#issuecomment-869578801

The packages are installed from the repos in the `Common` role, which occurs before any of the roles in which these packages would otherwise be compiled, so theres no risk of duplicate installation. However, the `ccache` role checks for an already installed version of ccache by checking the `/usr/local/bin/ccache` directory. It needs to be made sure that the ccache downloaded from a repository installs to this directory.

Am in the process of making the changes to Debian

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
@@ -124,6 +124,13 @@
     - ansible_architecture == "armv7l"
   tags: build_tools
 
+- name: Install additional build tools for Ubuntu 20 +
+  package: "name={{ item }} state=latest"
+  with_items: "{{ Additional_Build_Tools_Ubuntu20+ }}"
+  when:
+    - ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= "20"
+  tags: build_tools
+
 #########################
 # Additional Test Tools #
 #########################
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Ubuntu.yml
@@ -126,7 +126,7 @@
 
 - name: Install additional build tools for Ubuntu 20 +
   package: "name={{ item }} state=latest"
-  with_items: "{{ Additional_Build_Tools_Ubuntu20+ }}"
+  with_items: "{{ Additional_Build_Tools_Ubuntu20 }}"
   when:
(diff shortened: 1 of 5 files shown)


karianna (approved):
Perhaps missing some comments as to why but apart from that LGTM

Haroon-Khel:
@karianna Just to save time in the playbook runs really. The compile stage of cmake and Git are significantly long

karianna:
> @karianna Just to save time in the playbook runs really. The compile stage of cmake and Git are significantly long

Worth commenting even that :-)

Haroon-Khel:
On ubuntu 21, the only vm I have right now without ccache installed, installing ccache from a repo does not create a `/usr/localbin/ccache` directory. So the check in the `ccache` role will have to be changed

Haroon-Khel:
Kicked off a vagrant playbook check job for Centos and Ubuntu https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1190/

Haroon-Khel:
Getting `11:39:33 fatal: [127.0.0.1]: FAILED! => {"msg": "template error while templating string: unexpected 'end of print statement'. String: {{ Additional_Build_Tools_Ubuntu20+ }}"}` on the ubuntu jobs. I've changed `Additional_Build_Tools_Ubuntu20+` to `Additional_Build_Tools_Ubuntu20` after some tests on a vm which seemed to fix it

Haroon-Khel:
Kicked off another VPC job https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1191/

Haroon-Khel:
The centos 7 job keeps failing because it can't find ccache to install from a repo. The package is supposed to be in the epel 7 repo, which the centos 7 vagrant machine does have installed as per 
```
17:26:43 TASK [Common : Enable EPEL release (not CentOS8)] ******************************
17:26:43 ok: [127.0.0.1]
```

Despite behaving as expected to the changes in this pr, the ubuntu 16 job fails due to 
```
17:02:48 TASK [Common : Install additional Packages specific to Ubuntu 16 (Not armv7l)] ***
17:02:50 fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "No package matching 'openjdk-7-jdk' is available"}
```
which is unrelated to this pr. The other Ubuntu and Centos jobs run fine

sxa:
> On ubuntu 20 installing ccache from a repo does not create a `/usr/local/bin/ccache` directory. So the check in the `ccache` role will have to be changed

It should do what the other ones do and check if any one that's already installed is >= the required version.

PR #2263, opened 2021-07-07
unixPB: fix up jckservices playbook

gdams:
##### Checklist

Coverts the old playbook into a role-based playbook to match the other unix playbooks. This also allows us to use the vendor secrets without code duplication.

Switches ftpd to vsfptd (as ftpd doesn't work anymore)


sxa (changes requested):
This is a lot more difficult to review than it could be given that things have been moved around so much. I'm not sure there is much reason to move this into the main playbook directly since it's a very specific separate service. Also, can the description of the PR be updated to indicate the "real" functional changes that are occurring in here please as it's currently empty.

It looks like all the firewall stuff has been lost too.

karianna:
Likely a separate concern, but I see in https://github.com/adoptium/infrastructure/pull/2263/files that GH does some extra analysis at the end on linting, do you also see that @gdams ?

sxa:
> Likely a separate concern, but I see in https://github.com/adoptium/infrastructure/pull/2263/files that GH does some extra analysis at the end on linting, do you also see that @gdams ?

Is that the Beta "Unchanged files with check annotations"? That's been around for a while now I think and I've (so far) been ignoring them, but I guess if we want to start trying to clear them up we should create an issue for it (Cant you think of anyone that might be around and able to work on something like that over the next few weeks @karianna? ) :-)

karianna:
> > Likely a separate concern, but I see in https://github.com/adoptium/infrastructure/pull/2263/files that GH does some extra analysis at the end on linting, do you also see that @gdams ?
> 
> Is that the Beta "Unchanged files with check annotations"? That's been around for a while now I think and I've (so far) been ignoring them, but I guess if we want to start trying to clear them up we should create an issue for it (Cant you think of anyone that might be around and able to work on something like that over the next few weeks @karianna? ) :-)

It is and great call - I'll set that up.

gdams:
@sxa the main reason for moving the playbook into the main Unix directory was because it's already setup with the role for fetching vendor files. Moving this into the other directory would mean a huge amount of code duplication. I also prefer the role-based setup as it's far simpler to maintain. I'll add the firewall rules back

sxa:
Why the change from `ftpd` to `vsftpd`?

sxa:
>  I also prefer the role-based setup as it's far simpler to maintain.

I think that's debatable for such a small piece of code as the jckservices setup (and the way this has done does pollute the "UNIX" playbook directory with stuff that most other users of our playbooks will never have a need to use) but we really should be proposing and discussing any such changes before putting such alterations into production (I presume this new version is the one that has been deployed?) or at the very least explaining the changes in the PR descriptions and not just via a PR title like this one.

gdams:
> Why the change from ftpd to vsftpd?

As far as I can tell, ftpd no longer works on ubuntu. We had similar problems internally.

sxa (approved):
There may be some functional issues with passive FTP based on using `jckservices.adoptium.net` as the `pasv_address` as it might end up with an internal address if it's not resolved over DNS, but as long as that is not a concern on the live system I'm ok with this going in, particuarly as it matches the current production system.

PR #1979, opened 2021-02-26
aixPB: Stop overwriting /usr/bin/perl with freeware on

sxa:
Draft PR to document what I've done for my own purposes if nothing else in case I forget :-)

By default AIX has `/usr/bin/perl` pointing to `/usr/opt/perl5/bin/perl5.28.1`. Our playbooks currently override this to the `/opt/freeware` one - this PR reverts that in order to use a standard AIX setup as far as possible. Particularly with the removal of perl as a requirement for the overall test frameworks I believe this may not be necessary. I have reset the symlink on [test-osuosl-aix71-ppc64-1](https://ci.adoptopenjdk.net/computer/test-osuosl-aix71-ppc64-1/) to see if it causes any problems - we need to make sure it doesn't before merging this (at which point we'll need to manually switch all the machines back to using the default. Some test Grinders as follows (although obviously it can run other stuff from the main pipelines) as it has the `ci.role.test` label:
- [sanity.system JDK8/J9](https://ci.adoptopenjdk.net/job/Grinder/7263)
- [extended.system JDK11/HotSpot](https://ci.adoptopenjdk.net/job/Grinder/7264)
- [sanity.openjdk JDK11/J9](https://ci.adoptopenjdk.net/job/Grinder/7284)

##### Checklist

- [ ] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md#commit-messages)
- [ ] [FAQ.md](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md) updated if appropriate
- [ ] other documentation is changed or added (if applicable)
- [ ] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

PR #2288, opened 2021-08-03
AixPB: Put /usr/bin at the front of PATH for AIX

Haroon-Khel:

Copy of https://github.com/adoptium/infrastructure/pull/2057, but that wont merge due to the owner of that pr not having eclipse authorisation.

I've also removed the `/etc/shells` addition from https://github.com/adoptium/infrastructure/pull/2057 since it was unrelated to the initial reason for the pr and was preventing it from being merged. It can be added later in another pr

The changes here have been tested already in https://github.com/adoptium/infrastructure/pull/2057

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -109,14 +109,18 @@
         path: /etc/security/login.cfg
         regexp: 'shells = '
         replace: 'shells = /bin/bash,'
-      tags: login_shell
+      tags:
+        - login_shell
+        - adoptopenjdk
 
     - name: Add bash to available login shells
       blockinfile:
         dest: /etc/shells
         block: |
           /bin/bash
-      tags: login_shell
+      tags:
+        - login_shell
+        - adoptopenjdk
 
     # move to role later
     - name: Set variables for global environment
@@ -126,14 +130,18 @@
           AIXTHREAD_HRT=true
           PKG_CONFIG_PATH=/opt/freeware/lib64/pkgconfig:/opt/freeware/lib/pkgconfig
           PERL5LIB=/opt/freemarker/lib/perl5
-      tags: login_shell
+      tags:
+        - login_shell
+        - adoptopenjdk
 
     - name: Add freeware and xlc to PATH for global environment
       replace:
         path: /etc/environment
-        regexp: 'PATH=/usr/bin'
-        replace: 'PATH=/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin'
-      tags: login_shell
+        regexp: 'PATH=/usr/bin:/etc'
+        replace: 'PATH=/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:/etc'
+      tags:
+        - login_shell
+        - adoptopenjdk
 
   ##############
   # freemarker #


sxa (commented):
@Haroon-Khel Have we tested this on any of the machines with some builds and test jobs to ensure nothing breaks as a result if picking up the AIX system tools by default? If so I'm happy to merge.

sxa:
See also https://github.com/adoptium/infrastructure/pull/1979 which adjusts the `perl` symlink in `/usr/bin` so we'd ideally want to verify both of these together.

Haroon-Khel:
> @Haroon-Khel Have we tested this on any of the machines with some builds and test jobs to ensure nothing breaks as a result if picking up the AIX system tools by default? If so I'm happy to merge.

This pr was tested in https://github.com/adoptium/infrastructure/pull/2057#issuecomment-804208932

sxa (approved):
I'm ok with this, but we need a plan for updating the existing machines as well since I don't think these regexs will adjust the existing machines.

aixtools:
:)

aixtools (commented):
OK, I have been playing around with how to do this in an idempotent way. See this as a template for what the change could be.
* key point is to 'slurp' the file, search the shells variable for bash and append it, if it is missing - do nothing if it already exists.
```
# Playbook to test slurp module/builtin
---
- hosts: localhost
  gather_facts: no

  pre_tasks:
    - name: Examine login.cfg
      slurp:
        src: /etc/security/login.cfg
      register: logincfg
    - name: extract shells setting in login.cfg
      set_fact:
        shells: "{{ logincfg['content'] | b64decode | regex_search('shells = (.*)') }}"
    - name: check for bash in shells setting
      set_fact:
        bash_in_shells: "{{ shells.find('bash') }}"

  tasks:
    - name: "Add bash to shells variable, if missing."
      block:
      - name: "Add bash to shells variable, if missing."
        replace:
          path: /etc/security/login.cfg
          regexp: '(shells = [^ \n]+)'
          replace: '\1,/usr/bin/bash,/bin/bash'
        tags: login_shell
      - name: "Add bash to /etc/shells"
        blockinfile:
          path: /etc/shells
          block: |
            /usr/bin/bash
            /bin/bash
        tags: login_shell
      when: bash_in_shells == "-1"
```

sxa:
Just occurred to me - why do we need `bash` in the available shells? Is it just a "nice to have" because we've installed it? Seems unrelated to putting /usr/bin at the start of the PATH

sxa:
> I'm ok with this, but we need a plan for updating the existing machines as well 

@Haroon-Khel Is this ready to go in? Have you put together such a plan yet, and if not can we make sure this is done please.

aixtools:
> Just occurred to me - why do we need `bash` in the available shells? Is it just a "nice to have" because we've installed it? Seems unrelated to putting /usr/bin at the start of the PATH

* bash is not needed as a user shell.
* bash is needed by gmake

aixtools:
> > I'm ok with this, but we need a plan for updating the existing machines as well
> 
> @Haroon-Khel Is this ready to go in? Have you put together such a plan yet, and if not can we make sure this is done please.

You want a plan? Just say do it - and -
- I'll manually edit /etc/environment.
- change all jenkins default shell to 'ksh'
- check AIX systems are idle and restart jenkins.

- the hardest bit will be tracking which nodes have been done - guess, rather than just re-cycle jenkins, reboot (then I can use `uptime` to see if they have been updated and re-cycled.

aixtools (changes requested):
* remove /opt/freeware/bin from PATH.
* having /opt/freeware/bin in, or not in, the PATH seems to have no effect on jenkins jobs (see adopt04 PATH).
* my expectation is that the build commands going to jenkins are modifying the PATH variable anyway (e.g., as for xlc16 jobs).
* Better to ALWAYS rely on the build-scripts to setup/verify the environment variables - rather than expect a variable to be provided by /etc/environment and/or /etc/profile

e.g., the current Adoptium/AIX PATH variable definitions:
```
root@nim.bak:[/home/root]dsh-adopt.ksh "echo \$PATH"
adopt01:
/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java6/jre/bin:/usr/java6/bin
==============
adopt02:
/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java6/jre/bin:/usr/java6/bin
==============
adopt03:
/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java8_64/jre/bin:/usr/java8_64/bin:/opt/bin
==============
adopt04:
/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java7_64/jre/bin:/usr/java7_64/bin
==============
adopt05:
/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java6/jre/bin:/usr/java6/bin
==============
adopt06:
/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java6/jre/bin:/usr/java6/bin:/opt/bin
==============
adopt07:
/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java6/jre/bin:/usr/java6/bin:/opt/bin
==============
adopt08:
/opt/freeware/bin:/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java8_64/jre/bin:/usr/java8_64/bin:/opt/bin
==============
```

sxa:
> > Just occurred to me - why do we need `bash` in the available shells? Is it just a "nice to have" because we've installed it? Seems unrelated to putting /usr/bin at the start of the PATH
> 
>     * bash is not needed as a user shell.
> 
>     * bash is needed by gmake



> > Just occurred to me - why do we need `bash` in the available shells? Is it just a "nice to have" because we've installed it? Seems unrelated to putting /usr/bin at the start of the PATH
> 
>     * bash is not needed as a user shell.
>     * bash is needed by gmake

Hmmm in what sense is bash explcitly needed by `gmake`? Even if we reuqire bash in some of the makefils I'm still unclear on why that would require it to be listed in the shells.

sxa:
Implementing this is going to have to wait at the moment - we've got a release starting next week and I don't want to risk any stability issues.

aixtools:
ok.

aixtools:
> > > Just occurred to me - why do we need `bash` in the available shells? Is it just a "nice to have" because we've installed it? Seems unrelated to putting /usr/bin at the start of the PATH
> > 
> > 
> > ```
> > * bash is not needed as a user shell.
> > 
> > * bash is needed by gmake
> > ```
> 
> > > Just occurred to me - why do we need `bash` in the available shells? Is it just a "nice to have" because we've installed it? Seems unrelated to putting /usr/bin at the start of the PATH
> > 
> > 
> > ```
> > * bash is not needed as a user shell.
> > * bash is needed by gmake
> > ```
> 
> Hmmm in what sense is bash explcitly needed by `gmake`? Even if we reuqire bash in some of the makefils I'm still unclear on why that would require it to be listed in the shells.

* bash is not needed in the shells
* the make scripts make use of special i/o redirects (fifo redirects, which is why we have the /tmp/sh* garbage) that only work when bash is used to execute them.
* Again - bash is not needed in the list of valid user shells (for the command `chsh`)

aixtools:
> Implementing this is going to have to wait at the moment - we've got a release starting next week and I don't want to risk any stability issues.

- Please assign to me. I'll be testing that /opt/freeware/bin is not even needed in PATH, not `bash` in `shells`.
- testing `outside` of osuosl btw.

aixtools:
You can leave this open, but as I am re-writing most of these things - and even have a plan for not needing to modify PATH in /etc/profile - I recommend - DO Not merge.

aixtools (changes requested):
Actually, my request is to not do anything, better, close the PR and tag won't fix' or something like that.

We are a year along, nearly, and I have a way to do this without modifying the global PATH environment variable.

sxa:
> and I have a way to do this without modifying the global PATH environment variable.

@aixtools what is your current planned solution? I believe we currently run into problems if the default `make` in the path is not GNU make so we'd need to resolve that everywhere we invoke it as a minimum before implementing such a change

aixtools:
a) use this to get back onto the program.
b) regarding make, my proposal will be to create an additional directory with commands that must come first, such as GNU `make`. And have this directory be first in PATH.
c) for GNU make, and other commands, as they arise - if any, would have symbolic links placed here to the regular command.

* Why? This way AIX and GNU packages can update themselves normally without being impacted. Trying to rename programs, e.g., rename AIX `make` to `make.aix` - to _disable/downgrade preference_ it may break after applying an update.
* If you see this as an acceptable solution, I'll get started with it.

sxa:
I'm wondering if we can get away without that. In the build scripts we have https://github.com/adoptium/temurin-build/blob/feb021cc605840022713485062a7cebe6f4be9b0/configureBuild.sh#L281 which should make it invoke `gmake` for AIX so the `PATH` issue disappears.

Which leaves the test jobs. There is already some handling of this in https://github.com/adoptium/aqa-tests/blob/b60d63cdff92c0ef8079cb5a537ceca28df81b42/buildenv/jenkins/JenkinsfileBase#L429 to force it to invoke `gmake`, and outside jenkins it's also included in [compile.sh](https://github.com/adoptium/aqa-tests/blob/b60d63cdff92c0ef8079cb5a537ceca28df81b42/compile.sh#L6)

@smlambert as far as you're aware does this mean there should no longer be a dependency on `make` being GNU make on AIX, since that appears to be what the lines referenced above should achieve - the only question would be whether they are invoked somewhere that doesn't use those definitions.

@aixtools If we can verify that some [examples of openjdk/sanity/perf runs](https://github.com/adoptium/infrastructure/blob/master/FAQ.md#how-do-i-replicate-a-test-failure) run ok on the machines with the AIX make in the path first and `gmake` somewhere later then this will probably make life a lot easier....

sxa:
> If we can verify that some [examples of openjdk/sanity/perf runs](https://github.com/adoptium/infrastructure/blob/master/FAQ.md#how-do-i-replicate-a-test-failure) run ok on the machines with the AIX make in the path first and gmake somewhere later then this will probably make life a lot easier....

> `adopt04:`
> `/opt/IBM/xlC/13.1.3/bin:/usr/bin:/etc:/usr/sbin:/usr/ucb:/usr/bin/X11:/sbin:/usr/java7_64/jre/bin:/usr/java7_64/bin`

Based on the absence of `/opt/freeware/bin` in the default path on these machines I've kicked off a couple of build+test runs on this machine ([test-osuosl-aix72-ppc64-2](https://ci.adoptium.net/computer/test%2Dosuosl%2Daix72%2Dppc64%2D2)) to see what happens to evaluate what state we are currently in:

JDK21: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-aix-ppc64-temurin/63/console
JDK8: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-aix-ppc64-temurin/470/console

aixtools:
what I recall as the proposed solution was to add PATH to the client environment - via a jenkins parameter.

Then the PATH does not impact system administration.

sxa:
> what I recall as the proposed solution was to add PATH to the client environment - via a jenkins parameter.
> Then the PATH does not impact system administration.

Yep. I think some of the scripts are probably explicitly adding it anyway so it may not even need to be across the whole of the jenkins client environment.

sxa:
The JDK21 job above seems to be running through ok on the machien without /opt/freeware/bin at the start of the path however the JDK8 one is less happy:

> https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-aix-ppc64-temurin/470/console

```

11:54:11  mk-ca-bundle.pl generates 148 certificates
11:54:11  Subject: CN=GlobalSign_Root_CA,OU=Root_CA,O=GlobalSign_nv-sa,C=BE
11:54:11  Generated alias: CN=GlobalSign_Root_CA,OU=Root_CA,O=GlobalSign_nv-sa,C=BE
11:54:11  Processing certificate with alias: CN=GlobalSign_Root_CA,OU=Root_CA,O=GlobalSign_nv-sa,C=BE
11:54:11  keytool error: java.lang.Exception: Input not an X.509 certificate
```
There is some logic in https://github.com/adoptium/temurin-build/blame/7dc73f3a9a78fee748c34be720193b50ea30cbbb/sbin/prepareWorkspace.sh#L649 which explicitly uses the keytool from the boot JDK. It looks like the above error occurs if this is set to a J9VM which is is on some machines such as adopt04 (test-aix-2) which is causing this problem. From the command line I was able to download a Temurin 8 into a temporary location to use as a boot JDK (consistent with the build machines) and it ran through ok :-)

karianna:
@Haroon-Khel will need a rebase

sxa:
@Haroon-Khel Do you think we can get this one in soon? That will give us a bit of runway before the next set of releases.

Haroon-Khel:
Uh using github's tool to rebase did not work out as I expected 😅

Haroon-Khel:
Continuing this in https://github.com/adoptium/infrastructure/pull/3716

Issue #2069, opened 2021-03-22
build-osuosl-aix71-ppc64-* failing jdk_lang jdk_util tests


sxa:
Both grinders on the build-osuosl machines still failed the following tests despite the symlinks being in place
```
jdk_lang_1
jdk_util_1
```
with the same error
```
LD_LIBRARY_PATH= : 	0509-022 Cannot load module /home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/lib/libawt_xawt.so.
13:52:59  	0509-150   Dependent module /opt/freeware/lib/libXrender.a(shr_64.o) could not be loaded.
13:52:59  	0509-152   Member shr_64.o is not found in archive
```

_Originally posted by @Haroon-Khel in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1697#issuecomment-804111475_

Haroon-Khel:
Related link 
https://www.ibm.com/support/pages/ibm-java-aix-how-resolve-unsatisfiedlinkerror-gui-based-application

sxa:
@Haroon-Khel See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1772#issuecomment-750373467 for potential things to try to reproduce the problem. That issue contains a small standlone java test case that might allow you to replicate it outside all the test harnesses.

aixtools:
Perhaps we need `ojdk06` in the grinder system - so that test results get captured for all - but only as a wildcard - so that different setups can be tested.

* When these tests fail - are they failing from something they just built? Or are they failing against a java image they just installed - meaning, was built on a different AIX system?
* If the failing java was not `just built` then the issue may have it's origin in the system config of the building system, not the config of the running system.

aixtools:
@Haroon-Khel your link above is for the unfortunate `user` that has to work with a JVM asis - and how to repair the AIX environment so that it can work.

IMHO: we are in a different situation: we are building the JVM and need to make sure that these issues do not occur rather than create a dependency on an AIX Toolbox (unsupported) package.

In the other issue: #1772 I guessed that this is being caused by a ./configure issue and a missing include file - and that libXrender-devel (plus it's requirements) gets installed to satisify a missing include file.

Looking further - it may be that this time - IBM does include the include file, but configure is not finding it:

```
p159a01:/ # find /usr /opt -name Xrender.h
/usr/lpp/X11/include/X11/extensions/Xrender.h
[...]

Issue #1772, opened 2020-12-21
(large j9core.dmp) test-osuosl-aix72-ppc64-1: error: copy-fd: write returned There is not enough space in the file system. + libXrender problem


lumpfish:
Test buids are failing on test-osuosl-aix72-ppc64-1 with out of space messages - e.g. https://ci.adoptopenjdk.net/view/Failing%20Test%20Jobs/job/Test_openjdk11_j9_extended.system_ppc64_aix/83/console
```
Running on test-osuosl-aix72-ppc64-1 in /home/jenkins/workspace/Test_openjdk11_j9_extended.system_ppc64_aix
[Pipeline] {
[Pipeline] checkout
The recommended git tool is: git
No credentials specified
 > git rev-parse --is-inside-work-tree # timeout=10
ERROR: Workspace has a .git repository, but it appears to be corrupt.
hudson.plugins.git.GitException: Command "git rev-parse --is-inside-work-tree" returned status code 128:
stdout: 
stderr: fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2450)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2380)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2376)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommand(CliGitAPIImpl.java:1923)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommand(CliGitAPIImpl.java:1935)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.hasGitRepo(CliGitAPIImpl.java:347)
	at hudson.plugins.git.GitAPI.hasGitRepo(GitAPI.java:235)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:90)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:55)
	at java.lang.reflect.Method.invoke(Method.java:508)
[...]

PR #2330, opened 2021-09-29
aixPB: remove unneeded libXrender-devel and related packages

aixtools:
Closes: #2069 
Closes: #1772 

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
@@ -134,13 +134,17 @@
     - libXft
     - tk
     - renderproto
-  tags: yum
+  tags:
+    - yum
+    - adoptopenjdk
 
 - name: Remove, if exists, reference to libXrender-devel include files
   file:
     state: absent
     path: /opt/freeware/include/X11/extensions
-  tags: yum
+  tags:
+    - yum
+    - adoptopenjdk
 
 #######################################################
 # Additional Tools not available directly through yum #
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
@@ -109,7 +109,6 @@
     - gawk
     - git
     - grep
-    - libXrender-devel
     - libffi-devel
     - make
     - m4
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
```diff
@@ -120,14 +119,29 @@
...
 
+- name: Remove packages - in case installed historically, or by accident
+  yum: name={{ item }} state=absent
+  with_items:
+    - libXrender-devel
+    - libXrender
+    - libXft
+    - tk
+    - renderproto
+  tags: yum
```
sxa, line 137: I don't want anything in place that could potentially damage a users' machine, so can we add the `adoptopenjdk`  tag to this so it is skipped when an end user uses the playbooks in the expected way please? I'm templted to suggest just running this as a one-off on all our machines rather than put it in here, but I'm ok with it satying too :-)

sxa (changes requested):

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
aixtools: OK, I added the tag - BUT - confused, because `adoptopenjdk` is not used anywhere else in this playbook. IMHO, the additional tag is misplaced. IMHO, if you are going to use this tag - that is a separate PR where either all playbooks are re-tagged, or a separate PR per playbook where re-tagging occurs.

Only adding it here, for these (two) plays adds confusion, imho.

And, for the record, there are likely several `stray` RPMs that are installed to satisfy a dependency - that was never needed. (e.g., `expect`, `tk` and `tcl` so that the yum.sh script can run).

sxa (approved):

aixtools:
FYI: manually applying it to systems adopt-01, 03 and 04 (05, 06, 07, 08 were built after the merge, 02 was used to test PR effectiveness).

- Before:
```root@nim.bak:[/home/root]dsh-adopt.ksh "rpm -qa | grep libXrender"
adopt01:
libXrender-0.9.8-2waixX11.ppc
libXrender-devel-0.9.8-2waixX11.ppc
==============
adopt02:
==============
adopt03:
libXrender-0.9.8-3waixX11.ppc
libXrender-devel-0.9.8-3waixX11.ppc
==============
adopt04:
libXrender-0.9.8-2waixX11.ppc
libXrender-devel-0.9.8-2waixX11.ppc
==============
adopt05:
==============
adopt06:
==============
adopt07:
==============
adopt08:
==============
```
After:
```
root@nim.bak:[/home/root]dsh-adopt.ksh "rpm -qa | grep libXrender"
adopt01:
==============
adopt02:
==============
adopt03:
==============
adopt04:
==============
adopt05:
==============
adopt06:
==============
adopt07:
==============
adopt08:
==============
```

PR #2343, opened 2021-10-08
UnixPb: Install curl version 7.58.0 and greater 

Haroon-Khel:

ref https://github.com/adoptium/temurin-build/issues/2751

Curl version 7.58.0* and greater is needed on our machines to solve the following certificate verification error

```
curl: (60) server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
```

*build-alibaba-ubuntu1804-armv8-1 has version 7.58.0 installed, this is the oldest version of curl on our build machines which does **not** experience the above certificate error

The Curl role gets run after the Common role, so if the OS's package manager does not install a version of curl greater than or equal to 7.58.0 then the Curl role will install the latest version, 7.79.1

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/curl/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/curl/tasks/main.yml
@@ -28,7 +28,7 @@
 
 - name: Download curl {{ curl_latest }}
   get_url:
-    url: https://curl.se/download/curl-{{ curl_latest }}.tar.gz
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
@sxa Done

PR #2352, opened 2021-10-12
UnixPb: Build Git using the correct curl directory

Haroon-Khel:

Since https://github.com/adoptium/infrastructure/pull/2343, on CentOS6 the version of curl which is installed is 7.79.1, so the correct curl directory must be passed when compiling Git


Haroon-Khel:
> Why is the install of `curl-devel` needed? If you're building `curl` from source it should already have all the development bits that are needed in `/usr/local` I'd have thought.

https://stackoverflow.com/questions/8329485/unable-to-find-remote-helper-for-https-during-git-clone suggests it is needed during the build for GIT. But, on the same docker image, as a test I uninstalled curl-devel and rebuilt GIT with the `--with-curl=/usr/local/curl-7.61.1` flag. It worked fine. So I'll remove the check for libcurl-devel

PR #2362, opened 2021-10-21
VPC: Stop using get.sh '-t' option in testJDK.sh

Willsparker:
As of this [PR](https://github.com/adoptium/aqa-tests/pull/2760) , `get.sh` can't use the `-t` option. This is causing all of the platforms that currently are able to build JDKs, to fail when testing the JDKs. According to the issue that PR is in reference to, `get.sh` should be run under `aqa-tests` anyway, so this should _hopefully_ fix the issue.

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) YUP: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1326/

### DIFF
--- a/ansible/pbTestScripts/testJDK.sh
+++ b/ansible/pbTestScripts/testJDK.sh
@@ -11,7 +11,9 @@ fi
 
 mkdir -p $HOME/testLocation
 [ ! -d $HOME/testLocation/aqa-tests ] && git clone https://github.com/adoptium/aqa-tests.git $HOME/testLocation/aqa-tests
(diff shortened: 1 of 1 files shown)


Willsparker:
All platforms that ran in VPC, have passed (at the time of writing, CentOS7 is still running the test, but it got through the issue that this PR fixes, so good enough). Merging now

Issue #2377, opened 2021-11-05
System unavailable: build-alibaba-ubuntu1804-armv8-1


sxa:
- Please put the system name in the title of this issue.

- Link to any log file showing the problem: https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-aarch64-dragonwell/ (Jobs not being run just now)

- Please describe the issue: Dragonwell aarch64 builds are locked to the alibaba machine at present. It is quite likely that these will run ok on the new Altra systems so we should test whether those builds work ok on there (they have additional prereqs such as targetted `armv8.2` which we weren't able to support on all machines previously.

sxa:
@joeyleeeeeee97 Do you know what has happened to the aarch64 alibaba build machine (`119.8.164.219`)? It doesn't seem to be contactable any more.

sxa:
(I've disabled the build job until this gets resolved one way or another)

sxa:
Two new systems now available. @Haroon-Khel is setting them up:
```
build-alibaba-ubuntu1804-armv8-2 	159.138.100.163
test-alibaba-ubuntu1804-armv8-2 	114.119.175.125
```

Haroon-Khel:
build-alibaba-ubuntu1804-armv8-2 is in jenkins. Build and tests running here https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-aarch64-hotspot/993/console

Haroon-Khel:
test-alibaba-ubuntu1804-armv8-2 is now in jenkins. Running an openjdk sanity test
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2561/console

sxa:
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-aarch64-dragonwell/158 has ran ok on the machine too and that job has now been re-enabled in jenkins so will be included in future pipelines. Thanks Haroon - closing.

PR #2391, opened 2021-11-23
inventory: add new alibaba machines to inventory

Haroon-Khel:

ref https://github.com/adoptium/infrastructure/issues/2377


sxa (approved):
I don't think we'll get the original ones back now, so it may be pointless calling these `-2`, but it's what they're called just now so I'm approving :-)

PR #2210, opened 2021-06-10
pbTests: Stop using find -maxdepth as it is nonstandard

sxa:
Signed-off-by: Stewart X Addison <sxa@redhat.com>

##### Checklist

- [x] commit message has one of the [standard prefixes](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/CONTRIBUTING.md#commit-messages)
- [ ] [FAQ.md](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/FAQ.md) updated if appropriate
- [ ] other documentation is changed or added (if applicable)
- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access): [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1178/)
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

This fixes:
```
12:44:54 find: bad option -maxdepth
12:44:55 find: [-H | -L] path-list predicate-list
```
and the resulting:
```
12:48:36 makefile:20: *** Please provide TEST_JDK_HOME value..  Stop.
12:48:42 makefile:20: *** Please provide TEST_JDK_HOME value..  Stop.
```
which shows up in the Solaris VPC checks

PR #2405, opened 2021-12-07
pbTests: Use 'homemade' Solaris box

Willsparker:
Ref: https://adoptium.slack.com/archives/C53GHCXL4/p1637309425055700

The Solaris box we were using didn't have any of the X11 libraries that is required to build a JDK. Therefore, with the help of @sxa and @gdams , I've made a new vagrant box; which I have installed to every `vagrant*` user on the VPC machine. This box has a lot more libraries located in `/usr/openwin/`, so, fingers-crossed it means we can actually build a JDK on there!

Creating this box has also allowed for it to be setup how we want it. Therefore, the compiler is pre-installed on the box so we don't have to do that whenever creating the VM, we have a newer version of SSH on there, so we don't have to change the ssh `KexAlgorithm`, and the VBox Guest Additions are up to date.

The `.box` file that was used to add the box to the machine, is located on the `infra-ibmcloud-vagrant-x64-1` machine, at `/home/will/solaris10_homemade.box`, in case any additional config needs to be put in. The PR is in draft as I want to get some working JDK builds before merging it :-) 

Please note; I've taken the `adoptopenjdkSol10.vm.hostname` argument out of the Vagrantfile, as this was causing the `/etc/hostname` file to be deleted. Deleting this file causes the network interface configuration command to fail, due to `sudo`'s `unable to resolve host` issue. The hostname that's been set on the box is `adoptSolaris10`, and I haven't seen any additional errors come up from removing the argument.


Willsparker:
As an update, as I found whilst looking at #2411 , the new box does have `/usr/xpg4/bin`, it's just not in the path. I've updated the vagrant boxes on the VPC machine to have `/usr/xpg4/bin` at the front of the path (by exporting `PATH` in `/etc/profile`). This was done as it most accurately reflects the older box.

Willsparker:
FYI, the latest commit will increase the timeout for the vagrant box. By default the timeout is 5 minutes, however apparently it is taking a bit longer on the VPC machine. If it still doesn't boot after 10 minutes, something is probably very wrong. Latest run: [VPC-1352](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1352/OS=Solaris10,label=vagrant/console)

Willsparker:
Okay, I've managed to get a JDK build in [VPC-1355](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1355/OS=Solaris10,label=vagrant/console) . This managed to successfully build JDK8/HS ; however, the tests weren't able to run, I believe due to the problem that is mentioned in #2210 . I also noticed that in the actual tests, were `grep` error messages:
```
19:55:23 /export/home/local/vagrant/testLocation/aqa-tests/TKG/scripts/getTestenvProperties.sh --repo_dir /export/home/local/vagrant/testLocation/aqa-tests/openj9 --output_file /export/home/local/vagrant/testLocation/aqa-tests/testenv/testenv.properties --repo_name OPENJ9
19:55:23 Check sha in /export/home/local/vagrant/testLocation/aqa-tests/openj9 and store the info in /export/home/local/vagrant/testLocation/aqa-tests/testenv/testenv.properties
19:55:23 grep: illegal option -- P
19:55:23 grep: illegal option -- o
19:55:23 Usage: grep -hblcnsviw pattern file . . .
```
Also, I'm not 100% certain the long boot times fo rthe VM are fixed. I was unable to recreate the issue on my machine, nor directly using the VPC machine, and it seems the boot time varies wildly, wherein it took >10 minutes to boot in [VPC-1353](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1353/OS=Solaris10,label=vagrant/console), and 3 minutes to boot in [VPC-1355](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1355/OS=Solaris10,label=vagrant/console)

Willsparker:
Testing running the tests, locally, I've determined a few things: we need to install `ant`, as that isn't in the playbooks for Solaris. However, having done that, I'm getting this problem when running the `ant -f scripts/build-tools.xml ...etc` command:
```
Buildfile: /export/home/local/vagrant/testLocation/aqa-tests/TKG/scripts/build_tools.xml
  [taskdef] Could not load definitions from resource net/sf/antcontrib/antlib.xml. It could not be found.
```
Googling has told me that this means `ant-contrib.jar` is not in the classpath. The lib is in `/usr/local/apache-ant-1.10.5/lib/ant-contrib.jar`, but apparently that isn't picked up.

In addition, some kind of perl `Digest SHA` package needs to be installed, according to this error:
```
getDependentLibs:
     [exec] Can't locate Digest/SHA.pm in @INC (@INC contains: /usr/perl5/5.8.4/lib/i86pc-solaris-64int /usr/perl5/5.8.4/lib /usr/perl5/site_perl/5.8.4/i86pc-solaris-64int /usr/perl5/site_perl/5.8.4 /usr/perl5/site_perl /usr/perl5/vendor_perl/5.8.4/i86pc-solaris-64int /usr/perl5/vendor_perl/5.8.4 /usr/perl5/vendor_perl .) at /export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/scripts/getDependencies.pl line 19.
     [exec] BEGIN failed--compilation aborted at /export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/scripts/getDependencies.pl line 19.
```
On something like RHEL, it'd be `perl-Digest-SHA.x86_64`, presumably there's a Solaris equivalent.

Willsparker:
Carrying on from that, it seems we need to install a newer version of ant:
```
BUILD SUCCESSFUL
Total time: 10 seconds
0
/export/home/local/vagrant/openjdk-build/workspace/build/src/build/solaris-x86_64-normal-server-release/images/jdk8u322-b05/bin/java -cp ./bin/TestKitGen.jar org.openj9.envInfo.EnvDetector
System.getProperty('java.vm.name')=OpenJDK 64-Bit Server VM

System.getProperty('java.vendor')=Temurin

Warning: cannot determine vendor, use System.getProperty('java.vendor')=Temurin directly.

System.getProperty('os.name')=SunOS

System.getProperty('os.arch')=amd64

System.getProperty('java.fullversion')=null

System.getProperty('sun.arch.data.model')=64

Error: required antVersion: 1.9.6. Installed version: 1.8.4
make: *** [makefile:73: envDetect] Error 1
```
The test scripts do carry on, beyond that point, however it skips the test:
```
===============================================
Running test MBCS_Tests_pref_ja_JP_linux_0 ...
===============================================
MBCS_Tests_pref_ja_JP_linux_0 Start Time: Wed Dec 22 11:35:29 2021 Epoch Time (ms): 1640172929744
Skipped due to jvm options ( ) and/or platform requirements (os.linux) => MBCS_Tests_pref_ja_JP_linux_0_SKIPPED
MBCS_Tests_pref_ja_JP_linux_0 Finish Time: Wed Dec 22 11:35:29 2021 Epoch Time (ms): 1640172929762
make[5]: Leaving directory '/export/home/local/vagrant/testLocation/aqa-tests/functional/MBCS_Tests/pref'
make[4]: Leaving directory '/export/home/local/vagrant/testLocation/aqa-tests/functional/MBCS_Tests'
make[3]: Leaving directory '/export/home/local/vagrant/testLocation/aqa-tests/functional'
make[2]: Leaving directory '/export/home/local/vagrant/testLocation/aqa-tests'
cd "/export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/scripts"; \
perl "resultsSum.pl" --failuremk="/export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/failedtargets.mk" --resultFile="/export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/output_16401729292614/TestTargetResult" --platFile="/export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/resources/buildPlatformMap.properties" --diagnostic=failure --jdkVersion=8 --jdkImpl=hotspot --jdkVendor="Temurin" --spec=sunos_x86-64 --buildList=functional --customTarget= --testTarget=MBCS_Tests_pref_ja_JP_linux_0 --tapPath=/export/home/local/vagrant/testLocation/aqa-tests/TKG/../TKG/output_16401729292614/ --tapName=


TEST TARGETS SUMMARY
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
TOTAL: 1   EXECUTED: 0   PASSED: 0   FAILED: 0   DISABLED: 0   SKIPPED: 1
ALL TESTS PASSED
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

```

sxa:
Hmmm I would have thought we'd be pulling ant 1.10.5 into /usr/local/ like we do elsewhere ... Unless we've got two versions installed and it's picking up the wrong one by default.

Willsparker:
Ah yes, I've found it; you were right. Installing it via PKGUtil just puts it in `/opt/csw/bin`. Alright, so I need to put `/usr/local/bin` into the path on Solaris as it isn't there at all. I'm going to put this in as another 'special case' in `testJDK.sh`; the alternative is to put a task somewhere that adds `/usr/local/bin` into the path for Solaris, however I wouldn't want to change a person's environment that much, without putting the `adoptopenjdk/jenkins`  tag on the task - which VPC skips anyway.

Willsparker:
Testing that locally, (i.e. adding to the path `/usr/local/bin`), It seemed to work well, however it still skipped the test after having compiled all of of them: 
```
===============================================
Running test MBCS_Tests_pref_ja_JP_linux_0 ...
===============================================
MBCS_Tests_pref_ja_JP_linux_0 Start Time: Tue Jan  4 11:20:37 2022 Epoch Time (ms): 1641295237474
Skipped due to jvm options ( ) and/or platform requirements (os.linux) => MBCS_Tests_pref_ja_JP_linux_0_SKIPPED

...

TEST TARGETS SUMMARY
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
TOTAL: 1   EXECUTED: 0   PASSED: 0   FAILED: 0   DISABLED: 0   SKIPPED: 1
ALL TESTS PASSED
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

AFAIK, no special options have been put on the JVM, so I guess the test is skipping due to the platform. I wonder if there's an alternative for Solaris :thinking:

Willsparker:
Ref: The latest committ. As said by @sophia-guo , Solaris can't run the `MBCS_Tests_pref_ja_JP_linux_0` test. (See: [Slack comment](https://adoptium.slack.com/archives/C53GHCXL4/p1641311568115100?thread_ts=1641296204.114900&cid=C53GHCXL4)) . Instead I have made the JDK just run `_jdk_math`. While this isn't ideal as it doesn't test the perl requirements, it does test the JDK in some regard, and only took ~4 minutes locally, so it should suffice. (Though if anybody can think of any better tests, it can be changed).

Theoretically, with this, the VPC Solaris run _should_ now work - VPC Run: [VPC-1632](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1362/).

Willsparker:
It worked (!!!). Ready for review.

sxa (approved):
Nothing too controversial in here - just the result of a lot of debugging to figure out why things were failing :-)

I'd suggest we want some docs on the process of setting up the Vagrant machine somewhere too, but other than one quoting suggestion I'm ok for this to go in :-)

ansible/pbTestScripts/testJDK.sh
```diff
@@ -11,13 +11,28 @@ else
...
-$MAKE_COMMAND compile
-# Runs this test to check for prerequisite perl modules
-$MAKE_COMMAND _MBCS_Tests_pref_ja_JP_linux_0
+
+# Solaris runs a different test to Linux.
+# See: https://adoptium.slack.com/archives/C53GHCXL4/p1641311568115100?thread_ts=1641296204.114900&cid=C53GHCXL4
+if [[ $(uname) == "SunOS" ]]; then
+	export BUILD_LIST=openjdk
+	$MAKE_COMMAND compile
+	$MAKE_COMMAND _jdk_math
```
sxa, line 32: May be worth just going back to `_MachineInfo` if that test is Linux-specific, as we're just looking to do a quick sniff test to ensure the thing has built properly.

sxa (changes requested):
Let's go back to a simpler test instead of the whole `jdk_math` suite

ansible/pbTestScripts/testJDK.sh
Willsparker: Fixed! :-) It'll be much faster now, hopefully!

sxa (approved):
LGTM

Issue #2422, opened 2022-01-10
Ansible request to start using "changed_when: False" in the playbooks


sxa:
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [ ] Missing install
- [X] Bug in ansible playbook
- [ ] Request for new playbook addition

Details: Suggested by @Willsparker in https://github.com/adoptium/infrastructure/pull/2420#discussion_r781036367

Willsparker:
This is primarily for the times we use `shell` or `command` to query something- as it isn't changing the machine state, it should report as `ok`, not `changed`, however the `shell` and `command` modules will report `changed` by default. :-)

PR #2420, opened 2022-01-07
unixPB: Install gcc 10.3 into /usr/local/gcc10

sxa:
Fixes #2386 

Signed-off-by: Stewart X Addison <sxa@redhat.com>

##### Checklist

- [x] commit message has one of the [standard prefixes](https://github.com/adoptium/infrastructure/blob/master/CONTRIBUTING.md#commit-messages)
- [ ] [faq.md](https://github.com/adoptium/infrastructure/blob/master/FAQ.md) updated if appropriate
- [ ] other documentation is changed or added (if applicable)
- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1365) / [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/313)
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

Change to testVPC.sh was due to a bug when building later JDKs which made the process pick up the `static-libs` directory instead of the JDK, which causd a test failure (described in the comments below)

PR #2424, opened 2022-01-10
unixPB: Add changed_when fields to shell modules, in gcc roles

Willsparker:
Ref: #2422 

As per my comment in the [referenced issue](https://github.com/adoptium/infrastructure/issues/2422#issuecomment-1008772277), these queries are not changing the machine state, so they shouldn't report `changed`. 

I prioritized these GCC roles, so they accurately reflect the new GCC-10 role from #2420.  

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_48/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/gcc_48/tasks/main.yml
@@ -6,6 +6,7 @@
   shell: /usr/bin/gcc-4.8 >/dev/null 2>&1
   ignore_errors: yes
   register: gcc_installed
+  changed_when: false
   tags:
     - gcc-4.8
     # TODO: write a condition when NOT to run this step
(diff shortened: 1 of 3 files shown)


sxa (commented):
Lets make sure this has gone through a VPC run first without problems

Willsparker:
Agreed- I haven't set it going just yet because we have 3 runs currently going on :-)

sxa (commented):
Hmmm this appears to be causing failures in `ansible_lint`
[EDIT: Probably not specific to the PR - raised https://github.com/adoptium/infrastructure/issues/2426]

Willsparker:
Linter is broken, as per your issue (#2426)

sxa (approved):
LGTM based on https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1382 :-)

(I'm assuming you've verified that it does the right thing when it's already installed i.e. on the second run of the playbook)

Willsparker:
Running [VPC-1386](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1386/) to check- entirely forgot about this, sorry :sweat_smile:

Willsparker:
Double checking that now :-) Thanks for running VPC

Willsparker:
Yep, it worked when I ran it twice. These roles now only report "ok" or "skipped", which is what we want :-)

Issue #2178, opened 2021-05-10
Ansible request to fix MSVS_2013 broken installation


Willsparker:
Ref:
- https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1160/
- https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/OS=Win2012,label=vagrant/1153/

- [ ] Missing install
- [x] Bug in ansible playbook
- [ ] Request for new playbook addition

Details:
Registering the MSVS_2013 shared libraries ([this task](https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/3ecc482cd9d34545676efecab0626882321385bb/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/MSVS_2013/tasks/main.yml#L26)), intermittently fails with :
```
10:59:23 TASK [MSVS_2013 : Register Visual Studio Community 2013 DIA SDK shared libraries] ***
10:59:26 failed: [127.0.0.1] (item=C:\Program Files (x86)\Microsoft Visual Studio 12.0\DIA SDK\bin\msdia120.dll) => {"ansible_loop_var": "item", "changed": true, "cmd": "regsvr32 /s \"C:\\Program Files (x86)\\Microsoft Visual Studio 12.0\\DIA SDK\\bin\\msdia120.dll\"", "delta": "0:00:00.093433", "end": "2021-05-05 09:57:05.619367", "item": "C:\\Program Files (x86)\\Microsoft Visual Studio 12.0\\DIA SDK\\bin\\msdia120.dll", "msg": "non-zero return code", "rc": 3, "start": "2021-05-05 09:57:05.525933", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```
(return code 3).

There's no stderr because regsvr32 is being run with `/s`, however that may help with debugging. See: https://devblogs.microsoft.com/oldnewthing/20180921-00/?p=99795 which may give some insight. 

According to above, RC=3 means the `loadlibrary` call is failing (or that it's a red herring). Either way, removing `/s` will give some insight, so that's where I'll start.

Willsparker:
Adding `/s` did not work, unfortunately, as it caused that task to hang - I'll have to try testing this locally..

Willsparker:
[...]

PR #2449, opened 2022-01-18
UnixPB: Use 'changed_when: false' when using shell tasks to query

Willsparker:
Ref: #2424 
Fixes: #2422 

Going through the Windows and Unix playbooks, whenever a `shell` module has been used to query something, this hasn't technically changed anything on the machine, however the task will report `changed`. This is incorrect, so I've gone through and whenever I've found an instance of this, I added the `changed_when: false` option. 

Technically, we should avoid using the `shell` module, as we could use `stat` in most instances, however for the time being, this is a good compromise. 

**Note:** While I can test the UnixPB with VPC (see below), Windows is currently broken on it; so the changes to those files will require _(shudder)_ manual reviewing.

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) [VPC-1390](https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1390/)


sxa (commented):
Hmmm I'm not really comfortable with stacking up Windows playbook changes that we can't test at the moment, so can we split them out into a separate PR for now until we have #2178 resolved? Also have you considered having an option to VPC that says "Run the playbook twice"?

Willsparker:
PR's have been split :+1:

Willsparker:
VPC passed! :partying_face:

aixtools (changes requested):
I am not going to take a specific example, but instead mention that _blanket_ changes of adding `changed_when: false` disable ansible state checking.

There are times where you know this is true - and it is acceptable (imho), just as I consider `ignore_errors: true` to be acceptable in some situations.

But to just paste it into a task to get rid of a false changed: status to get rid of the changed status is dangerous, and I would be surprised in yamllint doesn't come up with an objection to it (just as it has to ignore_errors: true).

Wherever possible use a condition that can be resolved. Or, if you come across that you are sure CAN be better, mark a task with a comment saying it needs work (# TBD: develop conditional for changed_when, or # rewrite module to use ansible module xxx: (e.g., as your PR description mention, stat:).

Examples (from AIX playbooks): wrong ansible mode/options used): 
```
# TBD: touch should be replaced by file:, stat: present, rather than touch)
./roles/syslogs/tasks/main.yml:
- name: Ensure log files exist
  file:
    path: "{{ item }}"
    state: touch
    mode: 0644
  with_items:
    - /var/log/syslog
    - /var/adm/messages
    - /var/log/messages
  changed_when: false
  tags: syslog
```
- Instead, review the command behavior and come up with proper changed_when: and failed_when: conditionals
```
# Using set -o pipefail -o to quiet AnsibleLint breaks execution on AIX
# Use skip_ansible_lint tag instead
- name: Remove conflicting packages before install yum
  shell: rpm -e --nodeps $(rpm -qa | grep -E "cloud-init|perl|openssl")
  register: _cleanup
  changed_when: "'no packages given for erase' not in _cleanup.stderr"
  failed_when: _cleanup.changed and _cleanup.rc != 0
  when: _yum.stat.islnk is not defined
  tags:
    - yum
    - rpm_remove
    # TODO: rpm used in place of yum or rpm_key module
    - skip_ansible_lint
```

Willsparker:
I agree, that we should probably go through these and change the modules to something more sensible, rather than `shell`. The `shell` commands that I've added `changed_when` to, have only been tasks that query the system (I've been careful not to add it to shell commands that 'do something', such as in your 2nd example). In this instance, `changed_when: false` is technically always true because it isn't changing the system.  

I think going through all of these commands and changing them to a more sensible module falls under #2492 , and that this PR isn't necessarily meant for that, but more as an interim thing :-)

aixtools:
> I agree, that we should probably go through these and change the modules to something more sensible, rather than `shell`. The `shell` commands that I've added `changed_when` to, have only been tasks that query the system (I've been careful not to add it to shell commands that 'do something', such as in your 2nd example). In this instance, `changed_when: false` is technically always true because it isn't changing the system.
> 
> I think going through all of these commands and changing them to a more sensible module falls under #2492 , and that this PR isn't necessarily meant for that, but more as an interim thing :-)

Actually, I have no problem with `changed_when: false`, just as I had no issue with `ignore_errors: true`. However, the people who design yamllint do have very strong opinions re: `ignore_errors: true`. I am only wishing to point out that an interim change - without a quick follow-up - may just bring about new issues.

If everyone can live with that - I am silent as well.

sxa:
@Willsparker Apologies for not getting back to this - do you think you can resolve the conflicts on this one so we can try and get it in?

Willsparker:
@sxa I'll put it on my to-do list tomorrow :-)

Willsparker:
Merge conflict resolved :+1:

karianna (approved):
LGTM

sxa:
Runnign through one final VPC before merging: https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1460/

Willsparker:
@sxa Seems to have worked okay? :)

sxa (approved):

Issue #2468, opened 2022-02-11
Ensure power management features are disabled on our systems


Haroon-Khel:
ref https://github.com/adoptium/infrastructure/issues/2464

test-packet-ubuntu2004-x64-2 recently had power management features enabled (sleep/hibernate) which caused the machine to suspend unexpectedly. We need to ensure power management features are disabled for our other machines

- [ ] Check to see if power management is enabled on existing systems
- [ ] Add playbook feature to ensure it is disabled 

Related commands

```
systemctl mask sleep.target suspend.target hibernate.target hybrid-sleep.target
gsettings set org.gnome.desktop.screensaver lock-enabled false
gsettings set org.gnome.desktop.lockdown disable-lock-screen true
gsettings set org.gnome.settings-daemon.plugins.power idle-dim false
```

sxa:
Looking at the last three of those - which of our servers have the GUI desktop enabled? That's likely just chewing up resources so it would be preferable to disable the desktop starting completely if it's present.

Haroon-Khel:
For ubuntu, I found that `systemctl set-default multi-user` followed by a reboot turns off the gui. I've tested this on a ubuntu virtual box vm whose gui I could visually see.

On test-packet-ubuntu2004-x64-2: 

Before systemctl set-default multi-user
root@test-packet-ubuntu2004-x64-2:~# systemctl status gdm3
```
● gdm.service - GNOME Display Manager
     Loaded: loaded (/lib/systemd/system/gdm.service; static; vendor preset: enabled)
     Active: active (running) since Thu 2022-02-10 17:12:01 UTC; 5 days ago
   Main PID: 1489 (gdm3)
      Tasks: 3 (limit: 9425)
     Memory: 7.0M
     CGroup: /system.slice/gdm.service
             └─1489 /usr/sbin/gdm3

```
After
```
● gdm.service - GNOME Display Manager
     Loaded: loaded (/lib/systemd/system/gdm.service; static; vendor preset: enabled)
     Active: inactive (dead)
```

Haroon-Khel:
Other distributions should have similar commands. I'll try to find them

Haroon-Khel:
Closing since https://github.com/adoptium/infrastructure/pull/2481 is merged

PR #2481, opened 2022-02-21
UnixPb: Turn off unnecessary desktop environment

Haroon-Khel:
ref https://github.com/adoptium/infrastructure/issues/2468

Right now running `systemctl set-default multi-user` only works on centos 7+ and ubuntu18+

Other distros should have similar commands which I will add to this pr shortly

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml
@@ -3,16 +3,26 @@
 # Ensure that no desktop environment is running #
 #################################################
 
+- name: Check if desktop is installed
+  stat:
+    path: /usr/share/xsessions
+  when: (ansible_distribution == "CentOS" and ansible_distribution_major_version >= 7) or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= 18)
+  register: desktop_installed
+  tags: disable_gui
+
 - name: Disable GNOME for CentOS 7+ and Ubuntu 18+
   shell: systemctl set-default multi-user
-  when: (ansible_distribution == "CentOS" and ansible_distribution_major_version >= 7) or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= 18)
+  when:
+    - desktop_installed.stat.exists
+    - (ansible_distribution == "CentOS" and ansible_distribution_major_version >= 7) or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= 18)
   register: gui_disabled
   tags: disable_gui
 
 - name: Reboot after disabling gui
   reboot:
     reboot_timeout: 600
-  when: 
-  - gui_disabled.stdout == ''
-  - (ansible_distribution == "CentOS" and ansible_distribution_major_version >= 7) or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= 18)
+  when:
+    - desktop_installed.stat.exists
+    - gui_disabled.stdout == ''
+    - (ansible_distribution == "CentOS" and ansible_distribution_major_version >= 7) or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= 18)
   tags: disable_gui
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml
@@ -23,6 +23,6 @@
     reboot_timeout: 600
   when:
     - (ansible_distribution == "CentOS" and ansible_distribution_major_version >= "7") or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= "18")
-    - gui_disabled.stdout == ''
     - desktop_installed.stat.exists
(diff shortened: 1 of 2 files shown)


Haroon-Khel:
Still not sure why this is failing the centos6  check

```
The error appears to have been in '/ansible/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml': line 21, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Reboot after disabling gui
  ^ here
The command '/bin/sh -c set -eux;  cd /ansible;  ansible-playbook -i hosts ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --skip-tags="debug,hosts_file,hostname,adoptopenjdk,jenkins,nagios,superuser,docker,swap_file,crontab,nvidia_cuda_toolkit";  ansible-playbook -i hosts ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --tags="riscv"' returned a non-zero code: 4
Error: exit status 4
```
I've run these changes on build-digitalocean-centos69-x64-2 and the tasks skip as they should

Haroon-Khel:
Rerunning cento6 check job https://github.com/adoptium/infrastructure/runs/5277532146?check_suite_focus=true

Haroon-Khel:
With the `adoptopenjdk` tag, it probably will pass the centos6 check. I will still test this pr on live machines before merging

sxa:
> With the `adoptopenjdk` tag, it probably will pass the centos6 check. I will still test this pr on live machines before merging

You can also test that in VPC since [it skips `adoptopenjdk`](https://github.com/adoptium/infrastructure/blob/9cefb9130f86dfff3299fd27ebb3dbebbe35efea/ansible/pbTestScripts/vagrantPlaybookCheck.sh#L244)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: Disable GNOME for CentOS 7+ and Ubuntu 18+
+  shell: systemctl set-default multi-user
+  when:
+    - (ansible_distribution == "CentOS" and ansible_distribution_major_version >= "7") or (ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= "18")
+    - desktop_installed.stat.exists
+  register: gui_disabled
+  tags: disable_gui
+
+- name: Reboot after disabling gui
```
sxa, line 21: Can this not be done using a systemctl command to switch it into console mode instead of a reboot?

Haroon-Khel: I use a systemctl command on line 14. From what I have tested, this then needs a reboot to make the changes effective. I have not yet found a single systemctl command which does it without the reboot

sxa: Have you tried `systemctl isolate multi-user.target`?

sxa:
> VPC kicked off https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1395/
> 
> I am still failing to understand why the centos6 check fails at the reboot task despite the `adoptopenjdk` tags

Might be worth creating a second temporary PR with a cut-down version of the playbook with your extra role and most of the other bits commented out (perhaps leaving in the JDK17 install and later bits) and see if that goes through the checks)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/disable_gui/tasks/main.yml
Haroon-Khel: I have just tried that on a centos8 virtual box vm. Gui immediately turns off, but is left in a 'blinking' console state. I am observing to see if this changes

sxa: Press ALT-F1, or Alt+F2, etc and see if that shows a login prompt at some point. What you're seeing is almost certainly "correct"

Haroon-Khel: alt-f1 brings up a login prompt, but I am struggling to think how to ansibile-ise this

sxa: There is no need to ansiblise the ALT-F1 bit. Just leave it as is.

Haroon-Khel: Ok, I've added a `systemctl isolate multi-user.target` task and removed the reboot one. However, using `isolate` to remove the gui will reset upon a reboot, restarting the gui. So I have kept in the `systemctl set-default multi-user`

Haroon-Khel:
Kicked off another VPC https://ci.adoptopenjdk.net/job/VagrantPlaybookCheck/1396/

Haroon-Khel:
https://github.com/adoptium/infrastructure/pull/2488 should hopefully fix the centos6 docker check

sxa (approved):
On the basis that the check failure as nothing to do with this PR I'm ok with this going in regardless :-)

PR #2512, opened 2022-03-18
UnixPb: Install bootjdk10 on macos

Haroon-Khel:

Unless someone can correct me, I see no reason as to why bootjdk 10 does not install onto MacOS via the playbooks. The build machines have it installed anyway.


Haroon-Khel:
I guess the Amr64 macos specific condition is no longer needed since we now have jdk11 arm64 macos binaries

sxa:
> I guess the Arm64 macos specific condition is no longer needed since we now have jdk11 arm64 macos binaries

This PR is adding JDK10 isn't it? We don't build that on aarch64 so any API call to download that will fail.

Haroon-Khel:
> This PR is adding JDK10 isn't it? We don't build that on aarch64 so any API call to download that will fail.

That's correct. Bad brain moment.

Either way, I assume we are then using a x64 jdk10 as a bootjdk for the jdk11 arm64 macos build? In which case, we would still need jdk10on build macos machines. The api should still call an x64 jdk10 binary instead of an aarch64 one

Haroon-Khel:
https://github.com/adoptium/infrastructure/blob/eaa0a88df1977c5c2e6a343ab1a86249e76aab57/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml#L142

Ansible will attempt to grab an aarch64 binary you're right. This will have to be changed for MacOS bootjdks

Haroon-Khel:
To my surprise we use a jdk11 binary as a bootjdk for arm64 jdk11 

`19:06:11  /Users/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-mac-aarch64-temurin/build-farm/../makejdk-any-platform.sh --clean-git-repo --jdk-boot-dir /Users/jenkins/jdk-11.0.15+8/Contents/Home `

From https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-mac-aarch64-temurin/24/consoleFull

sxa:
I'm not a fan of running it out of the jenkins home directory, so it would be good to adjust the boot JDKs to run out of the place the playbooks install them to.

Haroon-Khel:
https://api.adoptopenjdk.net/v3/installer/latest/10/ga/mac/x64/jdk/hotspot/normal/adoptopenjdk?project=jdk returns nothing it seems. That doesn't make any sense since we do have x64 jdk 10 binaries on our macos machines

Haroon-Khel:
Now it should download and install a binary for jdk10, instead of an installer, on both x64 and arm64 systems. The `unarchive` step needs root privileges since the binary is being extracted in `/Library/Java/JavaVirtualMachines/`

Haroon-Khel:
>How were the JDK10 binaries installed? 

Manually I think. On each machine, the jdk10 installation directory is owned by administrator, while the other jdks are owned by root (presumably because they were run by the installer which uses sudo)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
```diff
@@ -157,13 +157,32 @@
...
 
 - name: Run installer (macOS)
   shell: sudo installer -pkg /tmp/{{ jdk_version }}-installer.pkg -target /
   when:
     - ansible_distribution == "MacOSX"
     - adoptopenjdk_installed.rc != 0
+    - jdk_version != 10
+  tags: adoptopenjdk_install
+
+- name: Download binary when installer is not available (MacOS)
```
sxa, line 171: Can we change `binary` to tarball, since the installer is technically a binary too.
```suggestion
- name: Download tarball when installer is not available (MacOS)
```

sxa (approved):
I'm ok to approve this but it would be good to update the message as per the comment I've made.

PR #2541, opened 2022-05-05
feat: add gcc 11.2 to adoptopenjdk centos7 image

### LINKED ISSUE #2538: Ansible request for GCC 11.2 for OpenJ9
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

- [ ] Missing install
- [ ] Bug in ansible playbook
- [x] Request for new playbook addition

Details:

We need GCC 11.2 added to the [centos 7 build image](https://hub.docker.com/r/adoptopenjdk/centos7_build_image) and therefore need it added to the Ansible Playbooks. This is to support building jdk19+. We need it in the containers for x64 and aarc64. For s390x and ppc64le we do machine installs.

zdtsw:
Fixes #2538 

##### Checklist

- [X] [faq.md](https://github.com/adoptium/infrastructure/blob/master/FAQ.md) updated if appropriate

### DIFF
--- a/ansible/docker/Dockerfile.CentOS7
+++ b/ansible/docker/Dockerfile.CentOS7
@@ -2,8 +2,16 @@ FROM centos:7
 
 ARG user=jenkins
 
-RUN yum -y update; yum -y install epel-release
-RUN yum -y install ansible sudo; yum clean all
+RUN yum -y update && yum -y install \
+	epel-release \
+	centos-release-scl-rh # enable devtoolset repo
+RUN yum -y install \
+	ansible \
+	sudo \
+	devtoolset-11-gcc-c++ && \ # install gcc 11.2
+	yum clean all
+
+
 
 COPY . /ansible
 


sxa (commented):
I don't think using `centos-release-scl-rh` is a good idea as it is not available across all architectures. We should stick with the self-compiled GCC images as they will provide a consistent way of doing it across all required platforms.

zdtsw:
> I don't think using `centos-release-scl-rh` is a good idea as it is not available across all architectures. We should stick with the self-compiled GCC images as they will provide a consistent way of doing it across all required platforms.

good point!
will mark this as draft till we get confirmed (must use 11.2 than 10.3)

gdams:
Is there a reason you're modifying the dockerfile rather than the ansible playbook? I'd prefer to have this change made in the playbook if possible

zdtsw:
This draft is not needed any more (should have it closed already), 
Ithe ansible part to install gcc11.2 has been done and we are building centos7 with 11.2

PR #2546, opened 2022-05-09
github: updates workflow

zdtsw:
- use CentOS7 than CentOS6 for check: ansible 2.6.20 is not
match for newer modules, better use 2.9.X of ansible

##### Checklist

- [ x] commit message has one of the [standard prefixes](https://github.com/adoptium/infrastructure/blob/master/CONTRIBUTING.md#commit-messages)
- [ x] [faq.md](https://github.com/adoptium/infrastructure/blob/master/FAQ.md) updated if appropriate

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,20 +17,20 @@ on:
     - master
 
 jobs:
-  build-and-push-centos6:
-    name: Centos6
+  build-and-push-centos7:
+    name: Centos7
     runs-on: ubuntu-latest
     steps:
 
     - uses: actions/checkout@v2
 
-    - name: Docker Build & Push Centos6 Image to Docker Hub
+    - name: Docker Build & Push Centos7 Image to Docker Hub
       uses: docker/build-push-action@v1
       with:
         username: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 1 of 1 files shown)


sxa:
Can you explain what this change is for? The labels on this section all say centos6 and centos6 is the build environment we use so I don't see why we'd change this.

zdtsw:
> Can you explain what this change is for? The labels on this section all say centos6 and centos6 is the build environment we use so I don't see why we'd change this.

the reason to change to use CentOS7 as GH action check is because it has newer version of ansible 2.9.X than 2.6.X
which enable features in ansible to be tested out. 
e.g new ansible.builtin modules cannot be used if we use CentOS6. 

I should have changed all places in this workflow file: from CentOS6 to CentOS7 in this PR, but want to get an opinion if we can use CentOS7 first.

sxa:
We can't just stop building the CentOS6 docker images used for building most of our Linux/x64 builds. This change will stop that image from being updated so we won't get the playbook updates, so I can't approve this.

zdtsw:
> We can't just stop building the CentOS6 docker images used for building most of our Linux/x64 builds. This change will stop that image from being updated so we won't get the playbook updates, so I can't approve this.

Ok, I see. Then lets close this PR for now

PR #2657, opened 2022-07-05
unixPB: Add pigz on Alpine to speed up final archiving

sxa:
Brings Alpine in line with what we do on other Linux platforms - the build process will detect and use pigz if it is installed as an alternative to gzip.

##### Checklist


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Alpine.yml
```diff
@@ -36,6 +36,7 @@ Build_Tool_Packages:
   - linux-headers
   - numactl
   - numactl-dev                   # OpenJ9
+  - pigz
```
karianna, line 39: Add comment as to why

karianna (changes requested):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/Alpine.yml
sxa: We don't do that for any of the other tools in here, and it's easily findable in the build scripts and used for all builds so i'm not sure what the value is of adding something here (and I guess you'd want it in the other OS files which install it). Surely if we want to go to that level in the playbooks we would want to go through a separate exercise of doing that for every package we use? Plus there is also an explanation in this PR covering it.

karianna: I think boy/girl scouting here is a good habit to get into and yes I do believe we broadly should comment each package as to why we're adding it :-)

karianna (approved):

gdams (approved):

PR #2775, opened 2022-10-06
ansible: Remove jinja name template errors

sxa:
We have a number of instances where this policy of requiring jinja templates in `name:` clauses does not make sense e.g. when a path name is used (e.g. `/home/{{ Jenkins_User }}/somedir`) and I have not seen a compelling reason for this to be enforced, therefore I propose disabling it.

Ref: https://ansible-lint.readthedocs.io/rules/name/

Similarly when running the checks on this PR it was noticed that there is another error showing up in some of the rules. The second commit in skips that one too (I tried moving to warn_list: as per the snippet below but that didn't seem to work and it still showed as an error). Removal of this as an error is consistent with the comments made upstream in https://github.com/ansible/ansible-lint/pull/2512

```
warn_list:
  - 'template-instead-of-copy' # Replaced upstream in https://github.com/ansible/ansible-lint/pull/2512
```

##### Checklist


karianna (approved):
LGTM

Haroon-Khel (approved):

PR #2791, opened 2022-10-18
How to update a host group name

julian55455:
How to update a host group name to the nagios core configurations

cc @steelhead31 @smlambert @sxa

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,3 +156,72 @@ of the potential security risks this presents.
 
 * During our quarterly maintenance window application patches will be applied manually.
 * When a critical or high severity vulnerability is announced patching will take place within the time frame stated above.
+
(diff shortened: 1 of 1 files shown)


steelhead31 (approved):

sxa (approved):
Since this is a doc change, and @steelhead31 is ok with it, I'm happy to approve too :-)
I would say that some of this is also in https://github.com/adoptium/infrastructure/wiki/Nagios%3A-Addition-and-Removal-of-Machines-Without-the-Playbooks#server-definition
Also was this PR to fix a particular issue? If so, it is good practice to put `Fixes <issue link>` into the issue - that way the related issue will be automatically closed once the PR is merged.

Haroon-Khel (approved):

PR #2855, opened 2022-12-13
Add perl-IPC-Cmd to Redhat/CentOS installs for OpenSSL v3 compile

jdekonin:
* compiling openssl v3 within container or system requires this package
* see https://github.com/eclipse-openj9/openj9/pull/14930


karianna (approved):
LGTM

sxa (approved):
@jdekonin I'm ok with this going in, but is an equivalent requird for non-RH/CentOS systems if people want to build on, say, Ubuntu?

jdekonin:
I compiled openssl v3 on a Ubuntu 20 s390x system without any added package requirements.  But I think it will depend on the o/s level.  Compilation requirements are not spelled out at https://github.com/openssl/openssl.git but if users encounter problem with compiling openssl I think the best direction would be to send them to that projects, although there would be nothing wrong with asking the questions in slack either. :)

sxa:
> I compiled openssl v3 on a Ubuntu 20 s390x system without any added package requirements. But I think it will depend on the o/s release & level. Compilation requirements are not spelled out at https://github.com/openssl/openssl.git but if users encounter problem with compiling openssl I think the best direction would be to send them to that projects, although there would be nothing wrong with asking the questions in slack either. :)

OK thanks for confirming. If you're reasonably confident it's covered I'm fine with that and will merge.

PR #2861, opened 2022-12-20
Workflow: Migrate QEMU playbook check to github workflow

Haroon-Khel:

Migrated QEMU playbook check functionality. Relevant docs here https://github.com/uraimo/run-on-arch-action

The gh action does not support riscv, so that will still need to remain in QPC.

This yaml file is alot easier to maintain than the entire QPC jenkins job, and its alot easier to add new arch distro combos

### DIFF
--- a/.github/workflows/build_qemu.yml
+++ b/.github/workflows/build_qemu.yml
@@ -30,7 +30,7 @@ jobs:
             distro: ubuntu18.04
           - arch: s390x
             distro: ubuntu18.04
-          - arch: riscv
+          - arch: riscv64
             distro: ubuntu22.04
           # - arch: armv7
           #   distro: jessie
--- a/.github/workflows/build_qemu.yml
+++ b/.github/workflows/build_qemu.yml
@@ -1,0 +1,54 @@
+name: Build QEMU
+
+on:
+  pull_request:
+    paths:
+    - .github/workflows/build_qemu.yml
+    - ansible/playbooks/AdoptOpenJDK_Unix_Playbook/**
+    branches:         
+    - master
+
+permissions:
+  contents: read
+
+jobs:
+  build_job:
+    # The host should always be linux
+    runs-on: ubuntu-18.04
+    name: Build on ${{ matrix.distro }} ${{ matrix.arch }}
+
+    strategy:
+      matrix:
+        include:
+          - arch: aarch64
+            distro: ubuntu18.04
+          - arch: aarch64
+            distro: buster
+          - arch: ppc64le
+            distro: ubuntu18.04
+          - arch: s390x
+            distro: ubuntu18.04
+          - arch: armv7
+            distro: jessie
+
+    steps:
+      - uses: actions/checkout@v2.1.0
+      
+      - name: Run on architecture
+        uses: uraimo/run-on-arch-action@v2.3.0
+
+        with:
+          arch: ${{ matrix.arch }}
+          distro: ${{ matrix.distro }}
+      
+      - name: Install Ansible
+        run: |
+          apt -y update
+          apt -y install ansible
+
+      - name: Run the UNIX playbook
+        run: |
+        echo "localhost ansible_user=runner ansible_connection=local" > ansible/hosts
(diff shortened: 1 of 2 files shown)


.github/workflows/build_qemu.yml
```diff
@@ -0,0 +1,54 @@
...
+            distro: ubuntu18.04
+          - arch: armv7
+            distro: jessie
+
+    steps:
+      - uses: actions/checkout@v2.1.0
+
+      - name: Run on architecture
+        uses: uraimo/run-on-arch-action@v2.3.0
+
```
gdams, line 39: remove the space here

.github/workflows/build_qemu.yml
```diff
@@ -0,0 +1,54 @@
...
+          distro: ${{ matrix.distro }}
+      
+      - name: Install Ansible
+        run: |
+          apt -y update
+          apt -y install ansible
+
+      - name: Run the UNIX playbook
+        run: |
+        echo "localhost ansible_user=runner ansible_connection=local" > ansible/hosts
```
gdams, line 51: spacing is out (needs another tab)

sxa:
Not sure this is the right approach to force a github action to do this on every PR - will this not end up in a situation where all changes will result in a significant (potentially 1 day+) turnaround for PR approvals ... And repeated for all subsequent pushes/review fixes? I'd personaly prefer to stick with the existing QPC implementation where it can be run by individuals as required, and also on a scheduled regular basis to check the state of the playbooks.

gdams:
> Not sure this is the right approach to force a github action to do this on every PR - will this not end up in a situation where all changes will result in a significant (potentially 1 day+) turnaround for PR approvals 

@sxa why not trigger it by adding a label? you could do something like this:

```yaml
on:
  pull_request:
    types: [ labeled ]

jobs:
  build:
    if: ${{ github.event.label.name == 'run-playbook-checker' }}
```

Haroon-Khel:
>It's an option but I'm not convinced there's a compelling advantage to running it that way unless we're fixated on running these things in GitHub

Its alot easier to add platforms; instead of creating a new disk, updating the QEMU bash script and the jenkins job, I can add 2 lines to the yaml file

Haroon-Khel:
Thats annoying. When one fails the rest get cancelled mid execution

Haroon-Khel:
Strange that it failed here 
```
  TASK [Debug : Display Debug Information] ***************************************
  fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible_default_ipv4' is undefined\n\nThe error appears to have been in '/home/runner/work/infrastructure/infrastructure/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Debug/tasks/main.yml': line 5, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n#####################\n- name: Display Debug Information\n  ^ here\n"}

```
Because all of the variables in debug default to undefined if they are undefined

Haroon-Khel:
I'm of the opinion that the Swap and Crontab roles should not run on non adoptium machines. They make changes to the machine that only we at adoptium need. Anyone else feel this way?

gdams:
> The gh action does not support riscv, so that will still need to remain in QPC.

@Haroon-Khel I've just pushed risc-v support to the action so you should be able to add that

Haroon-Khel:
QEMU workflows got triggered when I added the label to the pr. Works as expected 👍🏻

Haroon-Khel:
Looks like you have to remove and re add the label for subsequent pushes to the same pr

Haroon-Khel:
@sxa for review. This pr is at a stage where it can be merged

sxa (approved):
As discussed we still need the full  functionality of QPC but I'm ok to let this to go in for now.

sxa:
@Haroon-Khel Doesn't look like RISC-V has been added after the earlier comments.

steelhead31 (approved):
Looks good.

Issue #2753, opened 2022-09-27
Add Centos 6 docker build to centos7_docker_image_updater jenkins job


Haroon-Khel:
Currently our centos6 docker image is built and pushed from a github workflow, https://github.com/adoptium/infrastructure/blob/76d51f107ec24321ad498001d5b88474514518fe/.github/workflows/build.yml#L20 

It would make sense to add this to our centos7_docker_image_updater jenkins job as part of the docker images it builds

sxa:
NOTE: Any updates to this should be covered in a corresponding doc update to https://github.com/adoptium/infrastructure/blob/master/FAQ.md#what-about-the-builds-that-use-the-dockerbuild-tag

sxa:
@Haroon-Khel @steelhead31 it looks like we have consistent failures on the job after the related PR was merged on the CentOS6 part. Can one of you take a look at this please as we need to be sure that when PRs are merged we verify that they don't cause failures.

Haroon-Khel:
I'll take a look

Haroon-Khel:
Error log on latest build https://ci.adoptopenjdk.net/view/Failing%20Temurin%20jobs/job/centos7_docker_image_updater/331/execution/node/212/log/

```
+ docker push index.docker.io/adoptopenjdk/centos6_build_image:linux-amd64
The push refers to repository [docker.io/adoptopenjdk/centos6_build_image]
bc6a0138fc79: Preparing
513c80085c29: Preparing
54d345198942: Preparing
10506ac33692: Preparing
42efb8220fde: Preparing
14e739122a1b: Preparing
d4f51880a155: Preparing
aaa5621d7c01: Preparing
d4f51880a155: Waiting
aaa5621d7c01: Waiting
14e739122a1b: Waiting
denied: requested access to the resource is denied
```

Got a replica jenkins job which is just running the centos6 build at the moment https://ci.adoptopenjdk.net/job/centos7_image_update_haroon_test/

Haroon-Khel:
Might be related to the fact that the first build and push requires the image to be tagged https://stackoverflow.com/questions/41984399/denied-requested-access-to-the-resource-is-denied-docker

Rerunning my replica jenkins job with a tag stage
https://ci.adoptopenjdk.net/job/centos7_image_update_haroon_test/9/console

Haroon-Khel:
Still fails https://ci.adoptopenjdk.net/job/centos7_image_update_haroon_test/10/console

[...]

PR #2978, opened 2023-03-02
Temporarily remove the centos6 build from centos7_docker_image_updater

Haroon-Khel:
ref https://github.com/adoptium/infrastructure/issues/2753

We are seeing failures in the centos6 build of the centos7_docker_image_updater pipeline, https://ci.adoptium.net/job/centos7_docker_image_updater/. Seeing as I have no ETA on when I can get this fixed, I am temporarily removing it from the pipeline as it is holding up the Docker Manifest stage of that job

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,14 +3,6 @@ pipeline {
     stages {
         stage('Docker Build') {
             parallel { 
-                stage('CentOS6 x64') {
-                    agent {
-                        label "dockerBuild&&linux&&x64"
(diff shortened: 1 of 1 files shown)


sxa (approved):
LGTM, although as a point for the future I would recommend use `git revert <sha>` if you're doing a full undo of an existing change since it makes it a bit clearer to the reviewers that there's nothing else in there.

karianna (approved):

PR #2995, opened 2023-03-15
Docs: Add infrastrcuture map Miro pdf

Haroon-Khel:
Thought it would be best for these to be in the docs folder, and then a link to them on the main page. What do people think?


karianna (approved):
LGTM

sxa (commented):
Can you export these at the same resolution in `.gif` format instead of `.pdf`? I suspect the total size will be quite a bit smaller for this sort of image (Currently this will add about 1Mb to the repository which only has about 5½Mb in it which isn't a deal-breaker but it would be nice to avoid it)

Also, please try to avoid creating PR branches in the upstream infrastructure repository - use your own fork.

sxa:
@karianna My previous review still stands

Haroon-Khel:
> Can you export these at the same resolution in .gif format instead of .pdf? I suspect the total size will be quite a bit smaller for this sort of image (Currently this will add about 1Mb to the repository which only has about 5½Mb in it which isn't a deal-breaker but it would be nice to avoid it)

Miro does not give me the option to export to .gif

sxa:
For comparison (I screenshotted this on a 1080p display so not perfect) - these are about 150Kb in total.
I wonder if it would make more sense to include these on the wiki instead as a reference.

-rw-r--r--. 1 sxa sxa   32979 May 19 09:58  inframap-other.gif
-rw-r--r--. 1 sxa sxa   26283 May 19 09:58  inframap-dockerhost.gif
-rw-r--r--. 1 sxa sxa   65721 May 19 09:59  inframap-test.gif
-rw-r--r--. 1 sxa sxa   46403 May 19 09:59  inframap-build.gif

inframap-build
inframap-test
inframap-dockerhost
inframap-other

Issue #2908, opened 2023-01-31
Add GPG signature verification to adoptium binary downloads


Haroon-Khel:
Follow on from https://github.com/adoptium/infrastructure/issues/2233

- [x] The download of the adoptium and azul binaries need signature verification
- [x] Add sig verification features to windows playbook
- [x] Add sig verification features to AIX playbook

Haroon-Khel:
A bit of a problem on AIX. I've installed GnuPG2

```
Name        : gnupg2
Arch        : ppc
Version     : 2.2.35
Release     : 1
Size        : 34 M
Repo        : installed
From repo   : AIX_Toolbox
Summary     : A GNU utility for secure communication and data storage.
URL         : https://www.gnupg.org
License     : GPLv3+
Description : GnuPG is GNU's tool for secure communication and data storage.  It can
            : be used to encrypt data and to create digital signatures.  It includes
            : an advanced key management facility and is compliant with the proposed
            : OpenPGP Internet standard as described in RFC2440 and the S/MIME
            : standard as described by several RFCs.
            : 
            : GnuPG 2.0 is a newer version of GnuPG with additional support for
            : S/MIME.  It has a different design philosophy that splits
            : functionality up into several modules.
            : 
            : This package includes support for smart cards and S/MIME encryption
            : and signing.
```
I get the following error when trying to import keys

```
root@adopt10:[/root]gpg --keyserver keyserver.ubuntu.com --recv-keys A2115AE15F6B8B72
gpg: keyserver receive failed: A file or directory in the path name does not exist.
```
A lot of solutions claim the error lies in `/etc/resolve.conf`. I've added google's dns server to `/etc/resolve.conf` and still no luck.

@aixtools Any advice on GnuPG on AIX?

Haroon-Khel:
https://github.com/adoptium/infrastructure/pull/2999 is ready for review. In it i've added verification for jdk downloads made in the unix playbook

Haroon-Khel:
```
TASK [ANT : GPG Signature Verification of ANT] **********************************************************************************************************************************************************************
changed: [13.68.134.204] => {"changed": true, "rc": 0, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

This was the output, using the `script` ansible module, which is supposed to work on windows according to the docs. It should say FAILED but it isnt cooperating.

[...]

PR #3020, opened 2023-04-06
aixPb: Install GnuPG2 on Aix

Haroon-Khel:

Ref https://github.com/adoptium/infrastructure/issues/2908

Added a role to download and install GnuPG2 (v 2.0.30-2 as others do not seem to work properly (see https://github.com/adoptium/infrastructure/issues/2908#issuecomment-1460038373))

Added signature verification for Ant. Will add verification for Java binaries in a separate pr

~I have not yet tested this pr on an AIX box. I have tested the usability of the `package_signature_verification.sh` in a command line environment on an AIX box with GnuPG2 v 2.0.30-2 installed, but not these additions to the AIX playbook so do not merge until I have tested this. Nonetheless it can still be reviewed~ Has been tested and works fine


sxa (changes requested):
It should be possible to install a specific version through the ansible yum task - can we just use something like `GnuPG2-2.0.30-2` in the list of RPMs?
I haven't tried it, but something like that ought to work (Ref: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html)

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
```diff
@@ -252,3 +253,9 @@
   tags:
     - yum
     - adoptopenjdk
+
```
sxa, line 256: Any reason not to just invoke it as `gpg` when it's used here instead of using a symlink?

Haroon-Khel: The package_signature_verification.sh invokes it as gpg so id rather keep it consistent. The other versions of the package, the latest one for example, installs it with the symlink. Version 2.0.30 (the working version) keeps it as `/usr/bin/gpg2` oddly

Haroon-Khel:
I have now tested this pr, it works as it should. Ready for review/merge

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/ant/tasks/main.yml
```diff
@@ -18,11 +18,23 @@
         msg: "Ant installed, skipping download and installation"
       when: ant.stat.islnk is defined
 
-    - name: Download and extract ant
+    - name: Download Apache Ant
+      get_url:
+        url: https://archive.apache.org/dist/ant/binaries/apache-ant-1.9.9-bin.zip
```
karianna, line 23: Do we want the latest version of Ant?

karianna (changes requested):

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/ant/tasks/main.yml
sxa: In order to avoid any risk of inconsistency between systems I would suggest it is preferable to lock it and do upgrades of this tool in a more controlled manner. We fix the versions in many other places and if we wish to change the policy on it (similar to what we're trying to do with the compilers) we should raise and discuss as a separate issue to cover all instances of this in all platform playbooks.

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/yum/tasks/main.yml
sxa: I'd be a bit more tempted to adjust the shell script to handle both cases instead of having an untracked symbolic link in the system binaries directory. What would happen in the future if we tried to update to the latest version and this symlink was still in place (bearing in mind it wouldn't get removed along with the current version of the package)? Does this have the potential to cause is problems with future playbook deployments?

I won't block on the approach, however if we're going to stick with the symlink please add a comment above the section with an explanation even if it's just pointing to your comment e.g. `For why this symlink is necessary, see  https://github.com/adoptium/infrastructure/pull/3020#discussion_r1162721897`

Haroon-Khel: > What would happen in the future if we tried to update to the latest version and this symlink was still in place (bearing in mind it wouldn't get removed along with the current version of the package)? Does this have the potential to cause is problems with future playbook deployments?

It is similar to other package instalments such as `ant` where we manually make a `/usr/bin/ant` link https://github.com/adoptium/infrastructure/blob/558b23d2cc1436f658593b2d10dc444a88fb031a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/ant/tasks/main.yml#L28. 

In the case of a future update, I would suggest a general rule where we observe how we installed the former package via the playbook, and then undo any manual work if necessary before performing the update.

I'll add the comment

karianna (approved):

sxa (approved):
> It is similar to other package instalments such as ant where we manually make a /usr/bin/ant link 

ant is a little different as it's not installed via a package manager and is therefore not likely to fall foul of any updates we do. My concern with gnupg is that if we bump to a later on that tries to put something other than the symlink in place then it might fail.

But as I said, I'm not going to block on that - thanks for adding the comment. I now approve!

Issue #3049, opened 2023-05-05
GPG verification fails on some Mac machines


Haroon-Khel:
While running the playbook via awx

```
TASK [adoptopenjdk_install : GPG Signature verification (macOS)] ***************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NoneType: None
fatal: [test-macstadium-macos1015-x64-1]: FAILED! => {"changed": true, "msg": "non-zero return code", "rc": 2, "stderr": "Shared connection to 207.254.28.171 closed.\r\n", "stderr_lines": ["Shared connection to 207.254.28.171 closed."], "stdout": "gpg: directory '/var/root/.gnupg' created\r\ngpg: keybox '/var/root/.gnupg/pubring.kbx' created\r\n
gpg: error running '/usr/local/Cellar/gnupg/2.3.1_1/bin/dirmngr': terminated\r\n
gpg: failed to start dirmngr '/usr/local/Cellar/gnupg/2.3.1_1/bin/dirmngr': General error\r\n
gpg: can't connect to the dirmngr: General error\r\ngpg: keyserver receive failed: No dirmngr\r\n", "stdout_lines": ["gpg: directory '/var/root/.gnupg' created", "gpg: keybox '/var/root/.gnupg/pubring.kbx' created", "gpg: error running '/usr/local/Cellar/gnupg/2.3.1_1/bin/dirmngr': terminated", "gpg: failed to start dirmngr '/usr/local/Cellar/gnupg/2.3.1_1/bin/dirmngr':…
```
I have not looked too deep into this ie whether this affects all of our mac boxes

to do:

- Run on other mac boxes
- Run in a command line environment
- Run outside of playbook, ie with [package_signature_verification.sh](https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/Supporting_Scripts/package_signature_verification.sh) directly

Haroon-Khel:
This error does not show up on build-macstadium-macos11-arm64-1 and 2

Haroon-Khel:
The error occurs at the key import step

```
administrator@test-macstadium-macos1015-x64-1 ~ % gpg --keyserver keyserver.ubuntu.com --recv-keys "3B04D753C9050D9A5D343F39843C48A565F8F04B"
gpg: error running '/usr/local/Cellar/gnupg/2.3.1_1/bin/dirmngr': terminated
[...]

PR #3077, opened 2023-05-26
unixPb: upgrade p11-kit to its latest version on MacOS

Haroon-Khel:

ref https://github.com/adoptium/infrastructure/issues/3049#issuecomment-1564776559

Installing `p11-kit` explicitly will ensure it is upgraded to its latest version

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/MacOSX.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/MacOSX.yml
@@ -14,6 +14,7 @@ Build_Tool_Packages:
   - gnu-sed
   - gnu-tar
   - nasm # openj9 jdk13+
+  - p11-kit
   - wget
   - bash # OpenJ9 needs bash v4 or later
 


karianna (approved):

sxa (commented):
The comment I drafted on this seems to have got lost.
Presumably the earlier version has been installed as a result of it being a prereq for something else - could we get into this state (a version of p11-kit that's too old) on any new install or is this purely to fix existing machines that have an old version? And since we've got problems running brew on the older versions, is this ever going to get run on those systems?
We don't normally list every single package that we install, only the top level ones that include others so it would be slightly unusual to add this one explicitly.

Haroon-Khel:
> or is this purely to fix existing machines that have an old version?

It is more so this. I have fixed the issue on our existing 10.14 and 10.15 machines so I do not anticipate it being an issue in the future. I'll close https://github.com/adoptium/infrastructure/issues/3049 as it is complete

sxa:
Based on your last comment the likelihood of this situation recurring in the future is minimal, so I propose we cancel this PR since the fix has been applied to all affected machines.

Issue #3086, opened 2023-06-01
Various errors found in the aix playbook


Haroon-Khel:
`yum update` dependency error
```
--> Finished Dependency Resolution
Error: Package: openldap-2.5.12-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: sudo-1.9.13p2-1.ppc (AIX_Toolbox)
           Requires: libssl.a(libssl.so.1.1)
Error: Package: curl-8.0.1-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: python3-3.7.15-2.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: python-pycurl-7.43.0.5-1.ppc (AIX_Toolbox)
           Requires: libssl.a(libssl.so.1.1)
Error: Package: krb5-libs-1.20.1-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: python-2.7.18-4.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: curl-8.0.1-1.ppc (AIX_Toolbox)
           Requires: libssl.a(libssl.so.1.1)
Error: Package: git-core-2.38.3-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: libssh2-1.10.0-2.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: git-core-2.38.3-1.ppc (AIX_Toolbox)
           Requires: libssl.a(libssl.so.1.1)
Error: Package: sudo-1.9.13p2-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: python3-3.7.15-2.ppc (AIX_Toolbox)
           Requires: libssl.a(libssl.so.1.1)
Error: Package: rsync-3.2.7-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
Error: Package: python-pycurl-7.43.0.5-1.ppc (AIX_Toolbox)
           Requires: libcrypto.a(libcrypto.so.1.1)
[...]

PR #3087, opened 2023-06-02
aixPb: Fixes for Users task in aix playbook

Haroon-Khel:

ref https://github.com/adoptium/infrastructure/issues/3086

The `Set authorized key for jenkins user` should not be running on localhost (ie the ansible controller). It should run on the node, hence why it was failing.

```
TASK [users : Set authorized key for jenkins user] *****************************
fatal: [test-osuosl-aix715-ppc64-1 -> localhost]: FAILED! => {"changed": false, "msg": "Failed to lookup user jenkins: 
\"getpwnam(): name not found: 'jenkins'\""}
```

I also removed an unnecessary warning which had a condition that did not work anyway


Haroon-Khel:
Not sure, I'll remove the jenkins one too. The warning messages are unnecessary since to get the warning message to appear, the task must fail, but if the task fails the playbook wont get to the warning message 😅

karianna (approved):

aixtools:
The reason for the tests, and the warning message, and the reason for the "localhost" was because I have never had the key access that you guys have.

The `zeus` file is different - probably because "Superuser_Account" was not, or is not _Enabled_ - so I did not catch the differences.

Back to `jenkins` - and why the warning message is needed: so that someone NOT part of the ojdk/adoptium infrastructure can run these without them bring the system to a halt. I had tested jenkins with a "local" repository.

As such: it appears the the delegate to localhost may be wrong in the 'task' set authorized key.

IMHO: the warnings have a purpose (see above), but it is your call.

I'll rework above into my review.

aixtools (changes requested):
Warnings are for a reason. The idea here is that not everyone is adoptium - and won't have the access or need to integrate with jenkins.

The three tasks are:
a) check if I have access to key (needs to be done localhost aka ansible master)
b) ensure jenkins authorization is present IF (when:) I have access to the key
c) warn about jenkins (but let playbook proceed) when ther eis no access to the key OR the insertion succeeded.

My preference is to keep the warnings - hence request changes. The goal is that both conditions are tested and if either fails - warn.

Haroon-Khel:
My personal preference, for users who do not have access to the keys and other vendor_files, is to skip running those roles entirely, that way we dont end up with half a role being run. 

One the reasons I removed the warnings is because their when condition did not work. `jenkins_pubkey.skipped` for example 
```
fatal: [test-osuosl-aix715-ppc64-1]: FAILED! => {"msg": "The conditional check 'jenkins_pubkey is defined and (jenkins_pubkey.skipped == true or jenkins_pubkey.failed == true)' failed. The error was: error while evaluating conditional (jenkins_pubkey is defined and (jenkins_pubkey.skipped == true or jenkins_pubkey.failed == true)): 'dict object' has no 
attribute 'skipped'\n\nThe error appears to be in 
'/tmp/awx_1319_9_q0jfjc/project/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/users/tasks/jenkins.yml': line 21, 
column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to 
be:\n\n\n- name: Warn about missing authorization for jenkins\n  ^ here\n"}
```

aixtools:
Probably also needs (needed) Jenkins_pubkey.defined. That gets really boring with ansible.

 

i.e. (not Jenkins_pubkey.defined OR Jenkins_pubkey.skipped). It’s been a long while since I tested these over and over and over again, but not as an Adoptium project member.

 


 ***@***.***>  

Michael Felt 

 


Mobile +31 (0)6 5184 4181 

Email ***@***.***

sxa:
> My personal preference, for users who do not have access to the keys and other vendor_files, is to skip running those roles entirely, that way we dont end up with half a role being run.

I think that's reasonable - typically someone without access to the adoptium secret data would be expected to run with `--skip-tags adoptopenjdk` (At some point we should probably switch that to `adoptium`, or update to allow both for now!)

sxa (approved):
Since the primary purpose of this is to remove the (incorrect) `delegate_to: localhost` I'm approving this so it can go in. Anything further from teh reviews can be covered in separate issues/PRs if desired.

PR #3095, opened 2023-06-12
Nagios: Fix Check Container Spaces Plugin

steelhead31:
Update the Nagios "Check Docker Container Spaces Plugin" to allow for null values being returned where the workspace directory does not exist, and also update the threshold comparisons to use numeric comparisons rather than string.

##### Checklist

Script has been deployed and tested locally using vagrant and virtual machines, and also on dockerhost-marist-ubuntu2204-s390x-


sxa:
What issues were caused by using `-gt` ?

steelhead31:
> What issues were caused by using `-gt` ?

@sxa  this was the error , which I think was highlighted by a container running without a workspace, causing a null value comparison.

```
./s.sh: line 25: [: -gt: unary operator expected
./s.sh: line 31: [: -gt: unary operator expected
```

Haroon-Khel (approved):

Haroon-Khel:
Approved on the basis that it resolves the error youre seeing

karianna (approved):

PR #3145, opened 2023-07-28
unixPB: Install rng-tools to fix low entropy

aswinkr77:
Cryptography tests are failing as not enough random information is being generated on older distributions, causing tests to timeout.
So installing and enabling rng-tools and its services fix the issues. 

##### Fixes:
 - https://github.com/adoptium/aqa-tests/pull/4474
 - https://github.com/eclipse-openj9/openj9/issues/16720

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -141,3 +141,4 @@
     - role: logs
       position: "End"
       tags: always
+    - rngd
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/rngd/tasks/fedora.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/rngd/tasks/fedora.yml
@@ -1,0 +1,19 @@
+---
+- name: Install the rng-tools package (Fedora)
+  package:
+    name: rng-tools
+    state: latest
+
+- name: Update the rngd.service file (Fedora)
+  ini_file:
+    path: /usr/lib/systemd/system/rngd.service
+    section: service
+    option: ExecStart
+    value: "/sbin/rngd -f -r /dev/urandom -o /dev/random"
+    backup: yes
+
+- name: Start and enable "rngd" service (Fedora)
+  service:
+    name: rngd
+    state: started
+    enabled: yes
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/rngd/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/rngd/tasks/main.yml
@@ -1,0 +1,22 @@
+---
+################
+#  rng daemon  #
+################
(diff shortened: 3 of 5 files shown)


karianna (approved):

AdamBrousseau:
Do we need another review on this or just a merge?
cc @Haroon-Khel

sxa:
Needs a second review first :-)

A couple of questions from me:
- Will these rules work ok in a docker container? or is it only applicable on the host of such a container?
- This won't be run on other distributions such as Debian/RHEL as-is - is that intentional?

aswinkr77:
> Needs a second review first :-)
> 
> A couple of questions from me:
> 
> * Will these rules work ok in a docker container? or is it only applicable on the host of such a container?
> * This won't be run on other distributions such as Debian/RHEL as-is - is that intentional?

- We've never tested this on a container cause we don't run these tests on containers.
- The PB will run on Ubuntu, Rhel, Cent and SLES, not Debian. We've never come across the entropy issue on a Debian machine, mostly it happened with the other four distros.

sxa:
I'm on vacation until Tuesday but w will need to verify that these changes don't cause as problem in containers. Our build are performed in containers built with the playbooks so it's critical that this won't fail to execute or break anything there.
I would hope that it works be ok, but I'm nervous enough to want to be be verified first 😉
Surprised it hasn't been deemed necessary on Debian though - I wonder if they're doing anything different 🤷

karianna (approved):

AdamBrousseau:
@sxa does one of those automatic checks test the docker compile? If not will Adopt have the capacity to test what you want to verify? I don't think we have any Debian machines so it is untested on our end and unknown if needed.

sxa:
> I don't think we have any Debian machines so it is untested on our end and unknown if needed.

Understood. Thanks for clarifying

> @sxa does one of those automatic checks test the docker compile? 

No but we have automated triggers that try to run a docker build from the updated playbooks when things are merged. Normally it's not something that would cause a problem but when we're starting services and doing stuff related to things in `/dev` it makes me nervous that a docker environment is a bit special in those respects. If you're able to run a `docker build` with at least one of the dockerfiles in https://github.com/adoptium/infrastructure/tree/master/ansible/docker using these changes (Probably a CentOS one as a preference) that would be good.

aswinkr77:
@sxa ran the PB on the centos6 image. Build openjdk11. Everything went fine.

sxa (approved):
> @sxa ran the PB on the centos6 image. Build openjdk11. Everything went fine.

OK That basis I'm ok with giving this a shot :-) My only comment would be that it might be worth putting the package installs into the common role where we already have a good abstraction for the distributions.

I'm in two minds about whether to request that we have the `adoptopenjdk` tag on this one since it could make a notable change to end users systems. If anyone else has a view on these feel free to jump in :-)

@andrew-m-leonard FYI - I don't /think/ this should affect build reproducibility but let me know if you want to perform any additional checks prior to merging.

andrew-m-leonard:
Sounds reasonable to me

sxa:
I'll plan to merge this tomorrow after the overnight build cycles in the absence of any objections.

Issue #2962, opened 2023-02-23
Define policy for performing updates (new prereqs) on DockerStatic test machines


sxa:
As alluded to in https://github.com/adoptium/infrastructure/issues/2954 we do not currently have a policy for ensuring that the DockerStatic containers used for running tests can be update when a new test prerequisite is required. This issue will cover identifying a recipe/process for ensuring that in the fairly rare case when a new dependency appears that we can handle adding it to all the DockerStatic containers that we have.

Haroon-Khel:
Could create a jenkins job which connects to each of the static docker nodes to update and install any packages we need. This would involve giving the jenkins user partial sudo access, enough to run install commands. This would be pretty easy to setup and maintain since the static docker nodes are all in jenkins.

An alternative is to modify this script, https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/scripts/updatepackages.sh, to update/install a specified list of packages. This script is currently used to run as a cron job on dockerhost machines to keep its hosted containers updated

sxa:
I don't want any of the jenkins agents to have elevated access. Doing so would allow any jobs to make modifications to the system and would be a major security concern.

As you suggest, that job is already performing the security updates - the additional concern here is that if we update the dockerstatic playbooks to add some new prerequisite the machines don't get updated so they end up out of date and potentially start failing tests.

sxa:
@Haroon-Khel Now that #3152 is in can you do an update to the TODO section of https://github.com/adoptium/infrastructure/blob/master/FAQ.md#dockerhost-todo to document the process we now have please?

sxa:
@Haroon-Khel Anything outstanding here now?

PR #3152, opened 2023-08-03
unixPb: run the updatepackage.sh script via dockerhost playbook

Haroon-Khel:

ref https://github.com/adoptium/infrastructure/issues/2962

Ive added a job in AWX to run the updatepackage.sh script regularly (will initiate it when this gets merged). This is better than it being in a cron job since AWX will pull updates to the script. It will run tasks in the `dockerhost.yml` playbook tagged with `updateContainers` which only corresponds to the task I have added in this pr. The `never` tag ensures that the task will only run when specified (by use of the `updateContainers` tag) and not as part of the entire playbook. 

This is still a draft because though this script updates the containers, it does not install new packages if needed, so I need to think of a clean way to do that. 

My last resort is to add strings in the `updatepackages.sh` script which contain any new package that needs to be installed on the containers (sort of how we do in the [Common](https://github.com/adoptium/infrastructure/tree/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars) role but as bash strings) but im certain theres a better way to do this.

Any thoughts @sxa @steelhead31

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/scripts/updatepackages.sh
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/scripts/updatepackages.sh
@@ -1,16 +1,22 @@
- #!/bin/bash
+#!/bin/bash
+set -u
 
 export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 containerIds=$(docker ps -q)
 
+commonPackages=
+fedoraPackages=
+debianPackages=
+alpinePackages=
+
 for container in $containerIds
 do
     OS=$(docker exec -it $container sh -c "cat /etc/os-release" | head -n 1)
-    if [[ $OS =~ "CentOS" || $OS =~ "Fedora" || $OS=~ "Red Hat Enterprise Linux" ]]; then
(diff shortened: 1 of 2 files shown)


steelhead31:
@Haroon-Khel I guess you could include a config file alongside the update packages.sh script itself, and have that script read in from the config file. I'm not entirely convinced its a better solution, but it would mean changes to which packages wouldn't involve changing the driving script, ultimately its not that different.

steelhead31 (approved):
LGTM!

Haroon-Khel:
Ive added some packages to be maintained/installed (depending on its presence). I think it would be overkill to add every package we install. I will continue to improve the script, but for now this works as a means to keep our containers updated (via the awx job) and to add new packages to containers when we need to

sxa (approved):
While it solves the immediate problem and I'm ok with letting this in for now I'm not sure this is the best way to go about it long term as we'll be adding to it every time we have a new one and end up with a partial duplication of package names here and in the docker files. Could we pull the list from the Dockerfiles?

I almost feel like we should be rebuilding the containers when a new package is added if possible.

(Also just as a point of note, the referenced issue https://github.com/adoptium/infrastructure/issues/2962 was to come up with the process/policy proposal(s) first - implementation was intended to be a separate thing ;-) )

PR #3153, opened 2023-08-07
Update ansible for AIX - X11 forwarding.

pstankie:
EF issue: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3430#note_1190148
ansible: Update /etc/ssh/sshd_config
ansible: force_handlers: True

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -11,8 +11,6 @@
     PERL_MB_OPT: "--install_base /opt/freeware"
     PERL_MM_OPT: "INSTALL_BASE=/opt/freeware"
 
-  force_handlers: True
-
   roles:
     - role: logs
       position: "Start"
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -11,6 +11,8 @@
     PERL_MB_OPT: "--install_base /opt/freeware"
     PERL_MM_OPT: "INSTALL_BASE=/opt/freeware"
 
+  force_handlers: True
+
   roles:
     - role: logs
       position: "Start"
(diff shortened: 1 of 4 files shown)


eclipsewebmaster (approved):
LGTM

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/security/tasks/main.yml
```diff
@@ -26,3 +26,25 @@
...
+      lineinfile:
+        path: /etc/ssh/sshd_config
+        regexp: '^#?X11DisplayOffset'
+        line: 'X11DisplayOffset 10'
+
+    - name: Make sure X11UseLocalhost is in sshd_config
+      lineinfile:
+        path: /etc/ssh/sshd_config
+        regexp: '^#X11UseLocalhost'
+        line: 'X12UseLocalhost yes'
```
sxa, line 50: Looks like this line is wrong:
```suggestion
        line: 'X11UseLocalhost yes'
```

sxa (changes requested):

ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
```diff
@@ -11,6 +11,8 @@
     PERL_MB_OPT: "--install_base /opt/freeware"
     PERL_MM_OPT: "INSTALL_BASE=/opt/freeware"
 
+  force_handlers: True
```
sxa, line 14: Can this be localised into the security role or does it have to be in the top level yml file?

pstankie: Handlers are triggered when plays succeed and due to role-based directory organisation there's no other place it can be put and since there some warnings and errors, the handler is unlikely to be run. Also, this is the first handler in the entire playbooks collection, so it's not affecting other handlers. We can move it to ansible-playbook cli, though as the option " --force-handlers"

sxa: To be clear, I'm OK with the original line in the top level yml or the documentation - just wanted to check if it was possible to move it to the lower level, andit sounds like that's a "no" :-)

sxa (approved):

PR #3160, opened 2023-08-11
ansible: Local FTP server on AIX 7.2 systems

eclipsewebmaster:
EF issue: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3503

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/group_vars/all/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/group_vars/all/main.yml
@@ -32,6 +32,7 @@ Security: Enabled
 
 # JCK Variables:
 jckftp_Username: jckftp
+jckftp_Groupname: jck
 
 # Jcktestr User Variables:
 Jcktestr_Username: jcktestr
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -105,4 +105,3 @@
 
     # 8. Add local ftp server and jckftp user
     - ftp
-
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/main.yml
@@ -102,3 +102,7 @@
     - role: logs
       position: "End"
       tags: always
+
+    # 8. Add local ftp server and jckftp user
+    - ftp
+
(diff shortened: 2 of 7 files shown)


karianna (approved):

sxa:
I'm not sure we want this in the main playbooks - only on the TC systems so perhaps it should be in the other repository?

gdams:
> I'm not sure we want this in the main playbooks - only on the TC systems so perhaps it should be in the other repository?

We should keep it in a public repo if possible, perhaps add a skip to this role if it's not a TCK machine?

sxa:
> > I'm not sure we want this in the main playbooks - only on the TC systems so perhaps it should be in the other repository?

> We should keep it in a public repo if possible, perhaps add a skip to this role if it's not a TCK machine?

That would be inconsistent with how we've done everything else. If we want to split TCK stuff like that then we should have a concerted effort to do that, but at the moment the TC-specific parts are in their own repository and I don't really want to deviate from that in this single instance under this PR.

fredg02:
@pstankie please close this PR and move the change to https://github.com/temurin-compliance/infrastructure instead. Please take care to use your own GitHub user to create PRs.

pstankie:
Just some background, why I put it here.

There's security role, which disables inetd: https://github.com/adoptium/infrastructure/blob/72dde1048268ff55df262e87cb7c44c4549258aa/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/security/tasks/main.yml#L7C3-L7C3

And the role I implemented enables inetd later again m but with changed configuration for ftp only.

Unlike temurin-compliance/infrastructure roles we don't install vsftpd here. This role only changes configuration which seems as logical continuation what other AdoptOpenJDK_AIX_Playbook roles do.

Moving to temurin-compliance/infrastructure would cause conflict between adoptium/infrastructure and temurin-compliance/infrastructure, as one would disable inetd and the other enable again. After the move, If we run AdoptOpenJDK_AIX_Playbook it will turn off inetd again.

Even with this one, under some tags combination, we will brake idempotency.

Having said that, please confirm i should move it anyway.

pstankie:
Maybe more elegant solution would be to leave inetd turned off (security role) and instal vsftpd on AIX, similarily  as it is done on other systems.

pstankie:
@sxa What is your opinion about that ? (I already talked to Fred).
There's also role: https://github.com/adoptium/infrastructure/tree/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/vsftpd
Which isn't included in playbooks, but it's there.

The other place vsftpd is being installed is in JCK_Playbook  in for containerized agents.

sxa:
Will send a DM and then summarise here :-)

pstankie:
I closed this PR according to discussions that took place in here, Slack and meeting.

Issue #2536, opened 2022-04-22
Looks at moving macstadium machines to orka


sxa:
I need to request a new machine:

- New machine operating system (e.g. linux/windows/macos/solaris/aix): mac
- New machine architecture (e.g. x64/aarch32/arm32/ppc64/ppc64le/sparc): x64
- Provider (leave blank if it does not matter): macstadium
- Desired usage: Replacement for the non-orka machine that we have to reduce costs
- Any unusual specification/setup required: Standard playbooks, although if there is an opportunity for more lower spec ones that could be beneficial
- How many of them are required: Start with 2, look at increasing.

Please explain what this machine is needed for:

sxa:
As per discussion a few weeks ago that the action is on me to progress, George and I will look at this migration together.

sxa:
Related: https://github.com/adoptium/temurin-build/issues/3354

sxa:
Our orka systems have been deprovisioned due to inactivity - currently having negotiations to determine a way forward.

sxa:
Discussions with MacStadium have indicated that an orka-based solution (which would not be sponsored at present) would be approximately twice the cost of the static systems which we have at present so we are looking at alternative options.

Here is a breakdown of the number of systems and their types we have at macstadium:

Use | x64 | aarch64
---|---|---
Build | 2xG3 (4core) | 2xG5G
Test | 6xG3B  (4core) 1xG4B (6core) | 2xG5A
TCK | 2xC3D (sml) 1xG4D (lge) | 2xG5E

So that's a total of 4+9+5 = 18 systems.
We currently have two hosted with MacInCloud with a potential option to increase that, particularly for x64 capacity

sxa:
Looking at the performance of various systems, here are some runs of the JDK8/x64 extended.openjdk suite on the different machines:

System | Time | Failures?
---|---|--
TC G4D | 2h28 | 17 (hostname issues) 3702
TC G3D - i5/2C/8G | 6h51 | Same hostname issues as G4D 3701
G3B - i7/4C/16G | 3h03 | [All passed](https://ci.adoptium.net/job/Grinder/7445/testReport/)
G3B - i7/4C/16G | 3h38 | [Three failures in java.nio](https://ci.adoptium.net/job/Grinder/7446/testReport/)
aarch64 (Rosetta) | 2h24 | [14 failures](https://ci.adoptium.net/job/Grinder/7447/testReport/)
[...]

PR #3185, opened 2023-09-12
unixPb: Add role to install intel homebrew for arm64 macs

Haroon-Khel:

ref https://github.com/adoptium/infrastructure/issues/2536#issuecomment-1708716478

~Need to test before merging~


Haroon-Khel:
Tested. Works as expected. Ready to merge

Haroon-Khel:
Going forward, any brew install ansible task which intends to run on arm64 mac needs to have its path specified to /opt/homebrew/bin/brew seeing as we will now have both intel and arm64 homebrew on our machines

Haroon-Khel:
If this gets merged now AWX will run the changes on our build machines, so i recommend it gets merged until after the coming release

karianna:
Why do we need Intel brew on Arm?

Haroon-Khel:
@karianna arm64 macs need the intel library `libpng` to be able to compile freetype for x64 jdk8 which is installed using intel homebrew

See https://github.com/adoptium/infrastructure/issues/2536#issuecomment-1708716478

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
```diff
@@ -25,12 +25,12 @@
 
 - name: Set homebrew path (x64)
   set_fact:
-    homebrew_path: /usr/local/bin/brew
+    homebrew_path: /usr/local/bin
```
karianna, line 28: will this leave old installs lying about?

karianna (approved):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
sxa: @karianna I believe we're ok. `homebrew_path` is being changed from the executable location to the directory that the executable was in. The only two previous references to this variable were the check for its presence and one invocation, which have been modified. All the others are new references and I believe the `/usr/local/bin` and `/opt/homebrew/bin` locations are the defaults so this should not conflict with existing installation. I'll be honest the use of the `/bin` in these paths looks a bit strange to me, but that is consistent with the [ansible docs](https://docs.ansible.com/ansible/latest/collections/community/general/homebrew_module.html#parameter-path).

karianna:
OK, so just checks to go green and this is good

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Homebrew_intel/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  stat:
+    path: /usr/local/Homebrew/bin/brew
+  register: intel_homebrew_installed
+
+- name: Run Homebrew install script in a Rosetta shell
+  ansible.builtin.shell: yes | arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+  when: not intel_homebrew_installed.stat.exists
+
+- name: Install Intel libpng
+  ansible.builtin.shell: arch -x86_64 yes | /usr/local/Homebrew/bin/brew install libpng
```
sxa, line 16: This is presumably the bit that caused the problems from https://github.com/adoptium/infrastructure/issues/3231 so I'm going to block this change on the basis that we presumably fix this PR before it goes live.

sxa (changes requested):

Haroon-Khel:
As per https://github.com/adoptium/temurin-build/issues/3509#issuecomment-1790995038, we do not need libpng installed (either arm64 or intel) on our build machines to be able to cross compile x64 mac jdk8 which means we do not need intel homebrew on our arm64 macs

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
```diff
@@ -162,6 +180,14 @@
   when: not pdfwriter.stat.exists
   tags: jck_tools
 
+- name: Uninstall libpng
```
sxa, line 183: I'm `-1` on this as I don't want end-users who use our playbooks to have things removed from their machine. Suggest we do this as a one off on all of our machines unless there is a specific reason for this to be in the playbooks. If it has to be, let's tag it with `adoptopenjdk`.

Also bear in mind that [we have a bug](https://github.com/adoptium/temurin-build/issues/3509) which is causing libpng to be picked up even when we tell it not to, which is not ideal :-)

sxa: (Now tagged with adoptopenjdk so objection removed but we should plan to remove this in the future if we can)

sxa (approved):
OK I think this is good to go in now :-) Thanks for the changes.

Issue #2121, opened 2021-04-13
QPC: All platforms are unstable


Willsparker:
Ref: https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/229/

With the latest QPC run, all platforms have failed to some degree. 4 have failed due calling `buildJDK.sh` in an incorrect way (incorrect as of #1962 ). 

The `risc-v` platform is still blocked by #1483 

And the `arm32` platform seems to be running out of space during the playbook execution.

Willsparker:
Changes made to `qemuPlaybookCheck.sh` have been made to use the correct arguments ([willsparker/2123_1](https://github.com/Willsparker/openjdk-infrastructure/tree/2123_1)). Testing on QPC at: https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/230/

Willsparker:
The above PR should fix the `buildJDK.sh` issues. Once this has been merged, I'll look at running the arm32 box, to determine if there's still issues with the space.

Willsparker:
The PR to fix the arguments has been merged, and it appears to be working on all platforms. For the S390x and ppc64le architectures, the `linux.sh` script fails for JDK8, as it is unable to find JDK7 on the machines (as Zulu-7 is not installed). I thought that they were meant to default then to using JDK8, but apparently not. This is not the case for JDK11, where I have had the build start on both machines (though currently, I've only managed a [QPC run](https://ci.adoptopenjdk.net/job/QEMUPlaybookCheck/247/) that had ppc64le fail (maybe due to the wrong version of gcc..?), and s390x that core dumped (maybe due to a bad jdk-10 install..?))

I'm going to look into that platform-specific-configuration script, to see if it's meant to set the boot jdk to the build jdk in cases wehre the boot jdk can't be found. If it isn't, I'll look into install zulu-7 on those platforms.

sxa:
It can always be overriden with `JDK7_BOOT_DIR` if needed, but I know we've tried to let the autodetection work.
[...]

PR #3240, opened 2023-11-05
[ansible] Add Ubuntu 20.04 linux-riscv64 image

luhenry:
It will be used for cross-compilation and testing.

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
@@ -184,7 +184,7 @@
   block:
     - name: Download jdk{{ jdk_version }} release (Linux-riscv64)
       get_url:
-        url: https://ci.adoptium.net/job/build-jdk19-rv64-cross/392/artifact/archive/jdk19u-riscv64-20231104.tar.gz
+        url: https://ci.adoptium.net/userContent/riscv/jdk19u-riscv64-20231107.glib227.tar.gz
         dest: /tmp/jdk19.tar.gz
         mode: 0440
       retries: 3
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
@@ -134,6 +134,7 @@
     - ansible_distribution != "MacOSX"
     - not ((ansible_distribution == "RedHat" or ansible_distribution == "CentOS") and ansible_distribution_major_version == "6")
     - ansible_os_family != "Solaris"
+    - not (ansible_architecture == "riscv64" and jdk_version == 19) # Linux-riscv64 for JDK 19 is special cased
     - adoptopenjdk_installed.rc != 0
   tags: adoptopenjdk_install
     # Api does not return release information for JDK10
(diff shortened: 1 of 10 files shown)


luhenry:
I'm not sure how to update https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck to include a ubuntu2004 check for riscv

karianna (approved):

luhenry:
Could you please add the `QEMU-playbook-check` label so that it triggers the check on CI?

sxa:
Thanks - I'll try a run of the new dockerfile on my system :-)

sxa (commented):
The docs on how QEMUPlaybook is done is under https://github.com/adoptium/infrastructure/blob/master/docs/Setup-QEMU-Images.md and it's hosted on a specific machine that you'd need access to to be able to do any of the updates as they're not currently automated.

https://github.com/adoptium/infrastructure/blob/master/docs/Setup-RISCV-VMs.md has some info on how the existing RISC-V QPC virtual machine was initially set up.

As you can probably tell the QPC stuff hasn't been updated in a while as the latest is Debian 11/Ubuntu18 so it needs a bit of attention - there's an issue relating to that at https://github.com/adoptium/infrastructure/issues/2121 but to be clear I don't consider the lack of QPC for this version to be a blocker for this PR :-)

Can you clarify the reason for having mercurial-common, python2, and upc explicitly called out? We don't use mercurial any more and I'm a little surprised we'd need to special case the others too so would be good to clarify what they're needed for if we're going to include them.

Also have you been able to test that the github actions stuff being added here work ok?

sxa:
I can't remember if the actions will trigger on the contents of the PR of if the PR has to be merged first, so was wondering if it might be simpler to test it in your own repository if that's feasisble.

Yeah mercurial was useful before openjdk moved itself to GitHub so shouldn't be necessary now as far as I'm aware (especially not on RISC-V) so I'm ok with having that stuff removed.

sxa:
> Thanks - I'll try a run of the new dockerfile on my system :-)

FYI that seemed to work well so nice to see this seems to work - I'll try and kick off a build in the new container image overnight on my machine :-)

luhenry:
I've verified locally that the image allows us to build jdk17u with:
```
podman run --rm -it --platform linux/riscv64 -v /workspace:/workspace -w $(pwd) localhost/adoptopenjdk/ubuntu2004_build_image:linux-riscv64 bash -ex -c "cd build-farm; ./make-adopt-build-farm.sh jdk17u"
```

luhenry:
I've verified the latest changes work for building `jdk20u`:
```
podman run --rm -it --platform linux/riscv64 -v /workspace:/workspace -w $(pwd) localhost/adoptopenjdk/ubuntu2004_build_image:linux-riscv64 bash -ex -c "cd build-farm; bash -x ./make-adopt-build-farm.sh jdk20u"
```

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/adoptopenjdk_install/tasks/main.yml
```diff
@@ -172,6 +173,36 @@
...
+- name: Install JDK {{ jdk_version }} on Linux-riscv64
+  when:
+    - ansible_architecture == "riscv64" and jdk_version == 19
+    - adoptopenjdk_installed.rc != 0
+  tags: adoptopenjdk_install
+    # Api does not return release information for JDK10
+  block:
+    - name: Download jdk{{ jdk_version }} release (Linux-riscv64)
+      get_url:
+        url: https://ci.adoptium.net/job/build-jdk19-rv64-cross/392/artifact/archive/jdk19u-riscv64-20231104.tar.gz
```
sxa, line 187:
```suggestion
        url: https://ci.adoptium.net/userContent/riscv/jdk19u-riscv64-20231107.glib227.tar.gz
```

sxa:
VPC run at https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/1742 as a final check.

luhenry:
It succeeded to build jdk20u locally with the latest change.

sxa:
Noting that this will currently misbehave on many real boards due to https://github.com/adoptium/adoptium-support/issues/697 unless `--security-opt seccomp=unconfined` is added to the `DOCKER_ARGS` when starting the container.

sxa:
After setting those options to stop the JVM crashing I'm getting a segfault in the compiler during the openjdk build process at the moment which isn't great :'(
```
13:58:44  g++: internal compiler error: Segmentation fault signal terminated program cc1plus
```

luhenry:
> Noting that this will currently misbehave on many real boards due to https://github.com/adoptium/adoptium-support/issues/697 unless --security-opt seccomp=unconfined is added to the DOCKER_ARGS when starting the container.

We can add the flag in https://github.com/adoptium/ci-jenkins-pipelines/pull/836.

luhenry:
> After setting those options to stop the JVM crashing I'm getting a segfault in the compiler during the openjdk build process at the moment which isn't great :'(
> 
> 13:58:44  g++: internal compiler error: Segmentation fault signal terminated program cc1plus

From talking offline, that's when _building_ on a board. I want to push further on building inside docker containers, to free up board for testing. We'll also want to use QEMU for testing for the newer extensions that we are adding support for in OpenJDK (Vector, Vector Crypto, and others). We won't be able to do this testing on boards for the next few years, even though these extensions will be common-place.

luhenry:
I can't reproduce the g++ SIGSEGV when bumping to gcc-10/g++-10 and running on VF2.

sxa (approved):
I think I'm happy for this to go in now. Noting that jdk19u will likely not be required long term and I'll look at the4 feasibility of publishing a jdk20u build against the Ubuntu 20.04 glibc which can be consumed via the API at which point we can likely clean up the JDK download stuff a bit more for the future.

PR #3282, opened 2023-12-05
ITWPlaybook: Improve Xcode Installation in ITW playbook.

### LINKED ISSUE #3281: ITW Playbook - Improve Xcode Installation Script
The xcode installation script, in the ITW playbook, should only run on old ( 10.7 & 10.8 ) versions of OSX.

Details:

Identified as part of the  Trail of Bits Audit

```
The issue in question comes from this file:
https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh
Particularly, the security concerns are: allowing installation of untrusted software (-allowUntrusted) and downloading over an insecure channel (HTTP download):
https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh#L40
https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh#L26
https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh#L30
Installation of untrusted software from an insecure channel allows for MitM attacks, and installation of potentially malicious software. The issue appears to be limited to old versions of OSX, however the version check is buggy. The script performs the following command, and compares it against 7 and 8 (to check for OSX version 10.7 and 10.8):
osx_vers=$(sw_vers -productVersion | awk -F "." '{print $2}')
The problem is that this version check only compares against the minor version number. So 11.7, 12.7, 13.7, etc, would pass this check. For example, this is the output on my machine:
$ sw_vers -productVersion
14.1
$ sw_vers -productVersion | awk -F "." '{print $2}'
1
The current [latest versions for macOS](https://support.apple.com/en-us/HT201260) 11 and 12 would pass this check (and 13 may soon). This means that more recent versions of macOS may be vulnerable to this issue, if they execute this script.
The good news is that I don't see this script referenced anywhere in the repository, so I don't think it's being executed anywhere automatically. However, it may still be executed manually. So my question is, are you aware of this script being used anywhere? If not, can it be deleted to avoid accidental execution?
It looks like it was initially introduced in [#1008](https://github.com/adoptium/infrastructure/pull/1008), but not referenced anywhere. Similarly, it looks like a Unix playbook equivalent was introduced in [#826](https://github.com/adoptium/infrastructure/pull/826), and executed as part of an Ansible task, but later deleted in [#2063](https://github.com/adoptium/infrastructure/pull/2063).
```

steelhead31:
Fixes #3281 

HTTPS download is not available for the 2 files from Apple.

SHA256 checksum tests added, and scope of script restricted to MacOS10.

This script is not run automatically as part of the playbook.

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh
+++ b/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
+os_maj_vers=$(sw_vers -productVersion | awk -F "." '{print $1}')
 osx_vers=$(sw_vers -productVersion | awk -F "." '{print $2}')
 cmd_line_tools_temp_file="/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress"
 
+echo 1 = $os_maj_vers
+echo 2 = $osx_vers
+
 # Installing the latest Xcode command line tools on 10.9.x or higher
 
-if [[ "$osx_vers" -ge 9 ]]; then
+if [[ "$os_maj_vers" -ge 10 ]] && [[ "$osx_vers" -ge 9 ]]; then
   touch "$cmd_line_tools_temp_file";
   PROD=$(softwareupdate -l |
     grep "\*.*Command Line" |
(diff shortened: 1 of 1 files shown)


ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh
```diff
@@ -1,10 +1,14 @@
...
 osx_vers=$(sw_vers -productVersion | awk -F "." '{print $2}')
 cmd_line_tools_temp_file="/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress"
 
+echo 1 = $os_maj_vers
+echo 2 = $osx_vers
+
 # Installing the latest Xcode command line tools on 10.9.x or higher
 
-if [[ "$osx_vers" -ge 9 ]]; then
+if [[ "$os_maj_vers" -ge 10 ]] && [[ "$osx_vers" -ge 9 ]]; then
```
sxa, line 11: Should this be `-eq 10` for the first clause?

steelhead31: Yup, as best I can tell ITW is restricted to OSX 10.x , fixed.

sxa (approved):

ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/Common/scripts/install-xcode.sh
```diff
@@ -20,25 +24,34 @@ fi
 # instead from public download URLs, which can be found in the dvtdownloadableindex:
 # https://devimages.apple.com.edgekey.net/downloads/xcode/simulators/index-3905972D-B609-49CE-8D06-51ADC78E07BC.dvtdownloadableindex
 
-if [[ "$osx_vers" -eq 7 ]] || [[ "$osx_vers" -eq 8 ]]; then
+if [[ "$os_maj_vers" -eq 10 ]] && [[ "$osx_vers" -eq 7 ]] || [[ "$osx_vers" -eq 8 ]]; then
```
andrew-m-leonard, line 27: How does bash handle  && || precendence?
Does this need to be
```
[[ "$os_maj_vers" -eq 10 ]] && ([[ "$osx_vers" -eq 7 ]] || [[ "$osx_vers" -eq 8 ]])
```

steelhead31: It does indeed need amending to be explicit... now done.

PR #3297, opened 2023-12-13
Github: Update Docker Login Credentials To Use Token

### LINKED ISSUE #3211: Adoptium Centos 6 docker build image in docker hub not updated since last 9 months
The Adoptium Centos 6 docker build image has not gotten the infrastructure updates for the last 9 months. Looks like something is broken.

https://hub.docker.com/r/adoptopenjdk/centos6_build_image

centos6

steelhead31:
Fixes #3211 

Update docker credentials to use PAT, but disable push on merge, due to replacement with docker multiarch jenkins job. Code left to enable easy re-enablement if required.

The docker push on merge has also been disabled, due to the relocation of the centos6 docker push to the jenkins job as per this PR: https://github.com/adoptium/infrastructure/pull/3296

##### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
       uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
       with:
         username: ${{ secrets.DOCKER_USERNAME }}
-        password: ${{ secrets.DOCKER_PASSWORD }}
+        password: ${{ secrets.DOCKER_TOKEN }}
       if: github.ref == 'refs/heads/master'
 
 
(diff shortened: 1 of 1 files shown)


Haroon-Khel (approved):
I am also in favour of the image push being in Jenkins instead of github workflow, since the Jenkins job is triggered everytime a pr is merged anyway. No need to push the same image twice

sxa (approved):
We could do a one-off run with `push: true` to verify that the functionality works as expected, but I'm ok for this to go in regardless since we're switching to using the jenkins job like we do for the other ones.

Ref: https://github.com/docker/login-action/blob/master/README.md#docker-hub (Tokens seem valid in the password field)
Ref: https://github.com/adoptium/infrastructure/pull/3296
Also noting that with these changes we should update https://github.com/adoptium/infrastructure/blob/master/FAQ.md#what-about-the-builds-that-use-the-dockerbuild-tag which talks about where the various images are built

Issue #3393, opened 2024-02-14
Document processes for out of support operating systems


sxa:
This is a follow-on to the work in https://github.com/adoptium/infrastructure/issues/2902

In addition to the main machines in the jenkins CI, we have systems in various other places that we use as part of the infrastructure. This issue will cover making sure we have documentation on what to do when an operating system goes out of support, so we can remove it from the system entirely where feasible, and also to cover policies surrounding new releases (Both STS and LTS) of the platforms we use and ideally timescale targets.

Relevant previous issues:
- [AIX 7.1](https://github.com/adoptium/infrastructure/issues/3030) 
- [Ubuntu 16.04 removal](https://github.com/adoptium/infrastructure/issues/1986)
- [Document https://github.com/adoptium/infrastructure/issues/2503

Places to consider:
- [ ] Jenkins / inventory (real machines)
- [ ] [DockerStatic containers](https://github.com/adoptium/infrastructure/tree/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/DockerStatic/Dockerfiles) used for test (See relevant [EPIC](https://github.com/adoptium/infrastructure/issues/3368))
- [ ] [VagrantPlaybookCheck](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/)/[QemuPlaybookCheck](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/)
- [ ] [DockerFileCheck](https://ci.adoptium.net/job/DockerfileCheck) (Is this still useful?)
- [ ] [Github actions](https://github.com/adoptium/infrastructure/tree/master/.github/workflows) (e.g. [this PR](https://github.com/adoptium/infrastructure/pull/3387))

Anything I've missed?

PR #3399, opened 2024-02-16
docker: Remove out of support fedora and alpine dockerfiles

Haroon-Khel:

fedora 33 and 34 are out of support https://docs.fedoraproject.org/en-US/releases/eol/
Alpine 312, 313 and 314 are out of support https://endoflife.date/alpine

The running containers will be replaced without the latest versions of both OSs, alpine 319 and fedora39


Haroon-Khel:
The dockerstatic checker workflow will fail for the removed OSs

sxa:
> The dockerstatic checker workflow will fail for the removed OSs

Surely they should be removed from https://github.com/adoptium/infrastructure/blob/master/.github/workflows/check_dockerstatic.yml in this PR too?

Haroon-Khel:
I was saving it for another pr but yeah it makes sense to do it here

karianna (approved):

karianna:
Should old Ubuntu versions also go?

Haroon-Khel:
I believe the out of date ubuntu distros are being discussed here https://github.com/adoptium/infrastructure/issues/3393. The'r removal can be saved for another pr

sxa:
> > Should old Ubuntu versions also go?

> I believe the out of date ubuntu distros are being discussed here #3393. Their removal can be saved for another pr

That's more about defining a policy and setting appropriate process in place going forward. We should certainly look at removing any old Ubuntu containers that we have at Temurin.

While I could be convinced either way I think we should leave the LTS Dockerfiles in the repository as long as Temurin can still run on them, especially since even Ubuntu 16.04 still has extended support if customers wish to pay Canonical for it until 2026 so is still viable if a customer with a suitable contract wished to use these dockerfiles to run an AQA suite. That would not be the case for the Fedora/Alpine versions being targeted by this PR.

steelhead31 (approved):
LGTM

PR #3410, opened 2024-02-21
macos ansible setup improvements

mahdipub:
unixPB: This PR is amed to add Xcode_offline which installs xcode without need to use any other 3rd party resource. Also some minor code changes and improvement added (e.g. install brew)

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/macos-setup.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/macos-setup.yml
@@ -1,0 +1,54 @@
+---
+
+- name: Ansible MacOS playbook
+  hosts: "all"
+  gather_facts: yes
+  become_method: sudo
+  vars:
+    ansible_python_interpreter: /Library/Frameworks/Python.framework/Versions/3.11/bin/python3
+    Ant_Download_URL: https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.12-bin.zip
+    known_keys:
+      - {name: "github.com", value: "github.com ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOMqqnkVzrm0SdG6UOoqKLsabgH5C9okWi0dh2l9GKJl"}
+    xcode_version: '15.0.1'
+    ansible_user: adminuser
+    jdk_version: 17
+
+
+  #########
+  # Roles #
+  #########
+  roles:
+    - role: brew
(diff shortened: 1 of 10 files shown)


karianna (approved):

karianna (changes requested):
Some unhappy YAML lint

sxa (commented):
I've added a couple of other people as reviewers who are more knowledgeable about the macos stuff than I am but from a general perspective:
- At present we use the `main.yml` file for macos, so switching to use a separate one superficially feels like a slightly backwards step. Is it possible to integrate some of these changes into the normal execution of `main.yml` or is there a reason why you have chosen to do this as a separate top level file?
- This seems to have quite significant changes to some of the roles such as the ant-contrib one - are they specific to the goal of this PR to improve usability of the macos playbooks?
- Similarly this seems to change how things like the credentials are retrieved for the Jenkins_User role, which would be a change in behaviour.

Perhaps some of those things can be decoupled from this PR and proposed separately unless there is a reason they've all been put together.

mahdipub:
> I've added a couple of other people as reviewers who are more knowledgeable about the macos stuff than I am but from a general perspective:
> 
> * At present we use the `main.yml` file for macos, so switching to use a separate one superficially feels like a slightly backwards step. Is it possible to integrate some of these changes into the normal execution of `main.yml` or is there a reason why you have chosen to do this as a separate top level file?
> * This seems to have quite significant changes to some of the roles such as the ant-contrib one - are they specific to the goal of this PR to improve usability of the macos playbooks?
> * Similarly this seems to change how things like the credentials are retrieved for the Jenkins_User role, which would be a change in behaviour.
> 
> Perhaps some of those things can be decoupled from this PR and proposed separately unless there is a reason they've all been put together.

- Yes. This file (including comments on the bottom) is only to show what are changes, and how does this whole works for macos. For sure we can combine them. 
- Yes. But at the same time they are general improvements in code and logic too. e.g in ant-contrib most of it is to introduce a new variable {{ ant_binary_file_name }} which lets avoid repetition of extracting the path every time we need. 
- Agreed. 
- Agreed. I can focus on xcode offline and then ask for the rest in separate PR.

karianna:
@mahdipub Can you resolve the linter fixes?

mahdipub:
Based on comments on slack, I will close this PR and split changes in 2~3 PRs. The one related to Xcode-offline and the other for brew is https://github.com/adoptium/infrastructure/pull/3417.

PR #3455, opened 2024-03-06
UnixPB: Include check for ntpd service.

### LINKED ISSUE #3454: Centos 7 docker image update job failing
As seen in this log: https://ci.adoptium.net/job/centos7_docker_image_updater/461/execution/node/83/log/

The unix playbook role for NTP time fails inside a docker container, due to it attempting to start a service which isn't present.

The NTP_TIME role needs a fix to check if the service is present prior to attempting to start it.

steelhead31:
Fixes #3454 

##### Checklist

VPC Completed OK: https://ci.adoptium.net/job/VagrantPlaybookCheck/1842/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NTP_TIME/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/NTP_TIME/tasks/main.yml
@@ -37,12 +37,38 @@
     - (ansible_distribution == "Ubuntu") or (ansible_distribution == "SLES" and ansible_distribution_major_version == "11")
   tags: ntp_time
 
+- name: Gather Facts About The Services Present
+  service_facts:
+  tags: ntp_time
+
+- name: Check If NTPD Exists In The Service Facts
+  set_fact:
+    ntpd_entry_exists: "{{ 'ntpd.service' in services }}"
+  when: ansible_facts.services is defined
+  tags: ntp_time
+
+- name: Set Fact Where NTPD Is Not Available As A Service
+  set_fact:
+    ntpd_entry_exists: "false"
+  when: ansible_facts.services is not defined
+  tags: ntp_time
+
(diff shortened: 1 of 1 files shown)


sxa (changes requested):
Can you redo the commit message to prefix it with `unixPB:` (Can be done on squash but if someone else does it it's likely to get missed.

Do we install the ntp packages in containers? Assuming yes I guess it just doesn't get registered which is why this fails?

steelhead31:
> Can you redo the commit message to prefix it with `unixPB:` (Can be done on squash but if someone else does it it's likely to get missed.
> 
> Do we install the ntp packages in containers? Assuming yes I guess it just doesn't get registered which is why this fails?

Done, we do install NTPD, but the docker containers lack "services" , systemctl returns a bus error, which I guess is a docker thing, so I think this is probably the closes thing to a solution. It seems to work OK on Centos7 & Centos8, but the VPC on Centos6 failed, so I'll need to do a bit more investigation into that.

sxa (approved):
> systemctl returns a bus error

Yikes, but ok ... :-)

steelhead31:
> > systemctl returns a bus error
> 
> Yikes, but ok ... :-)

For documentation purposes...  
  
systemd requires CAP\_SYS\_ADMIN capability but Docker drops that capability in the non privileged containers, in order to add more security... I guess we could look to implement this if required in our containers.

steelhead31:
Rerun Of The VPC with improved handling for CentOS6..  
  
https://ci.adoptium.net/job/VagrantPlaybookCheck/1840/

sxa:
> systemd requires CAP\_SYS\_ADMIN capability ...I guess we could look to implement this if required in our containers.

I wouldn't have thought we need it. Is hope that time sync  would be handled by the host for the containerised environment

steelhead31:
Failing checks are known issues and not related.

gdams (approved):

PR #3509, opened 2024-04-08
inventory: Update public IP for win2019 test machine

steelhead31:
Following a restart , the public IP address of this machine has changed, and the inventory requires an update.

##### Checklist

- [X] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

### DIFF
--- a/ansible/inventory.yml
+++ b/ansible/inventory.yml
@@ -111,7 +111,7 @@ hosts:
 
       - azure:
           win2016-x64-1: {ip: 52.149.211.210, user: adoptopenjdk}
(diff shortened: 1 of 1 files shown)


Haroon-Khel (approved):

sxa (commented):
Do we know what caused the IP to change? That doesn't normally happen on the Azure machines just because of a restart

steelhead31:
> Do we know what caused the IP to change? That doesn't normally happen on the Azure machines just because of a restart

It looks like the network setup on this machine is a bit different to others, it has a dynamically allocated public IP attached to its static.

sxa (approved):
LGTM

PR #3544, opened 2024-05-02
UnixPB: Only Install CMake on mac where not already present.

### LINKED ISSUE #3549: GitHub Actions For Checking Ansible Playbooks
THe github action for checking the ansible playbooks on MacOS both have issues.

MacOS14 is failing due to the playbook attempting to install cmake on the github runner when it is already installed.

MacOS14 is also failing due to the keyserver.ubuntu.com hostname not resolving with a no route to host error ( raised upstream : actions/runner-images#9777 )

MacOS11 is performing so badly, that eventually the check appears to timeout, given that macos 11 was end of life back in November 2023, this check should be moved up to macos13.

steelhead31:
This should fix the failing GHA, as the MacOS 14 runner, already has cmake installed.

Fixes #3549 

##### Checklist

### DIFF
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -27,6 +27,9 @@ jobs:
     - name: Install dependencies
       run: brew install ansible
 
+    # This is to fix an issue with the github macos14 runner, that cant resolve keyserver.ubuntu.com
+    # but this does work when replaced with the IP address. Spotted during the GPG verification of ANT step
+    # Upstream Bug : https://github.com/actions/runner-images/issues/9777
     - name: Replace keyserver address with IP address
       run: |
         sed -i '' 's/keyserver\.ubuntu\.com/185.125.188.27/g' ansible/playbooks/Supporting_Scripts/package_signature_verification.sh
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -28,7 +28,8 @@ jobs:
       run: brew install ansible
 
     - name: Replace keyserver address with IP address
-      run: sed -i 's/keyserver\.ubuntu\.com/185.125.188.27/g' playbooks/Supporting_Scripts/package_signature_verification.sh
+      run: |
+        sed -i 's/keyserver\.ubuntu\.com/185.125.188.27/g' playbooks/Supporting_Scripts/package_signature_verification.sh
 
     - name: Run Ansible Playbook
       run: |
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -22,7 +22,7 @@ jobs:
           - os: [macos-14]
     steps:
 
-    - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+    - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
 
     - name: Install dependencies
       run: brew install ansible
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -31,6 +31,10 @@ jobs:
       run: |
         sed -i '' 's/keyserver\.ubuntu\.com/185.125.188.27/g' ansible/playbooks/Supporting_Scripts/package_signature_verification.sh
 
+    - name: Remove Cmake
+      run: |
+        brew uninstall cmake
+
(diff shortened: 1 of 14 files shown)


gdams (approved):

sxa (approved):
I think the location is different for x64 and aarch64, although given that we're only seeing the problem on one of them (AFAIK) I'm ok with this for now.

I'm curious though - do we know why this causes the problem? I'd have hoped that `state: present` would make it do the right thing and only install if it's not there - almost feels like cmake has been installed on the runners in a manner which isn't compatible with brew.
We should also consider that if it's an "unofficial" install we've no idea where it came from or what version it is.

steelhead31:
> I think the location is different for x64 and aarch64, although given that we're only seeing the problem on one of them (AFAIK) I'm ok with this for now.
> 
> I'm curious though - do we know why this causes the problem? I'd have hoped that `state: present` would make it do the right thing and only install if it's not there - almost feels like cmake has been installed on the runners in a manner which isn't compatible with brew. We should also consider that if it's an "unofficial" install we've no idea where it came from or what version it is.

@sxa Im not sure how the github runners are constructed, Im assuming without ansible, but I can ask the question upstream, on my bug... an alternate solution for the github action is to remove it from the runner, so the playbook can install the correct version, as this only occurs on the github runner, it works as expected everywhere else.

steelhead31:
Going to quickly test the removal from the runner, and allowing the playbook to run normally :)

steelhead31:
Removing cmake from the runner is not straightforward, so the ansible check is currently a tidier solution, I'll ask some questions upstream.

.github/workflows/build_mac.yml
```diff
@@ -27,6 +27,10 @@ jobs:
     - name: Install dependencies
       run: brew install ansible
 
+    - name: Replace keyserver address with IP address
+      run: |
+        sed -i '' 's/keyserver\.ubuntu\.com/185.125.188.27/g' ansible/playbooks/Supporting_Scripts/package_signature_verification.sh
```
karianna, line 32: Why?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml
```diff
@@ -125,6 +125,21 @@
   with_items: "{{ Build_Tool_Packages }}"
   tags: build_tools
 
+# Install CMAKE Where It Doesnt exist
```
karianna, line 128:
```suggestion
# Install CMAKE Where It Doesn't exist
```

karianna (changes requested):

karianna:
Ran the Mac OS X 11 action twice but I can't see the clear reason for failure.

.github/workflows/build_mac.yml
steelhead31: The github runner, cant resolve keyserver.ubuntu.com, but the IP address is reachable:  upstream bug raised: actions/runner-images#9777

karianna: Can we add a comment in the source code to that effect?

steelhead31: Done!

steelhead31: Done!

steelhead31:
Also a quick note that the macos-11 runner will be deprecated from 28th June see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners

```
The macos-11 label has been deprecated and will no longer be available after 6/28/2024.
```

gdams:
> Also a quick note that the macos-11 runner will be deprecated from 28th June see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners
> 
> ```
> The macos-11 label has been deprecated and will no longer be available after 6/28/2024.
> ```

I'd just kill it, maybe switch it out with macos-13?

sxa:
> > Also a quick note that the macos-11 runner will be deprecated from 28th June see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners
> > ```
> > The macos-11 label has been deprecated and will no longer be available after 6/28/2024.
> > ```
> 
> I'd just kill it, maybe switch it out with macos-13?

Yeah if it's being deprecated that seems like a reasonable course of action

PR #3628, opened 2024-06-26
WindowsPB: Update Wix5 Installation To Be Global

steelhead31:
Update Wix 5 installation to be global, rather than user specific. This change allows Wix 5 to be installed and be accessible by all users, and not just the jenkins user. This allows the playbook to follow the standard process of having the installation carried out by the administrative user, and executed by other users.

This also allows the playbook to be tested successfully via the VPC check process.

**To Clarify**
When using the --global switch , the Wix package gets installed into a user specific directory, this is usually the administrators home directory, and not accessible for other users e.g. Jenkins.

By using --tool-path, the tool can be installed into a directory that is available to the system, so that jenkins, vagrant and other users can execute wix.

##### Checklist

VPC in progress : https://ci.adoptium.net/job/VagrantPlaybookCheck/1888/


gdams (approved):

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/WiX/tasks/main.yml
```diff
@@ -35,25 +35,19 @@
...
-  become: true
   when: wix_installed.stat.exists
   tags: Wix
 
 - name: Install Wix Toolset when not installed or version is not correct
-  win_shell: dotnet tool install --global wix --version {{ wix_version }}
-  become_user: "{{ Jenkins_Username }}"
-  become_method: ansible.builtin.runas
-  become: true
+  win_shell: dotnet tool install wix --version {{ wix_version }} --tool-path 'C:\\Program Files\\dotnet'
```
sxa, line 50: Is the removal of `--global` here correct? To a non-expert that looks inconsistent with the goal of this PR

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/GPG_signature_verification/tasks/main.yml
```diff
@@ -14,7 +14,7 @@
   win_shell: C:/cygwin64/bin/bash.exe -c "chmod og-rwx ~/.gnupg"
 
 - name: Import GPG Key
-  win_command: gpg --keyserver keyserver.ubuntu.com --recv-keys "{{ GPG_key }}"
+  win_shell: C:/cygwin64/bin/bash.exe -c "gpg --keyserver keyserver.ubuntu.com --recv-keys "{{ GPG_key }}""
```
sxa, line 17: General question, but what's the advantage/need of win_shell over win_command here?

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/WiX/tasks/main.yml
steelhead31: Yes, strangely to install into a specific path, the --global-path has to be omitted, the --tool-path and --global parameters are mutually exclusive, by forcing a path that is available system wide, rather than under ../USERNAME/.dotnet/bin, the global flag is redundant.

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/GPG_signature_verification/tasks/main.yml
steelhead31: I was just aligning it with the commands above it, the commands running in the cygwin shell, appear to prefer the use of win\_shell, I suspect it actually doesn't matter.. It is important that the gpg signing does run in the cygwin shell though, as otherwise GPG isn't available in Windows, and the playbooks and Github Actions both hang during GPG signature verification of Ant.

steelhead31: I need to make a similar fix to the binary verification, so i'll try them both using win\_command.

steelhead31: Win_Command hangs when running the cygwin bash shell, reverting to win_shell

steelhead31:
Reverting to draft, whilst I figure out the fix for GPG verification of Ant

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/WiX/tasks/main.yml
sxa: That's weird, but ok ;-)

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/GPG_signature_verification/tasks/main.yml
steelhead31: PR for GPG validation is now fixed

sxa (approved):
Much nicer :-)

PR #3639, opened 2024-06-28
WinPB: Update GPG Verification Process For Ant Binary

### LINKED ISSUE #3638: Ansible bug: GPG Verification Of Ant Download Hangs Playbook Execution.
- Bug in ansible playbook

Details:

The current Windows playbook fails to complete, on Vagrantm GitHubActions or Azure VMs.

The GPG verification of the Ant binary hangs when importing the gpg key from the keyserver.

steelhead31:
Fixes #3638 

##### Checklist

VPC OK: https://ci.adoptium.net/job/VagrantPlaybookCheck/OS=Win2022,label=vagrant/1895/console 

Fails on wix installation, which is fixed in this PR, which will be rebased with these changes:

https://github.com/adoptium/infrastructure/pull/3628

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
@@ -30,7 +30,7 @@
   include_tasks:
     file: ../../GPG_signature_verification/tasks/main.yml
   vars:
-    file_path: c:\temp\ant.zip
+    file_path: c:/temp/ant.zip
     signature_link: "https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.5-bin.zip.asc"
     GPG_key: "{{ key.apache_ant }}"
   when: (not ant_installed.stat.exists)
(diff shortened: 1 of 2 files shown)


gdams (approved):

tellison (commented):
I'd like to just question the range of code within the disable/re-enable of the gpgagent and dirmgr 'block'. 

1) Excuse my ignorance - is there a way to wrap the code in a structure that will be assured to be run if the commands within the 'block' fail, so that a failure case doesn't leave the state disabled?

2) Can you assure that *all* commands in the 'block' require the state to be disabled, just so we know we have minimised all commands running in the 'block'.

steelhead31:
> I'd like to just question the range of code within the disable/re-enable of the gpgagent and dirmgr 'block'.
> 
> 1. Excuse my ignorance - is there a way to wrap the code in a structure that will be assured to be run if the commands within the 'block' fail, so that a failure case doesn't leave the state disabled?

I've updated the code, so that it will always run the chmod to restore the permissions

> 2. Can you assure that _all_ commands in the 'block' require the state to be disabled, just so we know we have minimised all commands running in the 'block'.

I've reduced the scope of the disablement to just be disabled for the import of the gpg key, as this is the step that causes issues by spawning unwanted subprocesses, that hang execution.

andrew-m-leonard (approved):

Haroon-Khel (approved):

PR #3647, opened 2024-07-02
Docker: Fix Use Of mirrorlist.centos.org in dockerfiles for arm64 & ppc64le

### LINKED ISSUE #3646: Centos7 Docker Image, Mirrors failing for arm64 & ppc64le
Centos 7 based elements in the dockerfile are failing due to mirrorlist.centos.org no longer being valid. This has been rectified for x64, however issues are being encountered on arm64 and ppc64le

```
Loaded plugins: fastestmirror, ovl
14Determining fastest mirrors
15Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
1614: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
17
18
19 One of the configured repositories failed (Unknown),
20 and yum doesn't have enough cached data to continue. At this point the only
21 safe thing yum can do is fail. There are a few ways to work "fix" this:
22
23 1. Contact the upstream for the repository and get them to fix the problem.
24
25 2. Reconfigure the baseurl/etc. for the repository, to point to a working
26 upstream. This is most often useful if you are using a newer
27 distribution release than is supported by the repository (and the
28 packages for the previous distribution release still work).
29
30 3. Run the command with the repository temporarily disabled
31 yum --disablerepo=<repoid> ...
```

steelhead31:
Fixes #3646 

Update various calls to mirrorlist.centos.org with vault.centos.org, now that mirrorlist has been deprecated following the EOL of Centos 7.

##### Checklist

### DIFF
--- a/ansible/docker/Dockerfile.CentOS7
+++ b/ansible/docker/Dockerfile.CentOS7
@@ -9,7 +9,7 @@ RUN if [ "$(uname -m)" == "x86_64" ]; then \
       echo "Running on x64 architecture"; \
       sed -i -e 's!#baseurl=http://mirror.centos.org/centos/\$releasever!baseurl=https://vault.centos.org/7.9.2009/!g' /etc/yum.repos.d/CentOS-Base.repo; \
     else \
-      echo "Running on another architecture"; \
+      echo "Running on non-x64 architecture"; \
       sed -i 's|#baseurl=http://mirror.centos.org/altarch/\$releasever/|baseurl=http://vault.centos.org/altarch/7.9.2009/|' /etc/yum.repos.d/CentOS-Base.repo; \
     fi
 
--- a/ansible/docker/Dockerfile.CentOS7
+++ b/ansible/docker/Dockerfile.CentOS7
@@ -3,9 +3,21 @@ FROM centos:7
 ARG git_sha
 ARG user=jenkins
 
-RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo; \
-    sed -i -e 's!#baseurl=http://mirror.centos.org/centos/\$releasever!baseurl=https://vault.centos.org/7.9.2009/!g' /etc/yum.repos.d/CentOS-Base.repo; \
-    yum -y update; yum clean all; \
+RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo
+
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
+      echo "Running on arm64 architecture"; \
+      sed -i 's|#baseurl=http://mirror.centos.org/altarch/\$releasever/|baseurl=http://vault.centos.org/altarch/7.9.2009/|' /etc/yum.repos.d/CentOS-Base.repo; \
+    elif [ "$(uname -m)" = "ppc64le" ]; then \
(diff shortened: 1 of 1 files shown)


steelhead31:
Tested on x64 & arm64 OK, testing on ppc4le is still in progress

andrew-m-leonard (approved):
lgtm

ansible/docker/Dockerfile.CentOS7
```diff
@@ -3,9 +3,21 @@ FROM centos:7
 ARG git_sha
 ARG user=jenkins
 
-RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo; \
-    sed -i -e 's!#baseurl=http://mirror.centos.org/centos/\$releasever!baseurl=https://vault.centos.org/7.9.2009/!g' /etc/yum.repos.d/CentOS-Base.repo; \
-    yum -y update; yum clean all; \
+RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo
+
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
```
sxa, line 8: Maybe I'm missing something but given that the non-x64 ones are the same apart from the echo statement, is there any reason not to collapse them and use `if [ "$(uname -m)" != "aarch64" ];` to keep this simpler?

sxa (commented):
One question ...
I guess the echo could be retained too but saying "Running from CentOS altarch repos" but I'd be tempted to just miss it out.

ansible/docker/Dockerfile.CentOS7
steelhead31: Good point, let me tidy that... I was expecting them to be different when I wrote the if statement, whilst I figured out the sed goodness...

steelhead31: Fixed

ansible/docker/Dockerfile.CentOS7
```diff
@@ -3,9 +3,17 @@ FROM centos:7
...
-RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo; \
-    sed -i -e 's!#baseurl=http://mirror.centos.org/centos/\$releasever!baseurl=https://vault.centos.org/7.9.2009/!g' /etc/yum.repos.d/CentOS-Base.repo; \
-    yum -y update; yum clean all; \
+RUN sed -i -e 's!mirrorlist!#mirrorlist!g' /etc/yum.repos.d/CentOS-Base.repo
+
+RUN if [ "$(uname -m)" == "x86_64" ]; then \
+      echo "Running on x64 architecture"; \
+      sed -i -e 's!#baseurl=http://mirror.centos.org/centos/\$releasever!baseurl=https://vault.centos.org/7.9.2009/!g' /etc/yum.repos.d/CentOS-Base.repo; \
+    else \
+      echo "Running on another architecture"; \
```
sxa, line 12:
```suggestion
      echo "Running on non-x64 architecture"; \
```

sxa (approved):
LGTM. You can take or leave my suggestion on the message (It won't be obvious what "another" if you're just watching the output  from building the dockerfile)

karianna (approved):

Issue #3286, opened 2023-12-06
EPIC: New Machine requirement: Windows dockerBuild containers


sxa:
I need to request a new machine:

- New machine operating system (e.g. linux/windows/macos/solaris/aix): Windows
- New machine architecture (e.g. x64/aarch32/arm32/ppc64/ppc64le/sparc): x64
- Provider (leave blank if it does not matter): Docker :-)
- Desired usage: Build containers, similar to what we have for Linux
- Any unusual specification/setup required:
- How many of them are required: n/a - they should be created dynamically

Please explain what this machine is needed for:
Running builds in an isolated way where we can achieve SLSA build level 3 compliance on Windows along with the other primary platforms. Ideally we'll be able to create windows-on-windows container images which we share and then download and run the builds in. 

As background info:
- https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=dockerce#windows-server-1 looks to be a good resource on the software options for containers on Windows
- Eclipse Temurin does provide container images for Windows at https://hub.docker.com/_/eclipse-temurin/ - these are created from the dockerfiles in https://github.com/adoptium/containers/tree/main/21/jdk/windows which may give a good starting point for this work on investigating windows containers.

So the tasks required would be:
- [x] Identify the appropriate software for running containers and ensure no licensing concerns (Likely something from the microsoft site linked above)
- [x] See if we can verify that a "basic" dockerfile works in that environment and whether we can map directories into it (same as `-v` on linux) which are read+write in the container
- [x] Determine whether we can create a container from the playbooks using a dockerfile equivalent to [the Linux ones](https://github.com/adoptium/infrastructure/tree/master/ansible/docker)
- [x] Once we create the container, map a directoryi from the host into it with `-v` and use that to build Temurin in the container on the mapped volume so that the output is visible on the host system.
- [x] Understand whether we can reasonably push the resulting container images with the compiler up to dockerhub _[Answer: We will push them to gcr.io and mirror to azurecr for better performance on azure dynamic machines.]_
- [x] Integrate this into [the build pipelines](https://github.com/adoptium/ci-jenkins-pipelines/blob/7b9559ce88321ff8111180fdc58421a0f9eadcef/pipelines/build/common/openjdk_build_pipeline.groovy#L1876)
[...]

PR #3702, opened 2024-08-08
winPB: Switch to VS2022 build tools and add Windows build dockerfile

sxa:
**This PR is in draft to give as many people as possible time to review and voice options and not having this one quickly merged. Approvals are welcome in the meantime**

Notes:
- Adds Dockerfile.win2022 which can be used to create a build image comparable to the Linux ones. This will only install VS2022, not the earlier other compilers (this will be fine for all versions once https://github.com/adoptium/temurin-build/pull/3907 is merged)
- Various modifications to support the above including:
  - Adding more 'win_reboot' tags to tasks to avoid attempting to reboot while running ansible inside the container
  - Use of gpg2 instead of gpg since that is what the new cygwin seems to have
  - Since the dockerfile does a base cygwin install, the ansible cygwin role now checks for jq.exe (the most recent addition) instead before deciding whether to skip the cygwin packages
  - Uses a windows command instead of cygwin (which may not be in the path at that point) as the "Dummy" command in the logs role
  - Changes the command used for default shortname creation to one that works in containers
- Uses Visual Studio Build Tools instead of the community edition (Currently will not take effect if "adoptopenjdk" is skipped but that is skipped in the new dockerfile so it takes effect there)

Not done:
- I haven't got the `git_sha` detection working in the logs role, so it is currently hard coded in the Docker file to 00000000. Suggestions on how to make this work are welcome
- The cygwin.exe checksum is not verified in the Dockerfile
- If the playbook fails the Docker process does not currently fail as `&` doesn't propagate the failure

The Dockerfile file can be build with an invocation such as the following (there is a password in the dockerfile, but perhaps that should be removed? The account is deleted afterwards so the password cannot be used once the image is complete):
- `docker build -t win2022_build_image --no-cache -e PW=S0meVal=dPW -f Dockerfile.win2022 .`
This requires docker to be installed on the host system and [8dot3names to be enabled on the host system](https://github.com/microsoft/Windows-Containers/issues/507#issuecomment-2270986045).

For discussion from reviewers:
- Is using gpg2 instead of gpg going to cause any problems (e.g. on existing machines)?
- I am now checking for the presence of `C:\Program Files\Microsoft Visual Studio\2022` instead of the `\Community` subdirectory so that community or build tools will be detected (Although I've just realised that since build tools goes into `C:\Program Files (x64) - as can be seen in the `Register Visual Studio Community 2022 DIA SDK shared libraries` operation - so that probably isn't going to work anyway for the build tools case
- I'd quite like to get this in with the "known issues" that I can raise follow-up issues on so that progress with the automation around this can proceed
- Should I remove the temporary password from the Dockerfile and force it to be explicitly specified?
- I'd rather not be using WinRM with a user and just run this in a local manner, but that would require more effort on Windows. Any interested parties can look at those changes for use later

Fixes https://github.com/adoptium/infrastructure/issues/3286 https://github.com/adoptium/temurin-build/issues/3787 (Note that for validation purposes I have also run through the process to verify that this dockerfile and compiler can produce a binary identical build of 21.0.4+7 on Windows)

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) ~https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/1948/~ (Failed due to a Dragonwell SHA failure - seems ok now so re-running at https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/1949/

### DIFF
--- a/ansible/docker/Dockerfile.win2022
+++ b/ansible/docker/Dockerfile.win2022
@@ -1,0 +1,32 @@
+FROM mcr.microsoft.com/windows/servercore:ltsc2022
+
+# Specify this with --build-arg PW=SomePassword
+ARG PW=T3mp=Passwd
+
+# Set up cygwin with git and ansible as a bootstrap, and add to system default path
+RUN powershell wget -UseBasicParsing https://cygwin.com/setup-x86_64.exe -OutFile setup-x86_64.exe & \
+    setup-x86_64.exe --packages git,ansible --download --local-install --delete-orphans --site https://mirrors.kernel.org/sourceware/cygwin --local-package-dir c:\cygwin_packages --root C:\cygwin64 --wait --quiet-mode & \
+    C:\cygwin64\bin\git config --system core.autocrlf false & \
+    del setup-x86_64.exe & \
+    setx PATH "c:\cygwin64\bin;%PATH%" & \
+    mkdir c:\temp
+
+# Set up WinRM for the ansible connection
+RUN powershell wget -UseBasicParsing https://raw.githubusercontent.com/ansible/ansible/dd4c56e4d68664e4a50292aa19ea61b15c92287c/examples/scripts/ConfigureRemotingForAnsible.ps1 -OutFile ConfigureRemotingForAnsible.ps1 & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -CertValidityDays 9999 & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -EnableCredSSP & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -SkipNetworkProfileCheck
+
+# Set up WinRM user, clone and run the playbook, then delete the user so it's not in any layer
+ENV TERM dumb
+RUN net user ansible %PW% /ADD & net localgroup "Administrators" ansible /ADD & net localgroup "Remote Management Users" ansible /ADD & \
+    C:\cygwin64\bin\git clone https://github.com/sxa/infrastructure -b windows_docker_support c:/infrastructure & \
+    sed -i -e 's/hosts: .*/hosts: localhost/' infrastructure/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/main.yml & \
+    echo localhost ansible_connection=winrm > infrastructure/ansible/hosts & \
+    cd infrastructure\ansible & \
+    C:\cygwin64\bin\python3.7m.exe /usr/bin/ansible-playbook -e git_sha=00000000 -e ansible_user=ansible -e ansible_password=%PW% -i hosts \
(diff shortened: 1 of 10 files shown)


ansible/docker/Dockerfile.win2022
```diff
@@ -0,0 +1,32 @@
...
+
+# Set up WinRM for the ansible connection
+RUN powershell wget -UseBasicParsing https://raw.githubusercontent.com/ansible/ansible/dd4c56e4d68664e4a50292aa19ea61b15c92287c/examples/scripts/ConfigureRemotingForAnsible.ps1 -OutFile ConfigureRemotingForAnsible.ps1 & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -CertValidityDays 9999 & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -EnableCredSSP & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert & \
+    PowerShell .\ConfigureRemotingForAnsible.ps1 -SkipNetworkProfileCheck
+
+# Set up WinRM user, clone and run the playbook, then delete the user so it's not in any layer
+ENV TERM dumb
```
gdams, line 22: Dockerfile changed this format recently so we should use the correct way

```suggestion
ENV TERM=dumb

```

sxa:
> how do you plan to build/push this? A GitHub action might be quite smart?

That's another step beyond this PR so we should discuss elsewhere (perhaps in the [related issue](https://github.com/adoptium/infrastructure/issues/3286)). All ideas welcome of course, but I was planning to prototype it on one of our machines under control of jenkins like we do for the others since my preference would not be to rely on GitHub actions and have this done on our own infrastructure.

andrew-m-leonard (approved):
looks good

gdams (approved):

Haroon-Khel (approved):

sxa:
Added `USER ContainerUser` to stop the build running as an Administrator inside the container - I trust this will not invalidate anyone's reviews.

Noting also that the container will need access to a git config which allows the workspace directory to be treated as "safe" due to the initial jenkins git checkout of the repository being done on the host system which has a different SID from the user in the container. I have been doing this by setting `HOME` in the jenkins agent definition to be `C:\jw` which is what I'm currently using for the jenkins workspace, and having a `.gitconfig` in there with the following to allow subsequent git operations in the directory to succeed. (I was trying to work with more restrictive versions of this but haven't got it to work yet)
```
[safe]
        directory = *
```
This is not an ideal solution as it would potential allow git configurations to be modified and persisted across the build runs, but that can be investigated later, possibly using a [`GIT_CONFIG` environment variable](https://git-scm.com/docs/git-config).

sxa:
VPC run looks ok - I'm going to run one more with a build test at https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/1950/

sxa:
> VPC run looks ok - I'm going to run one more with a build test at https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/1950/

That ran ok but there seems to be some failures in the test section that are unrelated to this PR which I've raised at https://github.com/adoptium/infrastructure/issues/3707

PR #3716, opened 2024-08-15
aixPb: Put /usr/bin in front of PATH

Haroon-Khel:

To escape the rebase hell that is https://github.com/adoptium/infrastructure/pull/2288/, ive simply created a new pr :)

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/bash_login/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_AIX_Playbook/roles/bash_login/tasks/main.yml
@@ -7,14 +7,18 @@
     path: /etc/security/login.cfg
     regexp: 'shells = '
     replace: 'shells = /bin/bash,'
-  tags: login_shell
+  tags:
+    - login_shell
+    - adoptopenjdk
 
 - name: Add bash to available login shells
   blockinfile:
     dest: /etc/shells
     block: |
       /bin/bash
-  tags: login_shell
+  tags:
+    - login_shell
+    - adoptopenjdk
 
 # move to role later
 - name: Set variables for global environment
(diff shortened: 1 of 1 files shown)


sxa (commented):
Why is `/etc` in there? I can't think of anything we need in the PATH that would be in there

Haroon-Khel:
Not sure. I'll remove it

steelhead31 (approved):
Looks ok.

sxa (approved):

sxa (approved):
We should also bear in mind that while this will be ok for new machines it likely won't change the existing machines which have `PATH=/opt/freeware/bin` at the start as [most machines currently won't match this regex](https://github.com/adoptium/infrastructure/pull/2288#pullrequestreview-847554635), so if we don't put another case in here for that we'll need to look at patching the existing machines up (as described in https://github.com/adoptium/infrastructure/pull/2288#issuecomment-1008679917). Also noting that this regex to update the config will trigger on every playbook execution since it sets the path to something that has `PATH=/usr/bin` at the start. Something like @aixtools' proposal in https://github.com/adoptium/infrastructure/pull/2288#pullrequestreview-771855793 may be preferable, but I'm also ok with a one-off manual fixup.

karianna:
> We should also bear in mind that while this will be ok for new machines it likely won't change the existing machines which have `PATH=/opt/freeware/bin` at the start as [most machines currently won't match this regex](https://github.com/adoptium/infrastructure/pull/2288#pullrequestreview-847554635), so if we don't put another case in here for that we'll need to look at patching the existing machines up (as described in [#2288 (comment)](https://github.com/adoptium/infrastructure/pull/2288#issuecomment-1008679917)). Also noting that this regex to update the config will trigger on every playbook execution since it sets the path to something that has `PATH=/usr/bin` at the start. Something like @aixtools' proposal in [#2288 (review)](https://github.com/adoptium/infrastructure/pull/2288#pullrequestreview-771855793) may be preferable, but I'm also ok with a one-off manual fixup.

holding off merging until we raise any appropriate work items here for consistency.

gdams (approved):

Haroon-Khel:
I've manually updated the PATH on the current machines to start with `/usr/bin:/opt/IBM/xlC/13.1.3/bin:/opt/freeware/bin:`

aixtools:
And, are there issues after this manual change? Thought I'd give it a bit of rest.

PR #3757, opened 2024-09-30
unixPB: removes become for brew installs in `macos - > common` task

mahdipub:
In macos when installing using brew and using a root/powered user, an error will show by brew: `Running Homebrew as root is extremely dangerous and no longer supported.`. In general we do not need `become` when installing with brew. This will remove become in common > macos brew installations.

Signed-off-by: mahdi@ibm.com

##### Checklist


karianna:
The Mac OS X GH Action failed with:

TASK [Common : Add AdoptOpenJDK Java Repo] *************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "added: 0, unchanged: 0, error: failed to tap: AdoptOpenJDK/openjdk due to Error: Running Homebrew as root is extremely dangerous and no longer supported.\nAs Homebrew does not drop privileges on installation you would be giving all\nbuild scripts full access to your system.\n"}

mahdipub:
> The Mac OS X GH Action failed with:
> 
> TASK [Common : Add AdoptOpenJDK Java Repo] ************************************* fatal: [localhost]: FAILED! => {"changed": false, "msg": "added: 0, unchanged: 0, error: failed to tap: AdoptOpenJDK/openjdk due to Error: Running Homebrew as root is extremely dangerous and no longer supported.\nAs Homebrew does not drop privileges on installation you would be giving all\nbuild scripts full access to your system.\n"}

I am not sure how did you set up your environment but I think it is because you are running ansible with root/privileged user. In your case you need become to switch to less privileged user. But most of the time, the user we run on mac is not root and either is not privileged and need sudo to do something.

karianna:
> > The Mac OS X GH Action failed with:
> > TASK [Common : Add AdoptOpenJDK Java Repo] ************************************* fatal: [localhost]: FAILED! => {"changed": false, "msg": "added: 0, unchanged: 0, error: failed to tap: AdoptOpenJDK/openjdk due to Error: Running Homebrew as root is extremely dangerous and no longer supported.\nAs Homebrew does not drop privileges on installation you would be giving all\nbuild scripts full access to your system.\n"}
> 
> I am not sure how did you set up your environment but I think it is because you are running ansible with root/privileged user. In your case you need become to switch to less privileged user. But most of the time, the user we run on mac is not root and either is not privileged and need sudo to do something.

That's a separate fix that may need to be applied to how we set up the GH action in that case.  Is that something you can fix as part of this PR?

mahdipub:
> > > The Mac OS X GH Action failed with:
> > > TASK [Common : Add AdoptOpenJDK Java Repo] ************************************* fatal: [localhost]: FAILED! => {"changed": false, "msg": "added: 0, unchanged: 0, error: failed to tap: AdoptOpenJDK/openjdk due to Error: Running Homebrew as root is extremely dangerous and no longer supported.\nAs Homebrew does not drop privileges on installation you would be giving all\nbuild scripts full access to your system.\n"}
> > 
> > 
> > I am not sure how did you set up your environment but I think it is because you are running ansible with root/privileged user. In your case you need become to switch to less privileged user. But most of the time, the user we run on mac is not root and either is not privileged and need sudo to do something.
> 
> That's a separate fix that may need to be applied to how we set up the GH action in that case. Is that something you can fix as part of this PR?

I guess the fix is you run MAC pb with a lower privileged users and then this change will work for you either. Nothing to do with ansible code itself.

sxa:
> That's a separate fix that may need to be applied to how we set up the GH action in that case. Is that something you can fix as part of this PR?

@karianna are you saying that the operations in mac actions is running as an administrator user by default? FYI @mahdipub the code that runs the action is in https://github.com/adoptium/infrastructure/blob/master/.github/workflows/build_mac.yml if you want to look at it.

From other invocations in GitHub actions it looks like the playbooks are running with `"ansible_user: runner"` so if that's a non-administrative user and this PR is no longer switching to it that would likely be the cause of the problem int he checks with this PR (I'm speculating here of course!)

mahdipub:
> @mahdipub the code that runs the action is in https://github.com/adoptium/infrastructure/blob/master/.github/workflows/build_mac.yml if you want to look at it.

@sxa, from the `build_mac.yml` I see that [this line](https://github.com/adoptium/infrastructure/blob/master/.github/workflows/build_mac.yml#L44) runs with sudo. 
```
sudo ansible-playbook -i hosts playbooks/AdoptOpenJDK_Unix_Playbook/main.yml --skip-tags="hosts_file,hostname,brew_upgrade,brew_cu,kernel_tuning,adoptopenjdk,jenkins,nagios,superuser,swap_file,crontab"
```
That could be the cause. Why do we need sudo there? I believe this sudo turns the user `root` then it tried to run ansible on remote host by `root` as no `-u` or other options provided.
The other support for my argument is, if the user is root before that line, then brew will argue [here](https://github.com/adoptium/infrastructure/blob/master/.github/workflows/build_mac.yml#L27):
```
    - name: Install Python
      run: brew install python@3.12 --overwrite
```
to not run brew as root. So before that playbook line user is not privileged.

karianna:
> > That's a separate fix that may need to be applied to how we set up the GH action in that case. Is that something you can fix as part of this PR?
> 
> @karianna are you saying that the operations in mac actions is running as an administrator user by default?

Yes :-)

sxa:
> That could be the cause. Why do we need sudo there?

That would be a question for @gdams

PR #3663, opened 2024-07-09
vagrant: Upgrade Fedora 35 vagrantfile to Fedora 40

Haroon-Khel:
- [x] commit message has one of the [standard prefixes](https://github.com/adoptium/infrastructure/blob/master/CONTRIBUTING.md#commit-messages)
- [ ] [faq.md](https://github.com/adoptium/infrastructure/blob/master/FAQ.md) updated if appropriate
- [ ] other documentation is changed or added (if applicable)
- [x] playbook changes run through [VPC](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)
- [ ] VPC/QPC not applicable for this PR
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

Fixes https://github.com/adoptium/infrastructure/issues/3662

PR #3761, opened 2024-10-02
UnixPB: Add Fedora 40 Support

### LINKED ISSUE #3599: Ansible request for Fedora playbook corrections
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

Delete as appropriate from this list:

- Bug in ansible playbook

Details:

- libnss3.so is not a valid package on Fedora so should be removed (From [this thread](https://adoptium.slack.com/archives/C53GHCXL4/p1697552390431979) I thought it had been but seems not universally). We've now removed it from the [UBI8 dockerfile](https://github.com/adoptium/infrastructure/pull/3595/files) but it is still in the playbooks.
- Fedora (verified on F37/riscv64) also seems to be trying to install java 1.7.0 which is not in its repositories any more so that should also be removed.
- `Create Japanese locale` in the `Common` role doesn't seem to be working on F37/rv64: `/etc/locale.gen and /var/lib/locales/supported.d/local are missing. Is the package \"locales\" installed?`
- Surprisingly `tar` is missing from the machines by default so the `ant-contrib` task falls over failing to extract that tarball
- Also while not generic to Fedora the gcc tags (`gcc_48,gcc_7,gcc_9,gcc_10,gcc_11`) need to be expliclty skipped on riscv64 otherwise it tries to download and install the adoptopenjdk gcc versions which aren't available for riscv64.
- Nagios_plugins role does not have a Fedora configuration so that role fails and `nagios_plugins` had to be added to the skip list in the previous bullet point (EDIT: That doesn't work as it wipes out the vendor files retrieval so the jenkins role fails - copied out that tag from the vendor files role for now)

Found as part of testing the machine referenced in https://github.com/adoptium/infrastructure/issues/3598#issuecomment-2167899948

Noting that:
- The [VPC runs with Fedora 35](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/OS=Fedora35,label=vagrant/1877/console) do not appear to be falling foul of this
- VPC should probably be updated to use Fedora 40 now.

steelhead31:
Fixes #3599 

Correct Unix playbooks to allow execution on Fedora 40.

##### Checklist

- [X] playbook changes run through [VPC](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)

Fedora 40: VPC : https://ci.adoptium.net/job/VagrantPlaybookCheck/1974/ 
Ubuntu 22: VPC : https://ci.adoptium.net/job/VagrantPlaybookCheck/OS=Ubuntu2204,label=vagrant/1975/
Debian 10: VPC : https://ci.adoptium.net/job/VagrantPlaybookCheck/OS=Debian10,label=vagrant/1975/

### DIFF
--- a/ansible/pbTestScripts/vmDestroy.sh
+++ b/ansible/pbTestScripts/vmDestroy.sh
@@ -61,9 +61,8 @@ checkOS() {
                         osToDestroy="D8" ;;
                 "Debian10" | "debian10" | "D10" | "d10" )
                         osToDestroy="D10" ;;
-								"Fedora40" | "fedora40" | "F40" | "F40" )
-				                osToDestroy="F40" ;;
-
+                "Fedora40" | "fedora40" | "F40" | "f40" )
+                        osToDestroy="F40" ;;
 		"FreeBSD12" | "freebsd12" | "F12" | "f12" )
 			osToDestroy="FBSD12" ;;
 		"Solaris10" | "solaris10" | "Sol10" | "sol10" )
--- a/ansible/pbTestScripts/vmDestroy.sh
+++ b/ansible/pbTestScripts/vmDestroy.sh
@@ -61,6 +61,9 @@ checkOS() {
                         osToDestroy="D8" ;;
                 "Debian10" | "debian10" | "D10" | "d10" )
                         osToDestroy="D10" ;;
+								"Fedora40" | "fedora40" | "F40" | "F40" )
+				                osToDestroy="F40" ;;
+
 		"FreeBSD12" | "freebsd12" | "F12" | "f12" )
 			osToDestroy="FBSD12" ;;
 		"Solaris10" | "solaris10" | "Sol10" | "sol10" )
@@ -70,7 +73,7 @@ checkOS() {
   	"Windows2022" | "Win2022" | "W22" | "w22" )
 	                       osToDestroy="W2022";;
 	              "all" )
-                        osToDestroy="U16 U18 U20 U21 U22 C6 C7 C8 D8 D10 FBSD12 Sol10 W2012 W2022" ;;
+                        osToDestroy="U16 U18 U20 U21 U22 C6 C7 C8 D8 D10 F40 FBSD12 Sol10 W2012 W2022" ;;
 		"")
 			echo "No OS detected. Did you miss the '-o' option?" ; usage; exit 1;;
 		*) echo "$OS is not a currently supported OS" ; listOS; exit 1;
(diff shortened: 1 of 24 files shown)


ansible/pbTestScripts/vmDestroy.sh
```diff
@@ -61,6 +61,9 @@ checkOS() {
                         osToDestroy="D8" ;;
                 "Debian10" | "debian10" | "D10" | "d10" )
                         osToDestroy="D10" ;;
+								"Fedora40" | "fedora40" | "F40" | "F40" )
```
karianna, line 64: whitespace seems out

karianna (changes requested):

ansible/pbTestScripts/vmDestroy.sh
steelhead31: fixed, thanks!.

karianna (approved):

sxa (approved):
I'm ok with this, but noting:
- are there any implications on older Fedora versions of completely removing `libnss3.so` here - I think that was added for some of the security tests.
- Noting that in the `nasm` playbook we have a mix of syntax: `ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "22")` vs `(ansible_distribution == "Ubuntu" and ansible_distribution_major_version | int >= 24)`. I definitely have a recollection - possibly ansible version specific - that one is better than the other so maybe we should standardise. I have a feeling it was the `| int` version that was preferable but don't hold me to that :-)

steelhead31:
> *   are there any implications on older Fedora versions of completely removing `libnss3.so` here - I think that was added for some of the security tests.

For the currently supported versions of Fedora ( 39 & 40 ) the libnss3 libraries are installed via nss-devel & nss-util packages, so removing it has no impact on the currently supported versions.  
  
The vagrant playbook checks run the SSL tests as part of the TEST\_JDK flag, so this would highlight any issues for the Fedora 40 run.  
 

> *   Noting that in the `nasm` playbook we have a mix of syntax: `ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "22")` vs `(ansible_distribution == "Ubuntu" and ansible_distribution_major_version | int >= 24`

I've standardised this to be consistent for all the major version checks. I believe the int comparison is a more reliable comparison for numeric values, particularly as using \<= with string operations is a bit inconsistent. the string operations are fine for a simple ==.

sxa:
@steelhead31 I guess this supercedes https://github.com/adoptium/infrastructure/pull/3663?

steelhead31:
> @steelhead31 I guess this supercedes #3663?

Yup, I hadnt seen the earlier one.

Issue #3217, opened 2023-10-17
Identify suitable off-machine location for storing private build container images


sxa:
Due to licensing issues we cannot ship the RHEL/s390x build image to the public so it will need to be in a private location. This issue covers identifying somewhere suitable.

Follow-up to https://github.com/adoptium/infrastructure/issues/2731

Noting also that at present the images are not being automatically regenerated when the playbooks are updated, which could be a first step that can be done before pushing it externally.

sxa:
This is also relevant for the newly tested docker images for Windows which will also need to be in a private repository so they can be shared:

Options:

- Dockerhub private registry (Subject to any licensing restrictions on those repositories)
- [Azure Container Registry](https://azure.microsoft.com/en-us/products/container-registry/) (Potentially best suited for our Windows Azure hosts
- [GitHub container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) (ghcr.io)
- [quay.io](https://quay.io/)
- Self-hosted?
- Something else?

Noting that the OpenJ9 team use a private repository so that is known to work well.

sxa:
@gdams Since you're more likely to be familiar with it (and I understand from earlier discussions this might be outside our sponsored credits) can you look at what we need to get us set up with ACR for Windows (and probably s390x for now) build docker images?

gdams:
I had an old container registry setup a while ago. There's a few steps to getting it running in a GH actions env

sxa:
@gdams Where are we with this at the moment? Are we set up with appropriate credentials to be able to pull from the Azure registry during the builds?

sxa:
Reopening as this got auto-closed as a result of merging the change to pull Windows ONLY from azurecr

sxa:
GCHR should be suitable for this as per the suggestion in https://github.com/adoptium/infrastructure/issues/3830#issuecomment-2598794006 so I will close this on the basis that a location has been identified and selected (this was also discussed in the [Dec 4, 2024 PMC meeting](https://github.com/adoptium/adoptium/blob/main/Meetings/PMC%20Meeting%20Minutes/2024/2024-12-04%20Adoptium%20PMC%20Meeting%20Minutes.pdf))

PR #3780, opened 2024-10-15
docker: add action to deploy windows build image

gdams:
step 1 of https://github.com/adoptium/infrastructure/issues/3217

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
@@ -64,12 +64,7 @@
   tags: ANT
 
 - name: Download ant-contrib
-  win_get_url:
-    url: https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.zip
-    dest: c:\temp\ant-contrib.zip
-    force: no
-    checksum: 22bae6c3ddf1a464b285784599eef8698f64dde24378c77e42522a536b88cbbc
-    checksum_algorithm: sha256
+  win_shell: c:\cygwin64\bin\curl -L -o /cygdrive/c/temp/ant-contrib.zip https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.zip
   when: (not ant_contrib_installed.stat.exists)
   tags: ANT
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,11 @@ on:
     branches:
     - master
 
+# Cancel existing runs if user makes another push.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 permissions:
   contents: read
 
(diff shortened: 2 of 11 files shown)


sxa (changes requested):
Can you add in a doc change to the table in https://github.com/adoptium/infrastructure/blob/master/FAQ.md#what-about-the-builds-that-use-the-dockerbuild-tag too please as part of this PR?

gdams:
> Can you add in a doc change to the table in https://github.com/adoptium/infrastructure/blob/master/FAQ.md#what-about-the-builds-that-use-the-dockerbuild-tag too please as part of this PR?

done

sxa (approved):
Approving in principle for when we've got this tested (We probably don't want to be running of my infra branch so we should perhaps merge those changes - or equivalents - into this PR once verified as reliable.

Just as a point of clarification - the use of this won't be going near production until after the release cycle so if this is a bit rough round the edges (quite likely) it will still be nice to have the image creation/publish process in place.

sxa (changes requested):
I'm going to block this for now as the changes to remove openj9 support (added after I gave my approval) should not be in here. If we want to skip the inclusion of those installs in the adoptium build containers then let's skip the roles in the dockerfile for now but since our build scripts do support non-temurin builds and other projects may be using it this should be a separate issue for discussion and not just done inside this PR.

gdams:
> I'm going to block this for now as the changes to remove openj9 support (added after I gave my approval) should not be in here. If we want to skip the inclusion of those installs in the adoptium build containers then let's skip the roles in the dockerfile for now but since our build scripts do support non-temurin builds and other projects may be using it this should be a separate issue for discussion and not just done inside this PR.

updated PTAL

ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/ANT/tasks/main.yml
```diff
@@ -64,12 +64,7 @@
   tags: ANT
 
 - name: Download ant-contrib
-  win_get_url:
-    url: https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.zip
-    dest: c:\temp\ant-contrib.zip
-    force: no
-    checksum: 22bae6c3ddf1a464b285784599eef8698f64dde24378c77e42522a536b88cbbc
-    checksum_algorithm: sha256
+  win_shell: c:\cygwin64\bin\curl -L -o /cygdrive/c/temp/ant-contrib.zip https://sourceforge.net/projects/ant-contrib/files/ant-contrib/ant-contrib-1.0b2/ant-contrib-1.0b2-bin.zip
```
sxa, line 67: We need to add in something here to verify the download on this download in the absence of the checksum check that was in `win_get_url`

gdams: I couldn't consistently get the same shasum on that download URL (I have no idea why and not sure if we should be worried about this)

karianna: I would be concerned.

sxa: Agreed @karianna ... We had a SHA check with the previous `win_get_url` operation so if something is preventing that when using curl then that would definitely need to be understood before this can go in.

sxa: (Since I'm going to be away for most of the next week, if these issues are resolved then feel free to dismiss my review)

sxa:
ping @gdams is there anything blocking this?

PR #1480, opened 2020-07-30
UnixPB: Add RH7 to EPEL and add Docker repository for RH7 ppc64le

KevinBonilla:
- Found a different approach (thanks @Willsparker for the help), for installing `docker-ce` on RH7 ppc64le. 
- Requires EPEL repo, modified `Common/RedHat.yml` to add RH7 to EPEL repo list
- Also following this [IBM documentation ](https://developer.ibm.com/linuxonpower/docker-on-power/), and using the `yum_repository` module, I had to add the correct docker.repo in order for RH7 ppc64le to install `docker-ce` 

Fyi: previous PR #1460

Signed-off-by: kevin-bonilla <kevin.bonilla@ibm.com>

PR #3802, opened 2024-11-06
Add conditionals for Rhel9 in common role

AswathySK:
The common role for Unix playbook doesnt include conditionals for rhel9 causing playbook failures.

Failures caused due to current condition:
- EPEL release not enabled for rhel9
- packags included in Installing additional build tools if NOT RHEL 8 task is applicable for rhel9 as well. For example NTP is replaced by chronyd.
-  Additional build tools for rhel 8 is also required by rhel 9

##### Checklist


AdamBrousseau (commented):
@AswathySK can we make these conditions smarter so we don't have to keep updating them every time a new OS is released? Or do we prefer to leave it like this so we can test the new major OS releases?

AswathySK:
@AdamBrousseau instead of changing conditional to do it for everything greater than or equal to rhel 8 instead?
There could be change in package dependencies for the future releases.
So how to make the change right now is up to the community I suppose.

AdamBrousseau:
Sure. We can get other opinions. If we change it to be something like `>=`, and it stops working, the PBs will presumably fail and someone can investigate at that point. That would have to happen either way, but at least this way, you don't have to monitor and update it when it works.

AdamBrousseau:
@sxa do you have an opinion on the above?

AswathySK:
@AdamBrousseau , I have made the change as per your suggestion.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -16,12 +16,12 @@
...
   tags: patch_update
 
-- name: Enable EPEL release for RHEL8 or RHEL6 or RHEL7
+- name: Enable EPEL release for RHEL9 or RHEL8 or RHEL6 or RHEL7
   yum: name=https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
   failed_when: false
   when:
     - ansible_architecture != "s390x"
-    - (ansible_distribution_major_version == "8") or (ansible_distribution_major_version == "6") or (ansible_distribution_major_version == "7")
+    - (ansible_distribution_major_version | int >= 6)
```
AdamBrousseau, line 24: If this is needed for rhel6+ why do we even need a condition for it?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -136,17 +136,17 @@
...
   package: "name={{ item }} state=latest"
   with_items: "{{ Java_RHEL6_PPC64 }}"
   when: (ansible_distribution_major_version == "6" and ansible_architecture == "ppc64")
 
-- name: Install Java when RedHat 8
+- name: Install Java when RedHat 8 and RedHat 9
   package: "name={{ item }} state=latest"
   with_items: "{{ Java_RHEL8 }}"
-  when: (ansible_distribution_major_version == "8")
+  when: (ansible_distribution_major_version | int >= 8)
```
AdamBrousseau, line 149: Are these 3 cases still what we want?
From https://github.com/adoptium/infrastructure/blob/894d84829c15361afe29e72dc66387a4be5e986a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml#L95
```
Java_RHEL8:
  - java-1.8.0-openjdk-devel

Java_NOT_RHEL6_PPC64:             # Not RHEL8 either
  - java-1.7.0-openjdk-devel
  - java-1.8.0-openjdk-devel

Java_RHEL6_PPC64:
  - java-1.7.0-ibm-devel
  - java-1.8.0-ibm-devel
```
What is java 7 and 8 used for?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -52,11 +52,11 @@
 ##########################
 # Additional build tools #
 ##########################
-- name: Install additional build tools if NOT RHEL 8
+- name: Install additional build tools if NOT RHEL 8 AND RHEL9
   package: "name={{ item }} state=latest"
   with_items: "{{ Additional_Build_Tools_NOT_RHEL8 }}"
```
AdamBrousseau, line 57: Should we rename this variable to `Additional_Build_Tools_NOT_RHEL8Plus`?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -102,11 +102,11 @@
     - (ansible_distribution_major_version == "6") or (ansible_distribution_major_version == "7")
   tags: build_tools
 
-- name: Install additional build tools for RHEL 8
+- name: Install additional build tools for RHEL >= 8
   package: "name={{ item }} state=latest"
   with_items: "{{ Additional_Build_Tools_RHEL8 }}"
```
AdamBrousseau, line 107: Should we rename this variable to `Additional_Build_Tools_RHEL8Plus`?

AdamBrousseau (changes requested):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: That sounds good.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: @AdamBrousseau, I am not sure. Should we bump it to 17 or 21?

AdamBrousseau: As far as I can tell it's been there since the beginning. Java is needed for 2 purposes that I know of.
- Connecting agents to jenkins.
- Bootjdk for compiling java.

Since Jenkins now requires jdk17 to run agents, and jdk11 was required 2 years ago, I don't think leaving this at java8 for jenkins connections is necessary.
For bootjdk requirement, the build scripts will pull the needed jdk during the build process so having a permanent install of 8 is not necessary in my opinion.

I would want an opinion from Adopt/@sxa before we remove it completely though.

sxa: As far as the boot JDK is concerned I don't want the builds pulling it down dynamically in most cases - that's mostly just a fallback mechanism. Other than pulling down the source, the builds should be able to run without network access.

I suspect those java-1.8.0* packages are no longer required, at least by Temurin, so it would likely be safe to remove them.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: Yes that makes sense.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: Yep

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: No reason that I can think of 🤷🏻 
Looks like it initially had rhel 6 or rhel 8 and then 7 was added in under https://github.com/adoptium/infrastructure/pull/1480
Stripping out the condition altogether sounds good to me unless anyone can think of a good reason to leave it there.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: Yes I think that should be safe. It will hopefully be obvious if some part of the process requires it, in which case it's easy to back it out again :-)

Can the 1.7 version be installed from the RHEL9 repositories?

karianna:
@AswathySK Linter failures as well

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -52,11 +51,11 @@
 ##########################
 # Additional build tools #
 ##########################
-- name: Install additional build tools if NOT RHEL 8
+- name: Install additional build tools if NOT RHEL8 and above
```
sxa, line 54: Remove trailing whitespace
```suggestion
- name: Install additional build tools if NOT RHEL8 and above
```

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -52,11 +51,11 @@
 ##########################
 # Additional build tools #
 ##########################
-- name: Install additional build tools if NOT RHEL 8
+- name: Install additional build tools if NOT RHEL8 and above 
   package: "name={{ item }} state=latest"
-  with_items: "{{ Additional_Build_Tools_NOT_RHEL8 }}"
+  with_items: "{{ Additional_Build_Tools_NOT_RHEL8Plus }}"
```
sxa, line 56: We probably need to go through and sanitise some of these names. This one should probably be `Additional_Build_Tools_RHEL6_RHEL7` now but that doesn't need to be changed in this PR

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -136,17 +135,17 @@
...
-    - not (ansible_distribution_major_version == "8")
+    - not (ansible_distribution_major_version | int >= 8)
 
 - name: Install Java when RedHat 6 on ppc64
   package: "name={{ item }} state=latest"
   with_items: "{{ Java_RHEL6_PPC64 }}"
   when: (ansible_distribution_major_version == "6" and ansible_architecture == "ppc64")
 
-- name: Install Java when RedHat 8
+- name: Install Java when RedHat 8 and RedHat 9
```
sxa, line 145:
```suggestion
- name: Install Java when RedHat 8 and later
```

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
```diff
@@ -92,17 +92,6 @@ Additional_Build_Tools_RHEL7_s390x:
   - libstdc++.s390                # a dependency required for executing a 32-bit C binary
   - yum-utils                     # yumdownloader required for devkit creation
 
-Java_RHEL8:
```
sxa, line 95: Since we still have the reference to `Java_RHEL8` above this shouldn't be removed

sxa (commented):
A few comments/suggestions

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AdamBrousseau: @sxa is there another part of the PB installing java for use by the jenkins agent that Adopt is running?
We have an internal Semeru role that installs 21 for this.

sxa: Yep. The adoptopenjdk_install role: .https://github.com/adoptium/infrastructure/blob/028773c140a974f741d6b4ab5a78f381f2df7a68/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml#L131 
It's extracting as a tarball and doesn't set any default on the system to point to it, so the Jenkins agent is pointed directly to it under /usr/lub/jvm when it's started

sxa:
New VPC run at https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/2038/ (Expect at least CentOS6 to fail - unrelated to this PR)

karianna:
@AswathySK merge conflicts.

AswathySK:
@karianna @sxa , 
should this merge wait until code freeze is over?

karianna:
Yes and I'd love to see the CEntOS6 and Solaris GH actions start passing again, but I suspect that's a separate issue

AswathySK:
@karianna ,
when will the code freeze be over?

sxa:
Looks like it was done during my time away at FOSDEM :-)
https://github.com/adoptium/.github/pull/193
Currently catching up with notifications ...

AswathySK:
@sxa @karianna ,
Can this be merged then as the code freeze is over?

AswathySK:
@sxa @karianna ,
Can this be merged then as the code freeze is over?

karianna:
Hmm, I think it actually wants the two failing GH Actions to pass (although I realize they're not related)

sxa:
@AswathySK Have you confirmed that a system installed with these playbooks is capable of building and testing openjdk? When I tried to run a temurin build in a CentOS Stream 9 (Should be equivalent to RHEL9) containers configured with these changes it was still missing some things like `diffutils`.

sxa:
> Hmm, I think it actually wants the two failing GH Actions to pass (although I realize they're not related)

I'd be happy to dismiss the bots review for the purposes of getting this through when we agree it's ready.

sxa:
Side note: It looks like the playbooks are not happy running under the version of ansible that ships with CentOS Stream 10 (and presumably true for the upcoming RHEL10 release), but that's for the future.

sxa (approved):
From the limited testing I can do just now this looks good I think. Needs a second reviewer then we can get it in.

sxa:
@AswathySK Did you manage to get this working on RHEL9 given that the `curl` installation that the playbooks do would confict with `curl-minimal` which is there by default?

sxa:
@AswathySK Any thoughts on the curl concern in my last comment? It would be good to know that these changes have been tested on a clean RHEL9 system

AswathySK:
@sxa For rhel9 I think curl is available by default or else even by the package manager.
So it is not necessary to run the curl from the playbooks.

sxa:
> @sxa For rhel9 I think curl is available by default or else even by the package manager. So it is not necessary to run the curl from the playbooks.

I don't have easy access to the RHEL9 container images at the moment but both CentOS Stream9 (the base for RHEL) and Rocky 9 both use curl-minimal by default in their container images so I'd be a bit surprised if RHEL10 was different:

```
sxa@dockerhost-equinix-ubuntu2404-armv8-1:~$ docker run quay.io/centos/centos:stream9 rpm -qa | grep ^curl
curl-minimal-7.76.1-31.el9.aarch64
sxa@dockerhost-equinix-ubuntu2404-armv8-1:~$ docker run rockylinux:9 rpm -qa | grep ^curl
curl-minimal-7.76.1-26.el9_3.2.0.1.aarch64
sxa@dockerhost-equinix-ubuntu2404-armv8-1:~$ 
```

sxa:
Checked with someone who has access to the full RHEL9 containers and they also seem to have `curl-minimal`

AswathySK:
@sxa for rhel 9 and for future versions should we install curl using the curl role available in Adopt?
Cant we rely on the one provided by the package manager?
[Maybe change ](https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml) - here it installs curl using package manager itself.

AswathySK:
@sxa should any changes be made?

sxa:
> @sxa should any changes be made?

Yes since I do not believe this will work in a RHEL9/CS9 container, and I can't approve until that has been fixed.

Specifically on your previous comment:

> @sxa for rhel 9 and for future versions should we install curl using the curl role available in Adopt?

The separate curl role is intended for building from source if the package manager version is older than what we need (e.g. RHEL6). It should not be used for modern distributions which have new enough versions supplied in the distribution's repositories.

> Cant we rely on the one provided by the package manager?

Do you mean `curl` or `curl-minimal`? I have not tested the build and aqa test process with curl-minimal but I would hope it would be ok. It would be good to verify that before merging this, but even if we can use curl-minimal we will require some changes as the playbooks currently break trying to install curl.

> [Maybe change ](https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/> AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml)- here it installs curl using package manager itself.

Yes that code is in there but it fails if `curl-minimal` is already installed - easy to see the problem here:
```
sxa@fedora:/home/sxa/git/openjdk-infrastructure$ podman run -it quay.io/centos/centos:stream9 dnf install curl
CentOS Stream 9 - BaseOS                                                                           3.5 MB/s | 8.7 MB     00:02    
CentOS Stream 9 - AppStream                                                                        4.4 MB/s |  23 MB     00:05    
CentOS Stream 9 - Extras packages                                                                   17 kB/s |  20 kB     00:01    
Error: 
 Problem: problem with installed package curl-minimal-7.76.1-31.el9.x86_64
  - package curl-minimal-7.76.1-31.el9.x86_64 from @System conflicts with curl provided by curl-7.76.1-31.el9.x86_64 from baseos
  - package curl-minimal-7.76.1-23.el9.x86_64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.x86_64 from baseos
  - package curl-minimal-7.76.1-26.el9.x86_64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.x86_64 from baseos
  - package curl-minimal-7.76.1-28.el9.x86_64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.x86_64 from baseos
  - package curl-minimal-7.76.1-29.el9.x86_64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.x86_64 from baseos
  - package curl-minimal-7.76.1-31.el9.x86_64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.x86_64 from baseos
  - cannot install the best candidate for the job
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
sxa@fedora:/home/sxa/git/openjdk-infrastructure$ 
```

sxa:
I also tried on a new real (not containerised) RHEL9 box on ibmcloud and it fell over at this step:
`failed: [141.125.159.28] (item=texinfo) => {"ansible_loop_var": "item", "changed": false, "failures": ["No package texinfo available."], "item": "texinfo", "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}`

sxa:
`texinfo` needs the CRB repo to be available e.g. `dnf search --enablerepo codeready-builder-for-rhel-9-x86_64-rpms texinfo`

sxa:
`NTP_TIME` role also seems to fail:
```
TASK [NTP_TIME : Start NTP for RedHat, SLES 12 and CentOS 7] *******************
fatal: [141.125.159.28]: FAILED! => {"changed": false, "msg": "Could not find the requested service ntpd: host"}
```
Re-running with `--skip-tags NTP_TIME`

sxa:
@AswathySK Are you still looking at this? I managed to get it to run on CentOS Stream 9 other than [the issue with curl](https://github.com/adoptium/infrastructure/issues/3871) but it would be good to have your view on the other RHEL9 issues that I saw in the comments above.

AswathySK:
@sxa ,I have noticed the NTP issue before. I was thinking of creating a separate PR for that.
And for the codeready-builder-for-rhel-9-x86_64-rpms  we enable it with our internal PB tbh. Maybe we can include that under the Common role in adoptium to enable additional repos for rhel then?
And we depend on the curl installed by default rather than on the curl compiled with the playbook.
How do you want to approach this?

sxa:
> @sxa ,I have noticed the NTP issue before. I was thinking of creating a separate PR for that.

If it doesn't work with these playbook changes, then I feel they should be included in this PR otherwise these changes do not work for the purposes of running on RHEL9.

> And for the codeready-builder-for-rhel-9-x86_64-rpms we enable it with our internal PB tbh. Maybe we can include that under the Common role in adoptium to enable additional repos for rhel then?

If we can be sure it will always be available then yes that would be an option. I've also tested with just having it as a separate task which runs `dnf --enablerepo=crb install texinfo -y` instead of enabling the repository globally and that seems to work.

> And we depend on the curl installed by default rather than on the curl compiled with the playbook. How do you want to approach this?

`curl` should only be compiled by the playbook when the system one is too old (i.e. <7.58.0). Otherwise it uses the system `curl` package. I'm surprised you didn't hit a problem with the `Common` role trying to install `curl`. Was `curl-minimal` not on your RHEL9 system?

sxa:
> > And for the codeready-builder-for-rhel-9-x86_64-rpms we enable it with our internal PB tbh. Maybe we can include that under the Common role in adoptium to enable additional repos for rhel then?

> If we can be sure it will always be available then yes that would be an option. I've also tested with just having it as a separate task which runs dnf --enablerepo=crb install texinfo -y instead of enabling the repository globally and that seems to work.

Thinking about it a little more I think it's probably best to enable CRB on RHEL9+ as you suggest.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -162,7 +162,7 @@
 
 - name: Find Default JDK
   stat:
-    path: /usr/lib/jvm/java-1.8.0
+    path: /usr/lib/jvm/java-21-openjdk
```
sxa, line 189: This will try to set the default to 21 for all releases (other than 6 which is overridden in the next task) won't it? Earlier versions of RHEL (Such as RHEL7 which we use for s390x builds) won't have java-21 installed from the repositories by default.

AswathySK: For the rhel9 machine I tested there was no /usr/lib/jvm/java-1.8.0 by default.
That is why I updated it to jdk-21.
Should we separate the task for ones with jdk-21 y default and the one otherwise?

sxa:
> @sxa , I noticed one more thing with enabling the EPEL repo. Since the failed_when: false parameter is used the EPEL repo is not actually getting enabled. And for mercurial to be installed EPEL is needed as well.

I wonder if we could just install the `epel-release` package instead of the current method - that should do everything correctly I think.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: Yeah I think that's the best option for now - I'm surprised it was a problem with 9 though.
No RHEL versions come with java by default as far as I know but the playbooks should install `java-1.8.0-openjdk-devel` even on RHEL9 (just not on 10)

sxa:
@AdamBrousseau Looks like you still have a pending change request on this - is it still valid?

sxa:
> @sxa , I noticed one more thing with enabling the EPEL repo. Since the failed_when: false parameter is used the EPEL repo is not actually getting enabled. And for mercurial to be installed EPEL is needed as well.

Noting that mercurial is no longer needed (unless someone in the IBM team tells me you still need it!) and it was removed on most distributions as part of https://github.com/adoptium/infrastructure/pull/3913 - although I see that RedHat.yml got missed from that, so feel free to remove it under this PR too :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: Line number 97-98:
```
Java_RHEL8Plus:
  - java-21-openjdk-devel
```
I changed it like this.
https://github.com/adoptium/infrastructure/pull/3802/files#r1838438091 discussion.

AswathySK:
@sxa ,
The epel-release package is available in centOS only.
For rhel I think this is the way to go i guess.
If thats the case, should we import the gpg key or disable the check?
Reference: https://www.liquidweb.com/blog/enable-epel-repository/

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: @sxa , we can leave it as java-1.8.0-openjdk-devel itself then.
What do you suggest.

sxa: > Line number 97-98:

That removes the java 1.8.0 install for earlier distributions. For <=8 we would want java 8, for >=9 we'd need 21.

The two options are:
1. Make a change that allows a different version to be installed on each version
2. Remove the installation of the openjdk package since we do install Temurin elsewhere in the playbooks (but I think there are some things in the test suites that depend on the default JDK, which is why we set the default here)

The important thing is that /something/ works for every RHEL version.

sxa:
> @sxa , The epel-release package is available in centOS only. For rhel I think this is the way to go i guess. If thats the case, should we import the gpg key or disable the check? Reference: https://www.liquidweb.com/blog/enable-epel-repository/ https://www.redhat.com/en/blog/install-epel-linux

I definitely don't want to remove the check, so importing the key seems like the better option here.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: I corrected the jdk version to 21 , that is why 21 was installed instead.
I reverted the changes.
```
Java_RHEL8Plus:
  - java-1.8.0-openjdk-devel
```
```
- name: Find Default JDK
  stat:
    path: /usr/lib/jvm/java-1.8.0
  register: jdk_path
  tags: default_java
```

AswathySK:
@sxa ,
I have made the change to import the key and also made change to the jdk being installed and default path being set.
can you please take a look at it?

sxa:
> @sxa , I have made the change to import the key and also made change to the jdk being installed and default path being set. can you please take a look at it?

The changes look good - thanks! It wont' work for EL10 where the 1.8.0 package doesn't exist but I can cover that in some other changes I'm making for EL10 :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
```diff
@@ -153,7 +166,7 @@
   with_items: "{{ Java_NOT_RHEL6_PPC64 }}"
   when:
     - not (ansible_distribution_major_version == "6" and ansible_architecture == "ppc64")
-    - not (ansible_distribution_major_version == "8")
+    - not (ansible_distribution_major_version | int >= 8)
   tags: install_java
 
 - name: Install Java when RedHat 6 on ppc64
```
AdamBrousseau, line 172: Can we remove these 2? I don't think we need ppc64(BE) anymore. I doubt Adopt does either.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AdamBrousseau: It sounds like we don't need any java 8 installs here then?
Stewart has confirmed the Jenkins agent uses the jvm from the adoptopenjdk_install role. That role installs all the bootjdks for the builds. And for Semeru, we have a separate PB for the jenkins agent.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: Adopt definitely doesn't - as I recall that was left in for IBM internal use so from an Adoptium side I have no particular view one way or another on that one ;-)

sxa: > It sounds like we don't need any java 8 installs here then?

I believe there are some things that rely on having a "default" java in the path. These things in the common role are to set that correctly I believe. In the future we could switch to making one of the Adoptium ones the default on the system but since that's not in place I would not want to remove that from this PR.

That would be a good separate issue/PR though, and should apply to all distributions.

AdamBrousseau: Fair enough. So cleanup the ppc64 bits and then this will be ready :+1:

steelhead31 (approved):
Looks good

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: @AdamBrousseau ,
Sorry I didnt get what you meant.
Do you need me to remove the task to install java for rhel 6 BE?

sxa: Honestly I'm not sure. If we are 100% site than nothing requires JDK8 specifically then it should be ok but I'm not that confident at the moment and have requested input from the team as part of https://github.com/adoptium/infrastructure/issues/3949 (Also referenced in Slack at https://adoptium.slack.com/archives/C53GHCXL4/p1747759541294539)

AswathySK: Can I open a separate PR for removing the ppc64(BE) and updating jdk to 21 for rhel8+ or should it be done in this PR itself?

AswathySK:
@sxa I will make the change in this PR itself.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
sxa: I'm absolutely fine with it being separate from this PR since it's not specifically about RHEL9 👍
Hopefully @AdamBrousseau is ok with that approach too.

AswathySK:
@sxa Since tehre are still no final decision regarding the jdk, I have created a separate PR for correcting the NTP time:
https://github.com/adoptium/infrastructure/pull/3953

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/RedHat.yml
AswathySK: @AdamBrousseau ,
Thoughts?

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/RedHat.yml
```diff
@@ -94,16 +94,20 @@ Additional_Build_Tools_RHEL7_s390x:
...
-Java_RHEL6_PPC64:
-  - java-1.7.0-ibm-devel
-  - java-1.8.0-ibm-devel
+#Java_NOT_RHEL6_PPC64:             # Not RHEL8 either
+#  - java-1.7.0-openjdk-devel
+#  - java-1.8.0-openjdk-devel
+#
+#Java_RHEL6_PPC64:
+#  - java-1.7.0-ibm-devel
+#  - java-1.8.0-ibm-devel
```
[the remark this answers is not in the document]
AswathySK: That was a mistake from my part.
I removed it now.
Can you please take a loo kat it now @AdamBrousseau ?

AswathySK:
> This lftm now. Although I'm still confused why we install java 7 & 8 vs 8 only when on rhel8+.

@sxa , Do you know any history related to this?
Eventhough we install both even in Rhel6, java8 is set as default.

sxa:
> This lftm now. Although I'm still confused why we install java 7 & 8 vs 8 only when on rhel8+.

This really should be a discussion separate from this PR, but Temurin uses 7 to bootstrap 8 where possible (With one exception, [we aim to use `n-1` to bootstrap `JDKn`](https://github.com/adoptium/temurin-build/blob/792825a68785e7d7c4b29905c5044b6767a044de/build-farm/make-adopt-build-farm.sh#L139)).

RHEL8+ does not java `java-1.7.0` in the repositories which is why it is not installed there. That is ok, as we do not currently build on any RHEL8 systems, so it is not a hard requirement there at the current time, so we can let it slide. If it was available in those repositories, we could include it.

Issue #3797, opened 2024-11-04
Implement policy for managing docker build caches


sxa:
Seen recently on test-azure-ubuntu2404-x64-1 but this is also a follow-on to https://github.com/adoptium/infrastructure/issues/3007

 https://docs.docker.com/build/cache/garbage-collection/ has some information on how the cache can be maintained. We should ensure that the amount of space used by docker is managed appropriately. I believe we have implemented some regular cleanups on some machines and we should ensure that we deploy this more universally - likely as part of the `adoptopenjdk` tags in the UNIX playbook. The test jobs are increasingly making use of docker, and so a certain amount of build up will occur and this needs to be managed.

steelhead31:
Im currently writing a docker housekeeping script to help maintain the size of the overlay2.

steelhead31:
Im currently writing up a script to do some automated housekeeping on machines that use docker ( to try and ensure the builder cache is kept at a manageable volume ), and that the overlay2 directory is periodically housekept, as it has a tendency to grow and isnt managed by the normal "docker system prune" commands...  currently I plan to have a script that behaves as follows :

1.  Looks for any images that havent been used in the last 6 months
2.  Deletes any containers based on the images identified in 1.
3.  Deletes the images identified in 1.
4.  Prunes the docker builder cache by deleting everything older than 14 days
5.  Prune any unused images
6.  Prune any unused containers ( older than 7 days )
7.  Prune any docker volumes ( that arent in use )
8.  Prune any docker networks that arent in use

Does anybody see any issues with this, the initial version of the script will allow it to run in "reporting" mode, and highlight what it plans to delete, so we can do some shakedown testing ahead of implementation.

steelhead31:
Running my proposed script on test-ibmcloud-rhel7-x64-1 identifies the following :+1:   
  
`Cutoff date: 2024-05-01 (Wed May  1 00:00:00 CDT 2024)`  
`Mode: Listing images and containers that would be deleted.`  
`Would delete: alpine:3.14 (Created on: 2023-03-29 13:19:37 -0500 CDT)`  
`Would delete: rabbitmq:3.9-management (Created on: 2023-03-24 13:45:07 -0500 CDT)`  
`Would delete: postgres:14.1 (Created on: 2022-01-26 19:10:09 -0600 CST)`  
`Would delete: apicurio/apicurio-registry-mem:2.1.5.Final (Created on: 2021-12-22 12:00:15 -0600 CST)`  
[...]

PR #3825, opened 2024-11-27
UnixPB: Add Docker Overlay Check Plugin Script For Nagios

steelhead31:
Ref: #3797 

Add a new script to be used by nagios for monitoring the size of the /var/lib/docker/overlay2 folder, and ensure it is installed alongside the other nagios plugins.

##### Checklist

VPC OK : https://ci.adoptium.net/job/VagrantPlaybookCheck/2008/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
@@ -8,7 +8,7 @@ DEFAULT_WARN_THRESHOLD=20
 DEFAULT_CRIT_THRESHOLD=40
 
 # Folder to check
-TARGET_DIR="/var/lib/docker/overlay2"
+TARGET_DIR="/var/lib/docker/overlay2/"
 
 # Function to display usage
 usage() {
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# Nagios plugin to check the size of /var/lib/docker/overlay2
+# Allows passing thresholds as parameters
+
+# Default thresholds (in GB)
+DEFAULT_WARN_THRESHOLD=30
+DEFAULT_CRIT_THRESHOLD=40
+
+# Folder to check
+TARGET_DIR="/var/lib/docker/overlay2"
+
+# Function to display usage
+usage() {
+    echo "Usage: $0 -w <warn_threshold_in_GB> -c <crit_threshold_in_GB>"
+    echo "  -w: Warning threshold in GB (default: $DEFAULT_WARN_THRESHOLD)"
+    echo "  -c: Critical threshold in GB (default: $DEFAULT_CRIT_THRESHOLD)"
+    exit 3
+}
+
+# Parse command-line arguments
+while getopts "w:c:" opt; do
+    case $opt in
+        w) WARN_THRESHOLD=$OPTARG ;;
+        c) CRIT_THRESHOLD=$OPTARG ;;
+        *) usage ;;
+    esac
+done
+
+# Set default thresholds if not provided
+WARN_THRESHOLD=${WARN_THRESHOLD:-$DEFAULT_WARN_THRESHOLD}
+CRIT_THRESHOLD=${CRIT_THRESHOLD:-$DEFAULT_CRIT_THRESHOLD}
+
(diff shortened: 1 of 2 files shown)


ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+# Nagios plugin to check the size of /var/lib/docker/overlay2
+# Allows passing thresholds as parameters
+
+# Default thresholds (in GB)
+DEFAULT_WARN_THRESHOLD=30
```
sxa, line 7: I'd prefer to go lower if possible as an initial pass so that we see which ones are hitting this level. I don't think there's a lot we're doing on these systems that should push it above 20 in normal use, and if there is it would be good to collect data on it before increasing the number
```suggestion
DEFAULT_WARN_THRESHOLD=20
```

Noting that the current x64 dockerhost is currently sitting with 
```
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          72        11        64.79GB   40.02GB (61%)
Containers      12        11        32.89GB   410.4MB (1%)
Local Volumes   2         0         0B        0B
Build Cache     430       0         1.135GB   1.135GB
```
So it'll trigger a `critical` threshold immediately. I suspect it's going to make sense to have separate thresholds for `dockerhost` systems and other test machines which are running docker, but I think 20/40 is likely a good place to start for this.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
```diff
@@ -0,0 +1,66 @@
...
+DEFAULT_CRIT_THRESHOLD=40
+
+# Folder to check
+TARGET_DIR="/var/lib/docker/overlay2"
```
sxa, line 11: I'm not sure we have an such situations, but maybe have the `/` on the end so we can see into symlinks if that's been used to put the file system somewhere else without an explicit mount at that location.
```suggestion
TARGET_DIR="/var/lib/docker/overlay2/"
```

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
```diff
@@ -0,0 +1,66 @@
...
+FOLDER_SIZE_BYTES=$(du -sb "$TARGET_DIR" 2>/dev/null | awk '{print $1}')
+
+# Handle error if du fails
+if [[ -z "$FOLDER_SIZE_BYTES" ]]; then
+    echo "CRITICAL: Failed to determine the size of $TARGET_DIR."
+    exit 2
+fi
+
+# Convert size to GB for display
+FOLDER_SIZE_GB=$(echo "scale=2; $FOLDER_SIZE_BYTES / (1024 * 1024 * 1024)" | bc)
```
sxa, line 54: You don't see `bc` used often enough these days ;-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Determine Nagios status
+if (( FOLDER_SIZE_BYTES > CRIT_THRESHOLD_BYTES )); then
```
sxa, line 57: I was thinking that this needed `$` prefxes but apparently not.
Also noting that while these two numbers are calculated within the script, this syntax prevents the issues described in https://yossarian.net/til/post/some-surprising-code-execution-sources-in-bash/ with using `-gt` and similar constructs, which is giong to take a bit of getting used to

sxa (approved):
Approving on the assumption you check whether you agree with my symlink comment before merging :-)

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
steelhead31: I agree!

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Nagios_Plugins/tasks/additional_plugins/check_docker_overlay2_size.sh
steelhead31: I based it on an existing script, but I wanted the parameters to be simple, rather than in bytes/kilobytes/megabytes :)

andrew-m-leonard (approved):

PR #3832, opened 2024-12-02
Vagrant: Improve Test Coverage & Align Tests In VPC process.

steelhead31:
Update the vagrant playbook check tests of the built JDK, this will now run the standard smoke tests after each build, plus 2 sets of tests ( jdk_math and hotspot_sanity ). These tests are now identical across linux & windows.

##### Checklist

- [X] playbook changes run through [VPC](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)

VPC Run : https://ci.adoptium.net/job/VagrantPlaybookCheck/2025/

### DIFF
--- a/ansible/pbTestScripts/testJDK.sh
+++ b/ansible/pbTestScripts/testJDK.sh
@@ -17,28 +17,25 @@ if [[ "$(uname)" == "SunOS" ]]; then
 	export PATH="/usr/local/bin:/opt/csw/bin:${PATH}"
 fi
 
+## Run The Smoke Tests To Ensure The JDK Build OK
 mkdir -p $HOME/testLocation
 [ ! -d $HOME/testLocation/aqa-tests ] && git clone https://github.com/adoptium/aqa-tests.git $HOME/testLocation/aqa-tests
 # cd to aqa-tests as required by https://github.com/adoptium/aqa-tests/issues/2691#issue-932959102
 cd $HOME/testLocation/aqa-tests
-$HOME/testLocation/aqa-tests/get.sh
+$HOME/testLocation/aqa-tests/get.sh --vendor_repos https://github.com/adoptium/temurin-build --vendor_branches master --vendor_dirs /test/functional
 cd $HOME/testLocation/aqa-tests/TKG || exit 1
+export BUILD_LIST=functional/buildAndPackage
+$MAKE_COMMAND compile
+$MAKE_COMMAND _extended.functional
 
-# Solaris runs a different test to Linux.
-# See: https://adoptium.slack.com/archives/C53GHCXL4/p1641311568115100?thread_ts=1641296204.114900&cid=C53GHCXL4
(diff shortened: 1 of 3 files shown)


sxa (commented):
I'm not sure about this. Proof the smoke tests in is definitely good, but the system tests and MBCS ones were specifically chosen because they validate other bits of the playbook setup (The STF build processes, and the perl installation). It looks like this is removing that validation.

steelhead31:
> I'm not sure about this. Adding the smoke tests in is definitely good, but the system tests and MBCS ones were specifically chosen because they validate other bits of the playbook setup (The STF build processes, and the perl installation). It looks like this is removing that validation.

I'll look to identify some equivalents and add those in too, hopefully its a bit simpler to add more tests now I've rationalised the setup a little :)

Haroon-Khel (approved):

sxa:
> > I'm not sure about this. Adding the smoke tests in is definitely good, but the system tests and MBCS ones were specifically chosen because they validate other bits of the playbook setup (The STF build processes, and the perl installation). It looks like this is removing that validation.
> 
> I'll look to identify some equivalents and add those in too, hopefully its a bit simpler to add more tests now I've rationalised the setup a little.

👍🏻  The important thing here is that we're looking to test the playbooks, not the JDK specifically, so we need a fairly quick to run set that covers anything that might get missed.

karianna:
> > I'm not sure about this. Adding the smoke tests in is definitely good, but the system tests and MBCS ones were specifically chosen because they validate other bits of the playbook setup (The STF build processes, and the perl installation). It looks like this is removing that validation.
> 
> I'll look to identify some equivalents and add those in too, hopefully its a bit simpler to add more tests now I've rationalised the setup a little.
> 
> New VPC run in progress : https://ci.adoptium.net/job/VagrantPlaybookCheck/2027/

Had some failures although not 100% sure related to your changes!

steelhead31:
> > > I'm not sure about this. Adding the smoke tests in is definitely good, but the system tests and MBCS ones were specifically chosen because they validate other bits of the playbook setup (The STF build processes, and the perl installation). It looks like this is removing that validation.
> > 
> > I'll look to identify some equivalents and add those in too, hopefully its a bit simpler to add more tests now I've rationalised the setup a little.  
> > New VPC run in progress : https://ci.adoptium.net/job/VagrantPlaybookCheck/2027/
> 
> Had some failures although not 100% sure related to your changes!

I dont think they are, Im rerunning the 2 problem platforms centos7 and debian 10 ( I have a feeling JDK21 doesn't build on debian 10 )... but the first run was successful across Ubuntu 22,24, Fedora 40, Solaris, & Windows, and all builds and tests ran successfully.

sxa (approved):

PR #3878, opened 2025-02-11
UnixPB: Fix Problematic Centos6 Downloads

### LINKED ISSUE #3839: CentOS6 cannot download from https://www.kernel.org in GIT_Source role
Seen in VPC and [PR checks](https://github.com/adoptium/infrastructure/actions/runs/12178978609/job/33970110121?pr=3580). Last successful [VPC run](https://ci.adoptium.net/job/VagrantPlaybookCheck/OS=CentOS6,label=vagrant/) on that distribution was on October 28th.

```
19:19:17 TASK [GIT_Source : Download git source] ****************************************
19:19:17 FAILED - RETRYING: Download git source (3 retries left).
19:19:17 FAILED - RETRYING: Download git source (2 retries left).
19:19:17 FAILED - RETRYING: Download git source (1 retries left).
```
`19:19:17 fatal: [localhost]: FAILED! => {"attempts": 3, "changed": false, "msg": "Failed to validate the SSL certificate for www.kernel.org:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine  (the python executable used (/usr/bin/python) is version: 2.6.6 (r266:84292, Jun 20 2019, 14:14:55) [GCC 4.4.7 20120313 (Red Hat 4.4.7-23)]) or you can install the `urllib3`, `pyOpenSSL`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible. The exception msg was: hostname 'www.kernel.org' doesn't match 'm.sni-311-default.ssl.fastly.net'."}`

steelhead31:
Fixes #3839 

Downloads for Git Source and Nvidia Cuda toolkit are failing on CentOS6 due to the certificate errors, and the lack of support of SNI enabled validation, due to the age of the O/S.

This PR adds an alternative, but still secure download mechanism for CentOS6 only, to allow the playbook to be backwards compatible.

##### Checklist

VPC In Progress : https://ci.adoptium.net/job/VagrantPlaybookCheck/2061/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
@@ -42,6 +42,7 @@
     - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout is version_compare('2.15', operator='lt'))
     - ansible_distribution != "FreeBSD"
     - ansible_distribution != "CentOS" or ansible_distribution_major_version | int != 6
+    - ansible_distribution != "RedHat" or ansible_distribution_major_version | int != 6
   tags: git_source
 
 - name: Download git source for CentOS 6
@@ -50,7 +51,8 @@
     warn: false
   when:
     - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout is version_compare('2.15', operator='lt'))
-    - ansible_distribution == "CentOS" and ansible_distribution_major_version | int == 6
+    - (ansible_distribution == "CentOS" or ansible_distribution == "RedHat")
+    - ansible_distribution_major_version | int == 6
   tags: git_source
 
 - name: Verify checksum for CentOS 6 Git Source download ...
@@ -59,7 +61,8 @@
   failed_when: "'107116489f10b758b51af1c5dbdb9a274917b0fb67dc8eaefcdabc7bc3eb3e6a' not in checksum_result.stdout"
   when:
     - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout is version_compare('2.15', operator='lt'))
-    - ansible_distribution == "CentOS" and ansible_distribution_major_version | int == 6
+    - (ansible_distribution == "CentOS" or ansible_distribution == "RedHat")
+    - ansible_distribution_major_version | int == 6
   tags: git_source
 
 - name: Extract git source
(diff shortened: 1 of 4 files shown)


karianna (approved):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
```diff
@@ -41,6 +41,25 @@
   when:
     - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout is version_compare('2.15', operator='lt'))
     - ansible_distribution != "FreeBSD"
+    - ansible_distribution != "CentOS" or ansible_distribution_major_version | int != 6
```
sxa, line 44: We should probably have a RHEL clause alongside CentOS here, although I appreciate it's now an unlikely use case.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
```diff
@@ -41,6 +41,25 @@
...
+- name: Download git source for CentOS 6
+  shell: "wget -q -O /tmp/git-2.15.0.tar.xz https://www.kernel.org/pub/software/scm/git/git-2.15.0.tar.xz"
+  args:
+    warn: false
+  when:
+    - (git_installed.rc != 0 ) or (git_installed.rc == 0 and git_version.stdout is version_compare('2.15', operator='lt'))
+    - ansible_distribution == "CentOS" and ansible_distribution_major_version | int == 6
+  tags: git_source
+
+- name: Verify checksum for CentOS 6 Git Source  download ...
```
sxa, line 56:
```suggestion
- name: Verify checksum for CentOS 6 Git Source download ...
```

sxa (changes requested):
A couple of minor points and one thought (which I wouldn't address in this PR regardless): Is it worth just using wget/curl everywhere instead of having this duplicated?

steelhead31:
> A couple of minor points and one thought (which I wouldn't address in this PR regardless): Is it worth just using wget/curl everywhere instead of having this duplicated?

Ansible doesnt particularly like commands being delegated to the shell, it prefers the get\_url function... but definitely something to consider.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/GIT_Source/tasks/main.yml
steelhead31: Added.

sxa (approved):
As long as the `ansible_distribution in ["CentOS", "RedHat"]` syntax works everywhere we run ansible then I'm ok with this (Not sure we're written the expressions like that anywhere else)

steelhead31:
> As long as the `ansible_distribution in ["CentOS", "RedHat"]` syntax works everywhere we run ansible then I'm ok with this (Not sure we're written the expressions like that anywhere else)

I've just encountered an issue with this syntax, despite it working ok locally, I've amended it to be more uniform.

Issue #3871, opened 2025-02-06
Ansible request for `curl` on CentOS Stream9


sxa:
Please put the name of the software product (and affected platforms if relevant) in the title of this issue

Delete as appropriate from this list:

- Bug in ansible playbook

Details:
Running the playbooks against a CentOS Stream 9 docker containers gives the following error:
`258.4 failed: [localhost] (item=curl) => {"ansible_loop_var": "item", "changed": false, "failures": [], "item": "curl", "msg": "Depsolve Error occurred: \n Problem: problem with installed package curl-minimal-7.76.1-31.el9.aarch64\n  - package curl-minimal-7.76.1-31.el9.aarch64 from @System conflicts with curl provided by curl-7.76.1-31.el9.aarch64 from baseos\n  - package curl-minimal-7.76.1-23.el9.aarch64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.aarch64 from baseos\n  - package curl-minimal-7.76.1-26.el9.aarch64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.aarch64 from baseos\n  - package curl-minimal-7.76.1-28.el9.aarch64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.aarch64 from baseos\n  - package curl-minimal-7.76.1-29.el9.aarch64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.aarch64 from baseos\n  - package curl-minimal-7.76.1-31.el9.aarch64 from baseos conflicts with curl provided by curl-7.76.1-31.el9.aarch64 from baseos\n  - cannot install the best candidate for the job", "rc": 1, "results": []}`

Noting that `curl` is already in the base image so skipping this may be the simplest solution, although we should understand the underlying reason why this is happening - potentially a repository issue

sxa:
This is more fundamental - `dnf install curl` doesn't work out of the box ...
```
sxa@fedora:~$ podman run -it quay.io/centos/centos:stream9 dnf install curl
CentOS Stream 9 - BaseOS                                                                                6.5 MB/s | 8.4 MB     00:01    
CentOS Stream 9 - AppStream                                                                             8.3 MB/s |  22 MB     00:02    
CentOS Stream 9 - Extras packages                                                                        62 kB/s |  20 kB     00:00    
Error: 
 Problem: problem with installed package curl-minimal-7.76.1-29.el9.x86_64
[...]

PR #3870, opened 2025-02-06
unixPB: add diffutils and procps-ng to RH/Cent/Fedora

sxa:
Fixes https://github.com/adoptium/infrastructure/issues/3869
Related: https://github.com/adoptium/infrastructure/pull/3802 (The RHEL8-RHEL8Plus change in that PR may mean this needs an update to go it to avoid a conflict).

Note that as per the notes in 3869 CentOS6 will generally already have procps installed so, for simplicity,  I have not explicitly listed it here.

##### Checklist

- [x] commit message has one of the [standard prefixes](https://github.com/adoptium/infrastructure/blob/master/CONTRIBUTING.md#commit-messages)
- [ ] [faq.md](https://github.com/adoptium/infrastructure/blob/master/FAQ.md) updated if appropriate
- [ ] other documentation is changed or added (if applicable)
- [ ] playbook changes run through [VPC](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access) No due to https://github.com/adoptium/infrastructure/issues/3867 but I'm testing in containers New VPC run https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/2087/
- [ ] VPC/QPC not applicable for this PR
- [ ] for inventory.yml changes, bastillion/nagios/jenkins updated accordingly

PR #3942, opened 2025-05-15
unixPB: remove curl from CentOS Common role

sxa:
This seems like an odd thing to do, but the comment at https://github.com/adoptium/infrastructure/issues/3871#issuecomment-2883322700 is the best explanation. CS9 provides `curl-minimal` by default for `curl` whereas earlier versions and CS10 provide it via `curl`. Since they're all in the default package, it's less messy to just remove it unelss we find a deployment where it is not on the system by default.

##### Checklist

- [x] playbook changes run through [VPC](https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/) or [QPC](https://ci.adoptium.net/view/Tooling/job/QEMUPlaybookCheck/) (if you have access)  https://ci.adoptium.net/view/Tooling/job/VagrantPlaybookCheck/2088/

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/vars/CentOS.yml
@@ -12,7 +12,6 @@ Build_Tool_Packages:
   - bzip2
   - ca-certificates
   - cpio
-  - curl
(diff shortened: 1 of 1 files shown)


sxa:
Actually this change is also in https://github.com/adoptium/infrastructure/pull/3870 so I'll cancel this and merge that one which is already approved.

PR #4161, opened 2025-11-26
Add pipeline code for test_image_updater job

sxa:
This is the code that is being used by the [test_image_updater job](https://ci.adoptium.net/job/test_image_updater) in support of https://github.com/adoptium/aqa-tests/issues/5684

At the moment the job is running from my branch but the code is now in a reasonable state (it just doesn't have a full set of architectures) so I'm upstreaming this so that the job can be pointed at the upstream repository.

##### Checklist

### DIFF
--- a/ansible/docker/Jenkinsfile.test
+++ b/ansible/docker/Jenkinsfile.test
@@ -5,7 +5,7 @@ pipeline {
             parallel {
                 stage('Ubuntu24.04 x64') {
                     agent {
-                        label "dockerBuild&&linux&&x64&&dockerhost-azure-ubuntu2204-x64-1"
+                        label "dockerBuild&&linux&&x64"
                     } 
                     steps {
                         dockerBuild('amd64', 'ubuntu2404', 'Dockerfile.u2404')
@@ -21,7 +21,7 @@ pipeline {
                 }
                 stage('UBI10 x64') {
                     agent {
-                        label "dockerBuild&&linux&&x64&&dockerhost-azure-ubuntu2204-x64-1"
+                        label "dockerBuild&&linux&&x64"
                     } 
                     steps {
                         dockerBuild('amd64', 'ubi10', 'Dockerfile.ubi10')


ansible/docker/Jenkinsfile.test
```diff
@@ -0,0 +1,134 @@
...
+        dockerImage.push()
+        sh "docker inspect --format='{{index .RepoDigests 0}}' ghcr.io/adoptium/test-containers:${distro}-${architecture} > ${distro}_linux-${architecture}.sha256"
+        archiveArtifacts artifacts: '*linux*.sha256', fingerprint: true
+    }
+}
+
+def dockerManifest() { 
+    // dockerhub is the ID of the credentials stored in Jenkins
+    docker.withRegistry('https://ghcr.io', 'ghcr-adoptium') {
+        git poll: false, url: 'https://github.com/sxa/infrastructure.git'
```
smlambert: Are you still wanting to reference sxa/infrastructure here?

sxa:
> Added a few comments, but approving on the notion that updates can come later.
> 
> I am guessing you want it merged and then we can start changing the location from test-container to adoptium-test-images ? Or is now a good time to switch to that?

I'd say keep that separate. I'd like the outcome from https://github.com/adoptium/containers/issues/853 before making the change to use the newer repo.

ansible/docker/Jenkinsfile.test
sxa: To be honest I'm pretty sure those aren't needed so I'll run a test with those two lines commented out.

ansible/docker/Jenkinsfile.test
```diff
@@ -0,0 +1,132 @@
...
+                    agent {
+                        label "dockerBuild&&linux&&aarch64"
+                    } 
+                    steps {
+                        dockerBuild('arm64', 'ubuntu2404', 'Dockerfile.u2404')
+                    }
+                }
+                stage('UBI10 x64') {
+                    agent {
+                        label "dockerBuild&&linux&&x64&&dockerhost-azure-ubuntu2204-x64-1"
```
steelhead31: Should this be pinned to a named host?.. I know we probably only have that one capable one :)

sxa: That's a fair comment (both parts of it! We used to have a skytap x64 dockerhost at one point too...) Those tags without the hostname would also trigger on the dynamic azure machines too so I may have stuck that one a while back to prevent that for debugging purposes. I'll give it a run through without that fixed hostname.

PR #4238, opened 2026-02-04
Remove Solaris From Vagrant Checker

karianna:
We no longer build Solaris


karianna:
#4259 Now has the bot

sxa:
> I've only made the Vagrant Checker change as that was a GH Action triggered check on PR submissions (trying to keep that green). I agree there's a larger/further issue WRT to removing Solaris everywhere.

vagrantPlayBookcheck.sh is not exclusive to GH actions. It was originally written (and is still run and is green) via https://ci.adoptium.net/job/VagrantPlaybookCheck/ - having said that I believe @steelhead31 was prevented from doing some updates on the host machine due to still supporting this particular VM :-)

steelhead31:
I'd be happy to have Solaris removed from the VPC, on my ever growing to do list, is a plan to update/migrate the Jenkins based VPC jobs, Im just waiting on the destiny of its current host. If we lose the host, then I'll likely rework some parts of this, and losing solaris even in the short term means I can look at bumping the pinned versions of Virtual Box, Vagrant, Python & Ansible, as they all had to be held back to support Solaris.

PR #4325, opened 2026-04-24
UnixPB: New dwarf library role

aswinkr77:
- new dwarf library role
 - fix for https://github.com/eclipse-openj9/openj9/issues/15390

Signed-off-by: Aswin K R <aswinkr77@gmail.com>

cc: @pshipton 

##### Checklist

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
@@ -63,6 +63,10 @@
       tags: adoptopenjdk
     - role: NTP_TIME
       when: ansible_distribution != "MacOSX"
+    - role: dwarf_library         # OpenJ9
+      when:
+        - ansible_architecture == "aarch64"
+        - ansible_os_family == "RedHat"
     - gcc_48
     - role: gcc_7                 # OpenJ9
       tags: [build_tools, build_tools_openj9]
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
@@ -6,6 +6,7 @@
 - name: Download libdwarf
   get_url:
     url: "https://github.com/davea42/libdwarf-code/releases/download/v2.3.1/libdwarf-2.3.1.tar.xz"
+    checksum: sha256:28cf9a5d27aceff5c1f906244a4fe7ae208e41d20a6d8fc7e091c633a40b6e97
     dest: /tmp/libdwarf.tar.xz
   tags: libdwarf
 
@@ -17,7 +18,7 @@
   tags: libdwarf
 
 - name: Run cmake
-  command: cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO
+  command: cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO -DCMAKE_INSTALL_PREFIX=/
   args:
     chdir: /tmp/libdwarf-2.3.1
   tags: libdwarf
@@ -28,41 +29,14 @@
     chdir: /tmp/libdwarf-2.3.1
   tags: libdwarf
 
-- name: Backup existing headers (if present)
-  command: mv {{ item.src }} {{ item.dest }}
-  loop:
-    - src: /usr/include/libdwarf/dwarf.h
-      dest: /usr/include/libdwarf/dwarf.h.org
-    - src: /usr/include/libdwarf/libdwarf.h
-      dest: /usr/include/libdwarf/libdwarf.h.org
-  ignore_errors: true # if /usr/include/libdwarf/ doesn't exists
-  tags: libdwarf
-
 - name: Ensure include directory exists
   file:
-    path: /usr/include/libdwarf
+    path: /usr/include/libdwarf-2
     state: directory
   tags: libdwarf
 
-- name: Create symlinks for headers
-  file:
-    src: "{{ item.src }}"
-    dest: "{{ item.dest }}"
-    state: link
-    force: yes
-  loop:
-    - src: /usr/local/include/dwarf.h
-      dest: /usr/include/libdwarf/dwarf.h
-    - src: /usr/local/include/libdwarf.h
-      dest: /usr/include/libdwarf/libdwarf.h
-  tags: libdwarf
-
-- name: Copy libdwarf shared libraries to /lib64
-  shell: cp /usr/local/lib64/libdwarf.so.2.* /lib64
-  tags: libdwarf
-
-- name: Refresh linker cache
-  command: ldconfig
+- name: Move dwarf header files to libdwarf-2 directory
+  command: mv /usr/include/dwarf.h /usr/include/libdwarf.h /usr/include/libdwarf-2/
   tags: libdwarf
 
 - name: Verify installation
@@ -71,6 +45,7 @@
   changed_when: false
   tags: libdwarf
 
-- debug:
+- name: Output ldconfig
+  debug:
     var: ldconfig_output.stdout_lines
   tags: libdwarf
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
@@ -1,0 +1,76 @@
+---
+###################################
+#  Compile specific dwarf library #
+###################################
+
+- name: Download libdwarf
+  get_url:
+    url: "https://github.com/davea42/libdwarf-code/releases/download/v2.3.1/libdwarf-2.3.1.tar.xz"
+    dest: /tmp/libdwarf.tar.xz
+  tags: libdwarf
+
+- name: Extract libdwarf
+  unarchive:
+    src: /tmp/libdwarf.tar.xz
+    dest: /tmp
+    remote_src: true
+  tags: libdwarf
+
+- name: Run cmake
+  command: cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO
+  args:
+    chdir: /tmp/libdwarf-2.3.1
+  tags: libdwarf
+
+- name: Build and install
+  command: make install
+  args:
+    chdir: /tmp/libdwarf-2.3.1
+  tags: libdwarf
+
+- name: Backup existing headers (if present)
+  command: mv {{ item.src }} {{ item.dest }}
+  loop:
+    - src: /usr/include/libdwarf/dwarf.h
+      dest: /usr/include/libdwarf/dwarf.h.org
+    - src: /usr/include/libdwarf/libdwarf.h
+      dest: /usr/include/libdwarf/libdwarf.h.org
+  ignore_errors: true # if /usr/include/libdwarf/ doesn't exists
+  tags: libdwarf
+
+- name: Ensure include directory exists
+  file:
+    path: /usr/include/libdwarf
+    state: directory
+  tags: libdwarf
+
+- name: Create symlinks for headers
+  file:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    state: link
+    force: yes
+  loop:
+    - src: /usr/local/include/dwarf.h
+      dest: /usr/include/libdwarf/dwarf.h
+    - src: /usr/local/include/libdwarf.h
+      dest: /usr/include/libdwarf/libdwarf.h
+  tags: libdwarf
+
+- name: Copy libdwarf shared libraries to /lib64
+  shell: cp /usr/local/lib64/libdwarf.so.2.* /lib64
+  tags: libdwarf
+
+- name: Refresh linker cache
+  command: ldconfig
+  tags: libdwarf
+
+- name: Verify installation
+  shell: ldconfig -v | grep libdwarf
+  register: ldconfig_output
+  changed_when: false
+  tags: libdwarf
+
+- debug:
+    var: ldconfig_output.stdout_lines
+  tags: libdwarf
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
@@ -18,7 +18,7 @@
   tags: libdwarf
 
 - name: Run cmake
-  command: cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO -DCMAKE_INSTALL_PREFIX=/
+  command: cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO -DCMAKE_INSTALL_PREFIX=/usr/local/
   args:
     chdir: /tmp/libdwarf-2.3.1
   tags: libdwarf
@@ -27,16 +27,6 @@
   command: make install
   args:
     chdir: /tmp/libdwarf-2.3.1
-  tags: libdwarf
-
-- name: Ensure include directory exists
-  file:
-    path: /usr/include/libdwarf-2
-    state: directory
-  tags: libdwarf
-
-- name: Move dwarf header files to libdwarf-2 directory
-  command: mv /usr/include/dwarf.h /usr/include/libdwarf.h /usr/include/libdwarf-2/
   tags: libdwarf
 
 - name: Verify installation


aswinkr77:
@pshipton please review

pshipton:
Looks good to me.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+  loop:
+    - src: /usr/include/libdwarf/dwarf.h
+      dest: /usr/include/libdwarf/dwarf.h.org
+    - src: /usr/include/libdwarf/libdwarf.h
+      dest: /usr/include/libdwarf/libdwarf.h.org
+  tags: libdwarf
+
+- name: Ensure include directory exists
+  file:
+    path: /usr/include/libdwarf
```
keithc-ca: Perhaps this should be `/usr/include/libdwarf-2` for consistency with distributions that include libdwarf version 2.*.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+
+- name: Extract libdwarf
+  unarchive:
+    src: /tmp/libdwarf.tar.xz
+    dest: /tmp
+    remote_src: true
+  tags: libdwarf
+
+- name: Run cmake
+  command: cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO
```
keithc-ca: Adding `-DCMAKE_INSTALL_PREFIX=/` would eliminate the need to move/copy files from `/usr/local`.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
pshipton: Agreed. I've tried this and it's also working.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
pshipton: Wouldn't that put the include files directly into /usr/include?

keithc-ca: Yes, but they can be moved. Other files (e.g. `.so`) land where we want them.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    chdir: /tmp/libdwarf-2.3.1
+  tags: libdwarf
+
+- name: Ensure include directory exists
+  file:
+    path: /usr/include/libdwarf-2
+    state: directory
+  tags: libdwarf
+
+- name: Create symlinks for headers
```
pshipton: Pls fix the comment.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+  command: mv /usr/include/dwarf.h /usr/include/libdwarf.h /usr/include/libdwarf-2/
+  tags: libdwarf
+
+- name: Refresh linker cache
+  command: ldconfig
```
pshipton: This is unnecessary. Running `ldconfig` and `ldconfig -v` do the same thing, except with `-v` it's verbose.

aswinkr77: Fixed it

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
aswinkr77: Done

aswinkr77:
@pshipton made the necessary changes

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+    chdir: /tmp/libdwarf-2.3.1
+  tags: libdwarf
+
+- name: Ensure include directory exists
+  file:
+    path: /usr/include/libdwarf-2
+    state: directory
+  tags: libdwarf
+
+- name: Move  dwarf header files to libdwarf-2 directory
```
keithc-ca: Double space before "dwarf".

pshipton (approved):

aswinkr77:
@keithc-ca fixed the white spaces

keithc-ca (approved):
Looks good, but I think this should be squashed.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -63,6 +63,11 @@
       tags: adoptopenjdk
     - role: NTP_TIME
       when: ansible_distribution != "MacOSX"
+    - role: dwarf_library
+      when:
+        - ansible_architecture == "aarch64"
+        - ansible_distribution != "MacOSX"
+        - ansible_distribution != "Solaris"
```
sxa: Why does the Solaris check need to be in here - that's never going to trigger with the architecture limited to `aarch64`? Suggest removing this line.
With the macos exclusion this will trigger on both Linux/aarch64 and Alpine/aarch64. Is that the intention (and if so can you confirm it has been tested on Alpine)? In which case perhaps it would be better just to include both of them explicitly instead of the negative condition on macos

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
+---
+###################################
+#  Compile specific dwarf library #
+###################################
+
+- name: Download libdwarf
+  get_url:
+    url: "https://github.com/davea42/libdwarf-code/releases/download/v2.3.1/libdwarf-2.3.1.tar.xz"
```
sxa, line 8: This needs a checksum (or similar) check on the download as a minimum

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
```diff
@@ -63,6 +63,11 @@
       tags: adoptopenjdk
     - role: NTP_TIME
       when: ansible_distribution != "MacOSX"
+    - role: dwarf_library
```
sxa: Since this appears to be an OpenJ9 requirement can you comment it as such in a similar way to other entries in this file.

sxa (changes requested):
A few comments :-)
We install libdwarf from the package manager in a number of places but I presume from the comments in the related issue that it's an old version on certain distributions. If that's the case then should we stop installing it on those and leave it to this role to avoid confusion? Also since it sounds like this is specific to CentOS/RHEL7 do you know why this would be specific to aarch64? I'd expect the same problem with versioning to be present across all architectures. Please add this information into the PR before it can be approved.

pshipton:
I don't know why it's specific to aarch64, but it is. The libdwarf available for other platforms supports gcc11+ but the one on aarch64 doesn't. I suppose I can look into the images and check the versions, I know the aarch64 one is 0.20130207.0.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
aswinkr77: @pshipton do we need to include Alpine?

pshipton: No, Semeru/OpenJ9 doesn't support Alpine, and we've never tried it afaik.

aswinkr77: Updated the when conditional with os family support with `ansible_os_family == "RedHat"`

aswinkr77: @sxa is this enough?

sxa:
> I don't know why it's specific to aarch64, but it is. The libdwarf available for other platforms supports gcc11+ but the one on aarch64 doesn't. I suppose I can look into the images and check the versions, I know the aarch64 one is 0.20130207.0.

Hmmm the C7/x64 image seems to be 20130207, although the C6/x64 one, which I'd have expected to be older, has one dated 20140413. (I think you use the C7 one though so it should be the same across x64/aarch64). Odd

sxa:
> Looks good, but I think this should be squashed.

Changes will get squashed when merged. Please do NOT squash prior to that as it makes it extremely difficult to review what's changed since I last looked at it.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -28,41 +29,14 @@
...
-      dest: /usr/include/libdwarf/libdwarf.h
-  tags: libdwarf
-
-- name: Copy libdwarf shared libraries to /lib64
-  shell: cp /usr/local/lib64/libdwarf.so.2.* /lib64
-  tags: libdwarf
-
-- name: Refresh linker cache
-  command: ldconfig
+- name: Move dwarf header files to libdwarf-2 directory
```
sxa: I'm a bit nervous about this - forgive me if I'm mistaken but isn't this install going to overwrite/remove the system installed files which is potentially going to make the rpm database invalid and potentially cause problems for future upgrades which might overwrite them?
Not applicable for CentOS7 but certainly likely for later ones which are still supported.
Ideally we'd put the new version under /usr/local somewhere to keep it separate and clear that it's not part of what's shipped int he base OS.

keithc-ca: The base OS won't have the `libdwarf-2` directory (or we wouldn't need this at all).

sxa: Correct but my concern was over moving things from `/usr/include/dwarf.h` if this new installation with cmake was going to overwrite something there, although it looks like `libdwarf-devel` doesn't put anything in that particular path - only `/usr/include/libdwarf` so it might not conflict. It still makes me nervous to not be using /usr/local though.

Are we sure that this cmake install won't overwrite anything already installed by the system libdwarf packages?

keithc-ca: We could avoid potential for conflict by omitting `libdwarf-devel` for aarch64-linux.

sxa: Yeah we could although I feel it would be a lot cleaner to have it installed into `/usr/local` unless there's a significant reason for doing otherwise. We should be able to allow people to run these playbooks on their local machine and not cause any "damage" to things they already have.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/main.yml
sxa: Yep. LGTM

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
keithc-ca: A problem with installing in `/usr/local` is that tooling (e.g. cmake) may not find it in that non-standard location.

pshipton: How about we put the files into /usr/local when building the image here, and as part of the Semeru/OpenJ9 builds we can add a step that copies them to the places we want and creates a new image file (similarly to how the cuda files are added), or think about modifying OMR to find the files under /usr/local.

keithc-ca: That would work, but it seems a little silly to have a feature needed only by OpenJ9, but not done in a way that works for OpenJ9. We may as well put this change in a Semeru-specific fork/branch.

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+  tags: libdwarf
+
+- name: Ensure include directory exists
+  file:
+    path: /usr/include/libdwarf-2
+    state: directory
+  tags: libdwarf
+
+- name: Move dwarf header files to libdwarf-2 directory
+  command: mv /usr/include/dwarf.h /usr/include/libdwarf.h /usr/include/libdwarf-2/
```
steelhead31: I'd prefer this to install to /usr/local/libdwarf/ or something similar, to avoid any risks to any standard system libraries etc.

steelhead31 (changes requested):

ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/dwarf_library/tasks/main.yml
sxa: > think about modifying OMR to find the files under /usr/local

+1 to this @pshipton - IMHO it seems inadvisable not to have the process allow for this level of flexibility (and we're talking about doing something - installing into /usr/local - which is fairly standard). 

> A problem with installing in /usr/local is that tooling (e.g. cmake) may not find it in that non-standard location.

"may". The install into there produces a `/usr/local/lib64/cmake/libdwarf` which cmake seems to process without problems and `echo | gcc -E -Wp,-v -` indicates that by default it will be searching for headers in `/usr/local/include` in preference to `/usr/include`.

`cmake` during the J9 build appears to locate it automatically too:
```
-- Found LibDwarf: /usr/local/lib64/libdwarf.so  
```
During the build it looks like `omr_ddrgen` has trouble locating it at runtime but that can be resolved by ensuring that `LD_LIBRARY_PATH` has `/usr/local/lib64` in it. Using that approach and then not installing the system default libdwarf package (to avoid it being picked up accidentally, which seems to happen at about 23% into the build process) would still be my preferred option here rather than fragmenting the build system by using a Semeru-specific fork/branch as it would allow the adoptium scripts and containers to work for building both Temurin and openj9 build variants.

keithc-ca: It's less problematic than I expected: Putting things in /usr/local seems reasonable. As you say, `LD_LIBRARY_PATH` can get us the rest of the way if necessary (we're doing similar things on some platforms).

aswinkr77: @keithc-ca I've updated the installation location to `/usr/local/` also removed the moving of existing dwarf libraries. Please review now

steelhead31 (approved):
Looks good.

sxa (approved):
I think /usr/local might be the default install location so doesn't need to be explicit but it doesn't do any harm.

sxa:
@steelhead31 I'll let you merge this one if you're around today as I'm not at a computer today so won't be able to check that the image updater jobs run through ok.

pshipton:
How long does it take to get a new image at ghcr.io? Does something need to be triggered to do that?

keithc-ca:
See https://ci.adoptium.net/job/centos7_docker_image_updater/775, where it failed:
```
#11 1139.6 TASK [dwarf_library : Run cmake] ***********************************************
#11 1140.2 fatal: [localhost]: FAILED! => {"changed": false, "cmd": "cmake -DBUILD_SHARED=YES -DBUILD_NON_SHARED=NO -DCMAKE_INSTALL_PREFIX=/usr/local/", "msg": "[Errno 2] No such file or directory", "rc": 2}
```

pshipton:
Does that mean cmake isn't available, and the step to build libdwarf needs to happen later?

keithc-ca:
I can't tell from that log whether it's `cmake` that's missing, whether the archive was unpacked someplace other than `/tmp/libdwarf-2.3.1`, whether `/usr/local/bin` does not exist, or something else.

keithc-ca:
My money is on missing `cmake` - it's also missing in ghcr.io/adoptium/adoptium_build_image:alpine3_linux-amd64.

sxa:
> I can't tell from that log whether it's `cmake` that's missing, whether the archive was unpacked someplace other than `/tmp/libdwarf-2.3.1`, whether `/usr/local/bin` does not exist, or something else.

Almost certainly because the new `dwarf_library` role has been added to the top level `main.yml` playbook BEFORE the `cmake` role so cmake isn't available when the dwarf build tries to use it (unless it's a platform where it's installed from the OS repositories in the `Common` role, which won't be the case for CentOS7). 

> My money is on missing cmake - it's also missing in ghcr.io/adoptium/adoptium_build_image:alpine3_linux-amd64.

The `cmake` role is intentionally skipped on Alpine as OpenJ9 don't build on there and it's not needed for Adoptium, so that shouldn't be an issue but it could probably be added into roles/Common/vars/Alpine.yml if it was ever required.

PR #4448, opened 2026-07-08
UnixPB: Add Qemu Config For Ubuntu Dockerhosts.

### LINKED ISSUE #4447: Ansible: Add QEMU binfmt F-flag configuration for riscv64 container emulation on x86_64 dockerhosts
Ansible: Add QEMU binfmt F-flag configuration for riscv64 container emulation on x86_64 dockerhosts

Bug in ansible playbook / Request for new playbook addition

**Affected platforms:** x86_64 Ubuntu 20+ dockerhosts running `linux/riscv64` Docker containers via QEMU emulation

---

Jenkins pipelines building JDK (Temurin) for `linux/riscv64` on x86_64 Azure Ubuntu 24.04 dockerhosts were failing with two distinct issues rooted in missing or incorrect QEMU binfmt configuration.

### Issue 1 — Container exits with code 255

QEMU user-mode emulation was not configured on the host. The kernel had no registered handler for RISC-V ELF binaries, so any attempt to exec a riscv64 binary inside a container returned exit code 255.

**Fix:** Run `multiarch/qemu-user-static --reset -p yes` to register binfmt handlers — however this alone is not sufficient, see Issue 2.

### Issue 2 — `g++` internal compiler error: Segmentation fault mid-build

```
g++: internal compiler error: Segmentation fault signal terminated program as1
gmake[3]: *** [...metaspaceSettings.o] Error 4
```

The host's `qemu-user-binfmt` apt package had registered the riscv64 binfmt handler **without** the `F` (fix-binary) flag (`flags: PO` instead of `flags: POF`). Without `F`, the QEMU static binary must exist inside the container filesystem at runtime — it doesn't, so emulation is unreliable and g++ segfaults mid-compilation. Running `multiarch/qemu-user-static --reset` did not resolve this because the apt-managed registration took precedence.

steelhead31:
Fixes #4447 

##### Checklist

:heavy_check_mark: VPC (To Check It Doesnt Break Regular PBs ): https://ci.adoptium.net/job/VagrantPlaybookCheck/2289/

DockerHost playbook completes ok locally.

### DIFF
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/dockerhost.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/dockerhost.yml
@@ -37,6 +37,8 @@
     - Vendor
     - IPv6
     - Docker
+    - role: Qemu-Config   # Ubuntu 20+ x86_64 only — sets up QEMU binfmt with F flag for riscv64 container emulation
+      tags: [qemu_config, adoptopenjdk]
     - DockerStatic
     - role: housekeeping           # override mode when needed
       vars:
--- a/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Qemu-Config/tasks/main.yml
+++ b/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Qemu-Config/tasks/main.yml
@@ -1,0 +1,87 @@
+---
+###############
+# Qemu-Config #
+###############
+
+# Configures QEMU binfmt_misc on Ubuntu 20+ x86_64 dockerhosts so that
+# linux/riscv64 containers can be emulated via QEMU user-mode.
+#
+# Key fix: re-registers the qemu-riscv64 binfmt entry with the F (fix-binary)
+# flag so the QEMU static binary is loaded into memory at registration time
+# and is therefore available inside containers that don't ship QEMU themselves.
+# Without F, g++ segfaults mid-compilation with RISC-V ELF targets.
+
+- name: Install qemu-user-static package
+  package:
+    name:
+      - qemu-user-static
+    state: present
+  when:
+    - ansible_distribution == "Ubuntu"
+    - ansible_distribution_major_version | int >= 20
+    - ansible_architecture == "x86_64"
+  tags: [qemu_config, adoptopenjdk]
+
+- name: Enable systemd-binfmt.service
+  service:
+    name: systemd-binfmt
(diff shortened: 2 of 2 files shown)


sxa:
We should also consider dropping building under qemu and buliding natively. The times in https://github.com/adoptium/temurin-build/issues/3876 for the qemu-based x64 hosts are very much "the best" but in reality we have seen build times of around 6 hours in some cases. For consistency it probably makes sense to allocate one of the new RISC-V systems at Scaleway to be a build system (Noting that it will likely need an extra parameter on the `docker run` invocation to be able to run java in a container nativel - ref https://fosstodon.org/@sxa/111624807824299455)

The new k3 system I have builds java 25 in about an hour, although we don't have them fully set up in the CI yet. But in anticipation of that moving to running on real hardware without qemu probably makes sense.

steelhead31:
> We should also consider dropping building under qemu and buliding natively. The times in [adoptium/temurin-build#3876](https://github.com/adoptium/temurin-build/issues/3876) for the qemu-based x64 hosts are very much "the best" but in reality we have seen build times of around 6 hours in some cases. For consistency it probably makes sense to allocate one of the new RISC-V systems at Scaleway to be a build system (Noting that it will likely need an extra parameter on the `docker run` invocation to be able to run java in a container nativel - ref https://fosstodon.org/@sxa/111624807824299455)
> 
> The new k3 system I have builds java 25 in about an hour, although we don't have them fully set up in the CI yet. But in anticipation of that moving to running on real hardware without qemu probably makes sense.

I've created this issue.. https://github.com/adoptium/infrastructure/issues/4449  , we could potentially revert this change, once we;ve completed the switch to native builds, I think in the short term, adding these steps, gated by adoptopenjdk, and limited to x64 / Ubuntu dockerhosts, means we've captured these important setup steps.

sxa (commented):
Have you checked if this will correctly survive a reboot? Stuff in `/proc` typically won't.

steelhead31:
> Have you checked if this will correctly survive a reboot? Stuff in `/proc` typically won't.

It does locally... the binfmt persistence stuff 

> Have you checked if this will correctly survive a reboot? Stuff in `/proc` typically won't.

Yes, the task at line 35, adds it via a service so it persists past reboot....

karianna (approved):

sxa (approved):

sxa:
FYI @luhenry in case you want to give the binfmt configuration an extra pair of eyeballs :-)

